PR #3, opened 2020-05-01
Set required node version to 12.10.0

cpietsch:
The required node versions for the recursive flag of fs.mkdir & fs.rmdir are:
- fs.mkdir: >=v11.3.0
- fs.rmdir: >=v12.10.0


larionov:
I think this is fixed in https://github.com/vuejs/vitepress/commit/7beb172b00d0ad5fe20e31fae8ce206e01a5a598

PR #5, opened 2020-05-04
Fix typo

silvandiepen:
Little typo in the MarkdownOptions


yyx990803:
Thanks, the source interface needs to be fixed too. It's better to use TS refactor to do this type of updates. See 49fc3cd

PR #14, opened 2020-05-14
chore: adding documentation example + normalize root param

f3ltron:
Hello I am coming from vuepress it's refreshing to see vitepress. Simpler to read and smaller code base. I like it.

Here just a little example how to use vitepress. Maybe it can be the start of its own documentation too.

I have corrected a ts errors and normalize the root params because:

`vitepress dev folderName` it broked for me with  `Error: Cannot find module 'documentation/.vitepress/config.js'`. He is looking for an absolute path as you do with `process.cwd()`.

I am still reading the code base but it look really easy to understand. More than vuepress in my opinion.

### DIFF
--- a/bin/vitepress.js
+++ b/bin/vitepress.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 const chalk = require('chalk')
 const argv = require('minimist')(process.argv.slice(2))
+const { join } = require('path')
 
 console.log(chalk.cyan(`vitepress v${require('../package.json').version}`))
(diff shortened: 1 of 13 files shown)


jrappen:
Thanks for this.

Maybe it would make sense to break lines at the end of sentences in markdown paragraphs for easier readability of (future) changes.

f3ltron:
I will have a look later on this. Just want to be sure it's usefull what i am doing on it before continuing going deeper.

yyx990803:
Thanks for the PR - this project is still undergoing a lot of changes and it's not the best time for public contribution yet.

PR #18, opened 2020-05-25
fix: missing deps and build

posva:
- added missing types for build
- added missing files to the `dist` folder when building


yyx990803:
I turned off notification and didn't see this - this should've been fixed 😅

PR #20, opened 2020-05-29
fix: silent failure when port is used

dewfall123:
I am not sure it's good to add the code at bin/vitepress.js.
Ditch if you don't like it.


yyx990803:
Thanks for the PR - this project is still undergoing a lot of changes and it's not the best time for public contribution yet.

PR #25, opened 2020-06-11
Initialize tests

phanan:
This PR initializes and adds the first tests for the `container` plugin. There's no CI configured yet as I'm not sure which one you're planning to use, but the tests should be green. 

Let me know if there's anything I should change. Also, if tests are not among the high priorities in the current phase, feel free to close this PR 😬.


package.json
```diff
@@ -75,15 +76,19 @@
...
     "@types/node": "^13.13.4",
     "@types/postcss-load-config": "^2.0.1",
     "chokidar": "^3.4.0",
     "conventional-changelog-cli": "^2.0.31",
+    "jest": "^26.0.1",
+    "jest-each": "^26.0.1",
     "lint-staged": "^10.2.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.0.5",
+    "ts-jest": "^25.5.1",
```
yyx990803, line 91: `ts-jest` also has `26.x` now

package.json
```diff
@@ -75,15 +76,19 @@
   },
   "devDependencies": {
     "@types/fs-extra": "^8.1.0",
+    "@types/jest": "^25.2.2",
```
yyx990803, line 79: This has `26.0` now

phanan:
@yyx990803 Done!

yyx990803:
Seems like the lock file keeps getting conflicts. I think it's probably easier to just remove it from the PR.

phanan:
Sure let me take a look 😬

phanan:
@yyx990803 Done! I'm not sure how to remove a file from PR, so I checked out the `master` version instead. Hope it works.

yyx990803:
Sorry I just reverted this. I saw the conflict was resolved and merged without thinking too much.
I pulled the changes locally and had a better look at it and felt this might not be the best way to go about the tests:

1. It would probably be easier to use jest snapshots instead of manual files. I also prefer separated test cases with inline snapshots.

2. For TS projects I've found it better to separate src files and tests because (1) this avoid test files being accidentally included in `dist`; (2) makes it easier to `git grep` inside `src` (I do this very often and test files messes the results).

And since most of the markdown is just copied from VuePress, they should be rather solid. For now I think tests should focus on high level e2e tests that test default theme features and builds.

phanan:
Oh sure. This was more or less only a blatant copy/port from VuePress—not much thinking was put into it TBH, since I was in the assumption of "VuePress ~ VitePress." I totally agree with your points now that _I_ think about it as well 😬.

PR #44, opened 2020-07-03
fix: add the missing argument for watcher.send

antfu:
Partially fix for #47. PR https://github.com/vitejs/vite/pull/499 in vite should be merged together to solve the build issue.


yyx990803:
This should no longer be needed with latest version of vite.

PR #49, opened 2020-07-05
feat: add multi sidebar support (#38)

### LINKED ISSUE #38: Multi Sidebar
Ref: https://vuepress.vuejs.org/theme/default-theme-config.html#multiple-sidebars

Code in question: https://github.com/vuejs/vitepress/blob/master/src/client/theme-default/components/SideBar.ts#L117

kiaking:
close #38 

This PR adds multi sidebar support. It does only support nesting array of objects as currently that is the only format supported by the array sidebar.

```js
{
  sidebar: {
    '/guide/': [
      { text: 'Getting Started', link: '/guide/getting-started' },
      ...
    ],

    '/guide/nested/': [
      { text: 'Nested Index', link: '/guide/nested/' },
      ...
    ],

    // Fallback supported too.
    '/': [
      { text: 'Index', link: '/' },
      { text: 'Root', link: '/root' }
    ]
  }
}
```

- `children` property in each object works as well.
- Active state of Navbar item is not working correctly. It needs to compare correct path but thinking to create another PR for it.
- URL without trailing slash acts differently than VuePress. Currently, if we access `/guide`, it renders `/guide.html`. If we access `/guide/`, it renders `/guide/index.html`. I'm leaving it as is for now but maybe we should have follow up PR for it...?

Screen Shot 2020-07-05 at 19 49 01

### DIFF
--- a/src/client/theme-default/utils.ts
+++ b/src/client/theme-default/utils.ts
@@ -27,7 +27,7 @@ export function normalize(path: string): string {
  * path is `/guide/getting-started.html`, this method will return `/guide/`.
  * Always with a trailing slash.
  */
-export function resolvePath(path: string): string {
+export function getPathDirName(path: string): string {
   const segments = path.split('/')
 
   if (segments[segments.length - 1]) {
--- a/src/client/theme-default/utils.ts
+++ b/src/client/theme-default/utils.ts
@@ -21,3 +21,22 @@ export function isActive(route: Route, path?: string): boolean {
 export function normalize(path: string): string {
   return decodeURI(path).replace(hashRE, '').replace(extRE, '')
 }
+
+/**
+ * get the path without filename (the last segment). for example, if the given
+ * path is `/guide/getting-started.html`, this method will return `/guide/`.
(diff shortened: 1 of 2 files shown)


src/client/theme-default/utils.ts
```diff
@@ -21,3 +21,22 @@ export function isActive(route: Route, path?: string): boolean {
 export function normalize(path: string): string {
   return decodeURI(path).replace(hashRE, '').replace(extRE, '')
 }
+
+/**
+ * get the path without filename (the last segment). for example, if the given
+ * path is `/guide/getting-started.html`, this method will return `/guide/`.
+ * Always with a trailing slash.
+ */
+export function resolvePath(path: string): string {
```
antfu, line 30: I think `resolvePath` is a bit unclear here, would it better to be `getPathDirname` or something?

kiaking: Good point! Let me change this one

kiaking: Done 🙌

PR #55, opened 2020-07-20
feat: add git repo link and edit links

### LINKED ISSUE #52: Git repository and Edit Links
Ref: https://vuepress.vuejs.org/theme/default-theme-config.html#git-repository-and-edit-links

sin-tanaka:
Resolves #52 

- [x] add to navbar
- [x] add edit text to footer
- [x] This is built on top of #46, should be merged after it.

config file: https://github.com/sin-tanaka/vitepress-playground/blob/master/docs/.vitepress/config.js

スクリーンショット 2020-07-23 20 53 50

with locales

スクリーンショット 2020-07-23 20 54 08


kiaking:
@sin-tanaka Hi! So sorry for the late reply. Thanks for the PR! Could you:

1. Align configs with [current VuePress config](https://vuepress.vuejs.org/theme/default-theme-config.html#git-repository-and-edit-links)? It doesn't need to support all of the items, but it would be nice if we can config it as below.

```js
module.exports = {
  themeConfig: {
    repo: 'vuejs/vuex',
    docsDir: 'docs',
    docsBranch: 'dev',
    editLinkText: 'Edit me!!'
  }
}
```

2. Could you rebase the branch to the latest? #46 is merged, so it's much easier to read the changes 😅

sin-tanaka:
@kiaking Thanks for the review!

1. The format of the config is as follows: 

https://github.com/vuejs/vitepress/blob/master/src/client/theme-default/config.ts#L7
```ts
  export interface Config {
    logo?
    nav?: NavItem[] | false
    sidebar?: SideBarConfig | MultiSideBarConfig
    search?
    editLink?: EditLinkConfig | false
    lastUpdated?
    prevLink?
    nextLink?
  }
```

Would you prefer to be able to set it to the first level of themeConfig, just like VuePress?

2. rebased!

kiaking:
@sin-tanaka Thanks for the rebase 🙌 

> Would you prefer to be able to set it to the first level of themeConfig, just like VuePress?

Yeah I think that would be better. For example having configs as close as VuePress will make migration from VuePress to VitePress a lot easier 👍  Would you mind doing that 😃 ?

posva:
@kiaking I changed the edit config but then realized it already had a different structure than before. So I'm wondering if I should revert those changes to keep the original new editconfig type

kiaking:
@posva Nice! I think the current config is desired one, since it matches the VuePress config. So I would say we should keep it this way 👍

src/client/theme-default/components/NavBarLinks.ts
```diff
@@ -10,13 +15,46 @@ export default {
...
+    const siteData = useSiteData()
+    const repoInfo = computed(() => {
+      const theme = siteData.value.themeConfig as DefaultTheme.Config
+      // return theme.editLinks
+      //   ? {
+      //       repo: theme.docsRepo || theme.repo,
+      //       label: theme.repoLabel
+      //     }
+      //   : null
+
```
posva, line 28:
```suggestion
```

forgot this one

src/node/markdownToVue.ts
```diff
@@ -43,6 +43,7 @@ export function createMarkdownToVueRenderFn(
       title: inferTitle(frontmatter, content),
       frontmatter,
       headers: data.headers,
+      relativePath: file.replace(/\\/g, '/'),
```
posva, line 46: @sin-tanaka What does this represent? Why are you replacing `/` with `/`? is it for Windows? I saw it in the vuepress code base too

PR #56, opened 2020-07-25
feat: add prev/next links

### LINKED ISSUE #43: Prev / Next links
Ref: https://vuepress.vuejs.org/theme/default-theme-config.html#prev-next-links

This one may be tricky since VitePress doesn't include the whole site's page metadata (for bundle size considerations). We may need to extract this information at build time and include it directly in each individual page's metadata (`$page`)

Spice-Z:
Resolve #43 

config file is [this](https://github.com/Spice-Z/vite-local-playground/blob/master/docs/.vitepress/config.js)!

スクリーンショット 2020-07-26 14 06 50

prev/next data generated from siteData's sidebar config are injected to pageData.
When `themeConfig.prevLinks` or`themeConfig.prevLinks` is explicitly set false, links are not shown.

Not supported frontmatter's customize yet

### DIFF
--- a/src/client/theme-default/components/NextAndPrevLinks.ts
+++ b/src/client/theme-default/components/NextAndPrevLinks.ts
@@ -1,0 +1,22 @@
+import { defineComponent, computed } from 'vue'
+import { usePageData } from 'vitepress'
+
+export default defineComponent({
+  setup() {
+    const pageData = usePageData()
+    const next = computed(() => {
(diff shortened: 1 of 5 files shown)


antfu:
LGTM 👍

Spice-Z:
@antfu 
Thanks!
I'll make another PR to overwrite prev/next links with page's frontmatter setting.

PR #59, opened 2020-08-01
feat: support config alias

### LINKED ISSUE #16: Import a file that's not in the root directory
**Is your feature request related to a problem? Please describe.**

Currently, I have a directory structure like :
```
project/
├── dist/
├── docs/
│   ├── .vitepress/
│   └── index.md
├── src/
│   └── components/
├── package.json
└── ...
```

The `docs` directory contains a vitepress app.

If I'm not mistaken, that's not possible to import a file that's not in the docs directory (a `.vue` file from `src` directory for example).

**Describe the solution you'd like**

I don't know what's the best solution to this problem. <s>But if we can add a custom vite resolver, it should work I guess ?

Moreover, it would also be necessary to expose the vite watcher instance to watch these resolved files.</s>

Since the recent Vite features, we can alias a path to a fs directory, but in Vitepress, we can't override the vite configuration to add our aliases.

em2046:
Resolve #16

### DIFF
--- a/src/node/build/bundle.ts
+++ b/src/node/build/bundle.ts
@@ -3,7 +3,7 @@ import slash from 'slash'
 import fs from 'fs-extra'
 import { APP_PATH, createResolver, SITE_DATA_REQUEST_PATH } from '../resolver'
 import { BuildOptions } from './build'
-import { SiteConfig } from '../config'
+import { resolveUserConfig, SiteConfig } from '../config'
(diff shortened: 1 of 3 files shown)


kiaking:
@antfu Could you take a look at this one? Want to make sure everything align with latest Vite 😳

antfu (approved):
LGTM. /cc @underfin

PR #61, opened 2020-08-02
feat: overwrite prev/next link

Spice-Z:
This is built on top of https://github.com/vuejs/vitepress/pull/56, should be merged after it.
Enable to explicitly overwrite or disable prev/next link for individual pages setting like this
```
---
prev: false
next: /guide/getting-started
---
```

### DIFF
--- a/src/client/theme-default/components/NextAndPrevLinks.ts
+++ b/src/client/theme-default/components/NextAndPrevLinks.ts
@@ -1,0 +1,22 @@
+import { defineComponent, computed } from 'vue'
+import { usePageData } from 'vitepress'
+
+export default defineComponent({
+  setup() {
+    const pageData = usePageData()
+    const next = computed(() => {
+      return pageData.value.next
+    })
+    const prev = computed(() => {
+      return pageData.value.prev
+    })
+    const hasLinks = computed(() => {
+      return !!next || !!prev
+    })
+    return {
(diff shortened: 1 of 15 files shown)


kiaking:
@Spice-Z Hi! Could you rebase this one to the latest since #56 is merged now 🙏

Spice-Z:
@kiaking
I merged latest change to this one!

src/client/theme-default/components/NextAndPrevLinks.ts
```diff
@@ -1,13 +1,43 @@
 import { defineComponent, computed } from 'vue'
 import { usePageData } from 'vitepress'
+import { useSiteData } from 'vitepress'
```
kiaking, line 3: Can we merge this one with the above import?

```suggestion
import { usePageData, useSiteData } from 'vitepress'
```

kiaking (changes requested):
Thanks for the rebase! Just added 1 comment, please take a look.

src/client/theme-default/components/NextAndPrevLinks.ts
Spice-Z: Thank you!
Merged!!

kiaking (approved):
Nice! Thanks you so much for this 🙌

PR #67, opened 2020-08-08
fix: Don’t prevent the browser’s default behavior for modifier keys

yuheiy:
The router seems to prevent the browser’s default behavior for modifier keys on link clicks. This PR changes the browser’s default behavior to take precedence when users are using modifier keys.

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -98,6 +98,7 @@ export function createRouter(
           // only intercept inbound links
           if (
             target !== `_blank` &&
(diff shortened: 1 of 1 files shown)


kiaking (approved):
Nice! LGTM 👍

kiaking:
Hi thank you so much for the PR! We went and merged #69 which is the exact same PR with this one, but slightly different code stylings. Sorry It took long before looking into this, but we appreciate so much!

PR #69, opened 2020-08-22
fix(router): allow open new tab with ctrl + click

pikax:
Let the browser open the link if shift/ctrl/meta key is pressed

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -98,6 +98,9 @@ export function createRouter(
           // only intercept inbound links
           if (
             !e.ctrlKey &&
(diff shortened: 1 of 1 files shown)


kiaking:
Thanks! This is duplicate of #67. Code is almost identical, with small styling difference 😅  Let's Evan decide which style he likes more

antfu:
For the styling, I would say I like one over 67 as this is more straightforward with fewer types of operators 😁.

kiaking:
Ah, fewer types operators sounds yummy... OK let's merge this one 👍

PR #73, opened 2020-08-27
fix(sidebar): fix sidebar when you open a nested link

pikax:
Note this will make an assumption that the first `config` key is the default sidebar.

Maybe we need to change to `/` instead of the "first" key

### DIFF
--- a/src/client/theme-default/components/SideBar.ts
+++ b/src/client/theme-default/components/SideBar.ts
@@ -129,7 +129,8 @@ function resolveMultiSidebar(
   headers: Header[],
   depth: number
 ): ResolvedSidebar {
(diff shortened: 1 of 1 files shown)


kiaking:
@pikax Sorry what is the problem here? What happens when we click a nested link?

pikax:
This "fixes" the sidebar disappearing when you navigate to a child.

I'm not too happy with this fix, maybe do this somewhere else?

kiaking:
Ohhh now I see. Make total sense. Now I know why it was failing in VTU but not in Vuex and Router (because they don't have "nested" route). I think this place is fine for now 👍  Nice job!

PR #74, opened 2020-08-27
feat(sidebar): use base when creating link

pikax:
create a link based on the absolute url, this might cause problems if the children start with '/'


kiaking:
Yeah, I think we should handle links starting with `/`, because I think it's easier for users to add config like,

```js
{ text: 'Guide', link: '/guide/' }
```

Although, I think we could just enforce users to not add trailing slash on base url 😳

PR #75, opened 2020-08-30
feat: responsive sidebar

Spice-Z:
Resolve a part of #39 , this is the first step !
Just hiding sidebar in mobile view, and switching open / hide correspond with clicking button.

resp

### DIFF
--- a/src/client/theme-default/styles/vars.css
+++ b/src/client/theme-default/styles/vars.css
@@ -6,4 +6,9 @@
   --text-color-light: #476582;
   --code-bg-color: #282c34;
   --accent-color: #3eaf7c;
+
+  /* responsive breakpoints */
+  /* --mq-narrow: 959px; */
+  /* --mq-mobile: 719px; */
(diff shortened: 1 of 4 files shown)


src/client/theme-default/styles/vars.css
```diff
@@ -6,4 +6,9 @@
   --text-color-light: #476582;
   --code-bg-color: #282c34;
   --accent-color: #3eaf7c;
+
+  /* responsive breakpoints */
+  /* --mq-narrow: 959px; */
+  /* --mq-mobile: 719px; */
+  /* --mq-mobile-narrow: 419px; */
 }
```
Spice-Z, line 14: I cannot use CSS variables in media query.
This is just memos.

We should manage to share break points variables using other solution.

kiaking (approved):
LGTM! We should be also able to close the sidebar when clicking outside of the sidebar, though we could do that later and I think we can merge this one at this point.

antfu (approved):

PR #77, opened 2020-09-04
feat: add charset and viewport meta tags (#68)

### LINKED ISSUE #68: Add <meta charset="utf-8"> to rendered html.
When I want to deploy my website on CloudBase, I see messy code on it for Chinese. I think it's because CloudBase don't have `content-type: text/html; charset: utf-8` on it's response header, and vitepress don't have `<meta charset="utf-8">` on it's rendered html head.

kiaking:
close #68

This PR adds `charset` and `viewport` meta tag to the `head`.

@antfu For the `src/client/app/composables/head.ts` change, I'm being bit lazy here but I'm kinda thinking this should be done in Vite rather than in VitePress. Like it would be nice if we can pass these meta tag settings through Vite config, or maybe VitePress could include these tags by default. What do you think? Or is is it possible already?

### DIFF
--- a/src/client/app/composables/head.ts
+++ b/src/client/app/composables/head.ts
@@ -33,6 +33,14 @@ export function useUpdateHead(
     const pageTitle = pageData && pageData.title
     document.title = (pageTitle ? pageTitle + ` | ` : ``) + siteData.title
     updateHeadTags([
+      ['meta', { charset: 'utf-8' }],
+      [
+        'meta',
+        {
+          name: 'viewport',
+          content: 'width=device-width,initial-scale=1'
+        }
+      ],
       [
         'meta',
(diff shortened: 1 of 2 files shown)


antfu:
There is a limitation of current implementation of Vite that middlewares cannot directly transform index.html. Planned to fix it. After that, probably a simple vite-plugin-meta or something would do ;)

kiaking:
Ah good to know, thanks! 👍  Then I guess we can look forward to it, and while then we could just handle it the way this PR is. Do you find any concern with the implementation?

antfu (approved):
LGTM 👍

kiaking:
Thanks a lot 🙌

Issue #76, opened 2020-09-03
Close the sidebar when clicking outside of the sidebar


kiaking:
**Is your feature request related to a problem? Please describe.**
Currently, when sidebar is open, we can only close it by clicking the hamburger menu. We should also make it closable by clicking outside of the sidebar.

**Describe the solution you'd like**
We should also make it closable by clicking outside of the sidebar.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #78, opened 2020-09-05
feat: close sidebsr when click outside

Spice-Z:
Resolve https://github.com/vuejs/vitepress/issues/76
The solution is same to VuePress.

Sep-05-2020 20-47-41

And...
Sidebar should be closed when sidebar's contents are clicked.
I'll make another PR about this!


kiaking:
Niiice! Thanks a million!

PR #85, opened 2020-09-13
fix(prev/next link): fix logic to correspond with sidebar config change

Spice-Z:
Now, all of prev/next link are not shown without each pages's frontmatter setting.
From https://github.com/vuejs/vitepress/pull/74, sidebar config path was changed.
Prev/Next link is generated from sidebar config, so some changes are needed.

### DIFF
--- a/src/node/server.ts
+++ b/src/node/server.ts
@@ -145,9 +145,14 @@ function getNextAndPrev(themeConfig: any, pagePath: string) {
       if (!sidebarItem.children) {
         return
       }
-      sidebarItem.children.forEach((candidate: any) => {
-        candidates.push(candidate)
-      })
+      sidebarItem.children.forEach(
+        (candidate: { text: string; link: string }) => {
+          candidates.push({
+            text: candidate.text,
+            link: convertToAbsolutePath(candidate.link)
+          })
+        }
+      )
     })
   })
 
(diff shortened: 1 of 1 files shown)


posva:
Can you share a reproduction of the problem?

Spice-Z:
@posva 
Reproduction
1. Set vitepress sidebar config children with link not starting from '/',like this
```
children: [
            { text: 'Index', link: 'guide/index' },
            { text: 'Getting Started', link: 'guide/getting-started' },
            { text: 'Guide Second', link: 'guide/guide-second' },
          ]
```
2. Show page that should content Prev/Next link.
3. Prev/Next link are not shown!

This occurs because `pagePath` always starting from `/`, but sidebar config children are not starting from '/'.
When searching pagePath in sidebar, we should consider this difference.

(BTW, thank you to your Vue.js Global stage! I enjoyed it yesterday! )

posva:
Thank you! I'm glad you liked my talk 😄 

I see what you mean right now. I don't think it's a good idea to write relative links like that in children because a lot of people do not realize that a relative URL takes away the last section of the URL to replace it, some will think it appends to the parent, others would think it changes based on the current location. So I think it's better to use absolute URLs on the sidebar: `/guide/index`

I think there is also a chance here for us to display some messages in the console to warn users about implicit relative URLs and make them change it to absolute URLs instead

We also need to refactor a little part of the server regarding consuming the theme, so I think it's better to hold on on changes to this for the moment! Thank you!

Issue #40, opened 2020-07-02
Search box


yyx990803:
Ref: https://vuepress.vuejs.org/theme/default-theme-config.html#search-box

appbest:
国庆期间，编写全站搜索组件与 代码在线运行组件
[搜索文档模板](https://github.com/appbest/QuickDocs)

fimbault:
Would be nice to use https://www.meilisearch.com

meganmeiyang:
hi should you update the doc as well: https://vitepress.vuejs.org/guide/theme-search.html
thanks ~

Eviwang:
```
 themeConfig: {
    search: true,
    searchMaxSuggestions: 10
  }
```
not work.

brc-dd:
@Eviwang This was for older versions. Refer https://github.com/vuejs/vitepress/issues/1078 and https://github.com/vuejs/vitepress/issues/670 for updated stuff.

PR #86, opened 2020-09-14
feat(default-theme): native support for carbon ads

posva:
I also added the existing slots sidebar-top/bottom and page-top/bottom because I used them to add the ads components. To make them work and test them, you need to add this config to the `themeConfig`:

```js
    carbonAds: {
      carbon: 'CEBICK3I',
      custom: 'CEBICK3M',
      placement: 'routervuejsorg',
    },
```

### DIFF
--- a/src/client/theme-default/Layout.vue
+++ b/src/client/theme-default/Layout.vue
@@ -34,10 +34,24 @@
     <main v-else>
       <Page>
         <template #top>
-          <slot name="page-top" />
+          <slot name="page-top">
+            <CarbonAds
+              v-if="$site.themeConfig.carbonAds"
+              :key="'carbon' + $page.path"
+              :code="$site.themeConfig.carbonAds.carbon"
+              :placement="$site.themeConfig.carbonAds.placement"
+            />
+          </slot>
         </template>
         <template #bottom>
-          <slot name="page-bottom" />
+          <slot name="page-bottom">
+            <BuySellAds
+              v-if="$site.themeConfig.carbonAds"
+              :key="'custom' + $page.path"
+              :code="$site.themeConfig.carbonAds.custom"
+              :placement="$site.themeConfig.carbonAds.placement"
+            />
+          </slot>
         </template>
       </Page>
     </main>
@@ -53,6 +67,8 @@ import ToggleSideBarButton from './components/ToggleSideBarButton.vue'
 import SideBar from './components/SideBar.vue'
 import Page from './components/Page.vue'
 import { useRoute, useSiteData, useSiteDataByRoute } from 'vitepress'
+import CarbonAds from './components/CarbonAds.vue'
+import BuySellAds from './components/BuySellAds.vue'
 
 const route = useRoute()
 const siteData = useSiteData()
--- a/src/client/theme-default/Layout.vue
+++ b/src/client/theme-default/Layout.vue
@@ -34,7 +34,7 @@
     <main v-else>
       <Page>
         <template #top>
-          <slot name="page-top">
+          <slot name="page-top-ads">
             <CarbonAds
               v-if="$site.themeConfig.carbonAds"
               :key="'carbon' + $page.path"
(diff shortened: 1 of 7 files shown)


kiaking:
I kinda like the idea of having Google Analytics option, since usually everybody wants that. But not sure about the ads... Because while Vue's official docs uses Carbon, it really depends on locales what the primary ads formats are. I think for now, ads should be configured for each repo (Router/Vuex/etc.) using slots 🤔 What do you think? (Again, I'm all for Google Analytics though).

posva:
I think we should still add the slots to the default theme and then add configs for ads to the vue theme since we use it in every official vue lib documentation. We could also handle something per locale, and it could also be a plugin instead.

Adding analytics should also be in a plugin like in vuepress.

But I don't think we have the plugin architecture yet

kiaking:
Ah, that make sense. If we were to implement plugins feature (and I think we kinda need it), it make sense now to include these and pull it out once we have the plugin feature 👍

posva:
yeah, let's hold on to this for the moment. I will refactor it once there is some kind of plugin architecture in place

kiaking:
@posva I think at the moment, VitePress doesn't look like to have any strong plugin feature implemented in future. I think we should have this merged in to the core directly.

posva:
@kiaking I think it's fine to wait for the moment! Better have a plugin architecture built that allows adding such things.
Do you want to add this feature for Vuex or something? The current PR is not up to date with what exists in vue router next either. I would have to update it. I even plan on changing what I currently have on Vue Router docs to allocate spots for sponsors.

kiaking:
> I think it's fine to wait for the moment! Better have a plugin architecture built that allows adding such things.

No I mean the discussion we are having right now, there might be no plugin feature for VitePress at all. So we might be waiting for nothing... 🤔 

> Do you want to add this feature for Vuex or something?

This is another story but yes I would love to. And for that, I can of course wait!

posva:
I will give this a go and refactor next week and make it look like the current version on vue router next docs: https://next.router.vuejs.org/guide/essentials/navigation.html#navigate-to-a-different-location which was inspired by vuejs.org

kiaking:
Yeah great idea! Looking forward to it!

posva:
Here are the changes needed. `The overflow-x: auto` fixes this:

Screen Shot 2020-11-23 at 12 19 29

into

Screen Shot 2020-11-23 at 12 43 52

posva:
I think we should add by default the algolia search too, but I will do that in a separat PR. What do you think @kiaking ?

kiaking:
Yes we should definitely have algolia integration too, the issue #40 🙌

posva:
ah yeah, that's a different thing though 😆 : simple search out of the box based on titles. Algolia still needs you to request ids at https://docsearch.algolia.com/

src/client/theme-default/components/CarbonAds.vue
```diff
@@ -0,0 +1,95 @@
...
+</script>
+
+<style>
+.carbon-ads {
+  min-height: 102px;
+  font-size: 0.75rem;
+
+  width: 130px;
+  position: fixed;
+  z-index: 19;
```
posva, line 43:
```suggestion
  z-index: 1;
```

src/client/theme-default/components/CarbonAds.vue
```diff
@@ -0,0 +1,95 @@
...
+@media screen and (max-width: 425px) {
+  .carbon-ads {
+    width: auto;
+    right: 0;
+  }
+}
+
+@media screen and (max-width: 1300px) {
+  .carbon-ads {
+    z-index: 1;
```
posva, line 58:
```suggestion
```

src/client/theme-default/Layout.vue
```diff
@@ -34,10 +34,24 @@
     <main v-else>
       <Page>
         <template #top>
-          <slot name="page-top" />
+          <slot name="page-top">
```
kiaking, line 37: Maybe we should rename this slot to `page-top-ads` or something? Since people might want to add something other than ads in this section, while preserving the ads 😳 

Same goes for `page-bottom` as well (page-bottom-ads).

posva: Sure! Let me see if we can have both slots without one being nested into each other, it should be good

posva: Pushed the changes, let me know what you think. The preview should also work to make sure it works. I tested on Vuex repo

kiaking: Niiice! Looking good I think 🙌

posva:
This should be ready for review!

kiaking:
Here we go!

PR #87, opened 2020-09-15
feat: allow passing googleAnalytics option

posva:
Allows passing `googleAnalytics: '<id>'` to the `themeConfig` option on the default theme.

I'm starting to think I should group this PR, #86 and another PR for Algolia search into a different vue theme. What do you think?

### DIFF
--- a/src/client/theme-default/index.ts
+++ b/src/client/theme-default/index.ts
@@ -6,10 +6,21 @@ import './styles/custom-blocks.css'
 import Layout from './Layout.vue'
 import NotFound from './NotFound.vue'
 import { Theme } from '../app/theme'
+import { installGoogleAnalytics } from './ga'
 
 const theme: Theme = {
   Layout,
-  NotFound
+  NotFound,
+  enhanceApp({ router, siteData }) {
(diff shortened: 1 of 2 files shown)


src/client/theme-default/index.ts
```diff
@@ -6,10 +6,21 @@ import './styles/custom-blocks.css'
...
 
 const theme: Theme = {
   Layout,
-  NotFound
+  NotFound,
+  enhanceApp({ router, siteData }) {
+    const { googleAnalytics: GA_ID } = siteData.value.themeConfig
+    if (
+      GA_ID &&
+      process.env.NODE_ENV !== 'production' &&
```
yangmingshan, line 18: Did you mean `process.env.NODE_ENV === 'production'`?

posva, line 18:
```suggestion
      process.env.NODE_ENV === 'production' &&
```

posva: Yes, thanks

kiaking:
I think having GA featurue is awesome! But since we've started Next development, let's do that there instead 👍

PR #90, opened 2020-09-17
feat: top and bottom slots for sidebar and page

posva:
This is the same api as vuepress. Useful for ads slots


src/client/theme-default/Layout.vue
```diff
@@ -5,22 +5,36 @@
...
+        </template>
+      </Page>
     </main>
   </div>
   <Debug />
 </template>
 
 <script>
-import { ref } from 'vue'
+import { ref, computed } from 'vue'
```
posva, line 37:
```suggestion
import { ref } from 'vue'
```

PR #50, opened 2020-07-05
feat: i18n support

antfu:
Resolves #41 

The only thing left is nav dropdown, that would be another PR.

Working with config: https://github.com/antfu/vitepress-playground/blob/i18n/docs/.vitepress/config.js

`/zh/`
image

Static build can be handled correctly as well

`/index.html`
```html
<html lang="en-US">
  <head>
    <title>VitePress Playground | VuePress</title>
    <meta name="description" content="Vue-powered Static Site Generator">
    <link rel="stylesheet" href="/_assets/style.55d0251c.css">
```

`/zh/index.html`
```html
<html lang="zh-CN">
  <head>
    <title>VitePress 游乐场 ：） | 静态网站生成器</title>
    <meta name="description" content="Vue 驱动的静态网站生成器">
    <link rel="stylesheet" href="/_assets/style.55d0251c.css">
```

PR #91, opened 2020-09-19
feat(i18n): nav dropdown language selector

Spice-Z:
At https://github.com/vuejs/vitepress/pull/50, @antfu implemented i18n support, and said.

>The only thing left is nav dropdown, that would be another PR.

I created that!

Oct-01-2020 20-42-52

Working config is like this

```
...
themeConfig: {
 locales: {
    '/': {
      label: 'English',
      selectText: 'Languages',
    },
    '/zh/': {
      label: '中文',
      selectText: '选择语言',
    },
    '/ja/': {
      label: '日本語',
      selectText: '言語,
    },
  },
}
```

(To set config about site title, description, and lang, you have to use global locale config. Check https://github.com/vuejs/vitepress/pull/50)

### DIFF
--- a/src/client/theme-default/components/NavBarLinks.ts
+++ b/src/client/theme-default/components/NavBarLinks.ts
@@ -44,6 +44,9 @@ export default {
 
     const localeCandidates = computed(() => {
       const locales = siteData.value.themeConfig.locales
+      if (!locales) {
+        return null
+      }
       const localeKeys = Object.keys(locales)
       if (localeKeys.length <= 1) {
         return null
--- a/src/client/theme-default/components/NavBarLinks.ts
+++ b/src/client/theme-default/components/NavBarLinks.ts
@@ -1,5 +1,5 @@
 import { computed } from 'vue'
-import { useSiteData, useSiteDataByRoute } from 'vitepress'
+import { useSiteData, useSiteDataByRoute, useRoute } from 'vitepress'
 import NavBarLink from './NavBarLink.vue'
 import NavDropdownLink from './NavDropdownLink.vue'
 import { DefaultTheme } from '../config'
@@ -17,6 +17,7 @@ export default {
   setup() {
     const siteDataByRoute = useSiteDataByRoute()
     const siteData = useSiteData()
+    const route = useRoute()
     const repoInfo = computed(() => {
       const theme = siteData.value.themeConfig as DefaultTheme.Config
       const repo = theme.docsRepo || theme.repo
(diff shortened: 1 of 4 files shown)


src/client/theme-default/components/NavBarLinks.ts
```diff
@@ -40,14 +41,45 @@ export default {
...
+      const candidates = localeKeys.map((v) => {
+        return {
+          text:
+            siteData.value.locales[v].label || siteData.value.locales[v].lang,
+          link: `${v}${currentContentPath}`
+        }
+      })
+      return {
+        // TODO i18n text
+        text: 'Languages',
```
Spice-Z, line 69: This text also needs i18n

Spice-Z: DONE

kiaking (changes requested):
Thank you so much for the feature! Looking really good!

I have 2 requests.

Currently, VuePress locale config sits inside `themeConfig`. Would it be possible to move the new configs under there instead of `locales`?

```js
{
  themeConfig: {
    locales: {
      '/': {
        label: 'English',
        selectText: 'Languages',
        editLinkText: 'Edit this page on GitHub',
    },
    ...
  }
}
```

Ref: https://vuepress.vuejs.org/guide/i18n.html#default-theme-i18n-config

2. Can we move i18n dropdown before GitHub repo link at Nav?

src/client/theme-default/components/NavBarLinks.vue
```diff
@@ -7,6 +7,7 @@
       </template>
     </template>
     <NavBarLink v-if="repoInfo" :item="repoInfo" />
+    <NavDropdownLink v-if="localeCandidates" :item="localeCandidates" />
```
kiaking, line 10: Lets move this one above the GitHub repo nav 🙏

Spice-Z: Moved!

Spice-Z:
@kiaking 
I fixed both of your requests!

About first request, I didn't notice vitepress( and vuepress) has two kinds of locales in config...!
(`locales` is used as `Site Level i18n Config`, `themeConfig.locals` is used as `Theme i18n Config`.)

Now, this selector will use `themeConfig.locals`'s `label` and `selectText`

src/client/theme-default/components/NavBarLinks.ts
```diff
@@ -40,14 +41,49 @@ export default {
       }
       return null
     })
+
+    const localeCandidates = computed(() => {
+      const locales = siteData.value.themeConfig.locales
+      const localeKeys = Object.keys(locales)
```
kiaking, line 47: At here, we must check whether if the `locales` exists or not.

```js
const locales = siteData.value.themeConfig.locales
if (!locales) {
  return null
}
const localeKeys = Object.keys(locales)
```

If the config doesn't contain any locales, `Object.keys(locales)` will fail due to `locales` is `undefined` in that case.

kiaking (changes requested):
Thanks a lot! Just one more thing. I found a bug that this PR will not work on configs that does not contain any locales info. Please see the inline comment I made 🙏

Spice-Z:
@kiaking 
Thank you !
Sorry to late commit..., I fixed it!

I have to think of any every case of config, or have to write test code. 😅

kiaking (approved):
Great! Thanks a lot. Tested at my side and it is working well 🙌

PR #98, opened 2020-10-15
Add process.exit to stop build

silvandiepen:
For some reason my builds always kept running even though the 'done' message was shown. By adding process.exit() explicitly it was fix. 

This should not give any problems and will fix it there where necessary when done.


kiaking:
Thanks for the PR! This is fixed with VitePress 0.7.1 👍

PR #99, opened 2020-10-16
feat(client): add slot for a searchbar

posva:
I'm not sure how do we want to add this feature on vitepress yet, so I added a slot for the moment so we can implement the searchbox on userland. This is what it looks like on vue-router-next docs

Screen Shot 2020-10-16 at 17 17 54

### DIFF
--- a/src/client/theme-default/styles/layout.css
+++ b/src/client/theme-default/styles/layout.css
@@ -2,6 +2,10 @@ body {
   padding: 0;
   margin: 0;
   background-color: #fff;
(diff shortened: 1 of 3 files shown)


src/client/theme-default/styles/layout.css
```diff
@@ -2,17 +2,17 @@ body {
   padding: 0;
   margin: 0;
   background-color: #fff;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen,
```
posva, line 5: this is necessary to type the font of the search modal. Vuepress has the same thing

posva:
@knightly-bot build this

PR #116, opened 2020-11-05
feat: support lowercase url (#115)

### LINKED ISSUE #115: Can support lowercase url after vitepress build ?
## Summary

Some browsers/hostings change the urls to lowercase. Because of that, the `__VP_HASH_MAP__` can't find corresponding hash, it will be `undefined`.

## Motivation

I am using netlify, netlify will auto change url to `lowercase` ant this behavior can't opt out. See this [topic](https://community.netlify.com/t/my-url-paths-are-forced-into-lowercase/1659).

It makes me very uncomfortable. Because I am writing `composition-api` related libraries, so I naturally call the file name `useXXX` .... , such as `useClickAway` , This is work very well in dev, but once deployed, it's can't working, This problem was only discovered after I deployed ...

I don't want to rename my markdown file, i know i can use [command](https://stackoverflow.com/questions/152514/how-do-i-rename-all-folders-and-files-to-lowercase-on-linux) to replace, but I think we can support this feature .

vuepress look like support this feature, i find this [pr](https://github.com/vuejs/vuepress/pull/898)

## Examples

I used `vitepress` which I modified myself, It will both working now .

- https://vueposu.netlify.app/dom/useClickAway.html
- https://vueposu.netlify.app/dom/useclickaway.html

You can click other link and refresh page, you will notice the url is converted to lowercase.

yuxino:
closed #115

### DIFF
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -32,6 +32,12 @@ export async function build(buildOptions: BuildOptions = {}) {
       (chunk) => chunk.type === 'asset' && chunk.fileName.endsWith('.css')
     ) as OutputAsset
 
+    Object.keys(pageToHashMap).forEach((key) => {
+      if (!pageToHashMap[key.toLocaleLowerCase()]) {
+        pageToHashMap[key.toLocaleLowerCase()] = pageToHashMap[key]
+      }
+    })
+
     // We embed the hash map string into each page directly so that it doesn't
     // alter the main chunk's hash on every build. It's also embedded as a
     // string and JSON.parsed from the client because it's faster than embedding


kiaking:
OK, this is not a complete fix. It does resolve the file, but it gives "Hydration completed but contains mismatches." error when accessing originally camel cased URL in lower case url. You can see the error in https://vueposu.netlify.app/dom/useclickaway.html.

I think we need to dig in a bit deeper to this issue. Maybe we should just output all bundles, assets in lowercase, and request lowercased URL from the client side...?

yuxino:
> Maybe we should just output all bundles, assets in lowercase, and request lowercased URL from the client side...?

I researched some websites, some of them ignore case and some don’t ignore case 

ignore case:

- [https://stackoverflow.com/questions/7996919/should-url-be-case-sensitive](https://stackoverflow.com/questions/7996919/should-url-be-case-sensitive) - work
- [https://stackoverflow.com/questions/7996919/Should-URL-Be-case-sensitive](https://stackoverflow.com/questions/7996919/Should-URL-Be-case-sensitive) - work

not ignore case:

- [https://en.wikipedia.org/wiki/Travel_restrictions_related_to_the_COVID-19_pandemic](https://en.wikipedia.org/wiki/Travel_restrictions_related_to_the_COVID-19_pandemic) - work
- [https://en.wikipedia.org/wiki/Travel_restrictions_related_to_the_COVID-19_PANdemic](https://en.wikipedia.org/wiki/Travel_restrictions_related_to_the_COVID-19_PANdemic) - not found
- [google.com/about.html](google.com/about.html) - work
- [google.com/ABOUT.html](google.com/ABOUT.html) - not found

I can't assert that `all change to lowercase` is good or bad. In most cases, I think `router part url` should be case-insensitive, and users generally don’t enter the website manually ..

@kiaking I think maybe we can add an compile options like `urlCaseSensitive: boolean` to support `url case sensitive` this behavior

Related question [`should-url-be-case-sensitive`](https://stackoverflow.com/QUEStions/7996919/Should-url-be-case-sensitive)

kiaking:
> I can't assert that all change to lowercase is good or bad. In most cases, IMO, url should be case-insensitive

Hmmm, good point. Yeah I think so too. Making URL case-insensitive sounds like what we want.

> I think maybe we can add an compile options like urlCaseSensitive: boolean to support url case sensitive this behavior

Maybe. But first, we need to figure out how to do that in the first place. The hash map thing is unique to VitePress, and it doesn't exist in VuePress so, I think we need to come up with clean solution for this problem. And at the moment, I don't have ideas, so It would be nice if can get some good hints to tackle this issue 😅

posva:
If we set things to be case insensitive by default, when rendering static routes we must make sure that the server is also configured to handle them in a case insensitive fashion

kiaking:
Yeah, it's not super straight forward to implement this... 😳

PR #126, opened 2020-11-16
docs: add vitepress docs

kiaking:
This PR adds the docs for VitePress. It's not complete, but I wanted to start creating it since we get lots of questions, but don't have good place to put all those answers.

Also, this is really nice place to play around with VitePress features as well.


docs/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "private": true,
+  "scripts": {
+    "dev": "vitepress dev src",
+    "build": "vitepress build src"
+  },
+  "devDependencies": {
+    "vitepress": "^0.7.4"
```
yyx990803, line 8: Shouldn't we use the code in the repo itself to deploy the docs? I think that also serves as a form of e2e test.

yyx990803 (changes requested):

docs/package.json
kiaking: Oh yes I think we should! Just wasn't sure how...? I was looking into VuePress code and seems like it was doing this way... so...

Do you have any idea how we can do this 🤔 ?

docs/src/guide/customization.md
```diff
@@ -0,0 +1,51 @@
+# Customization
+
+You can develop your custom theme by adding the `.vitepress/theme/index.js` file.
+
+```bash
+.
+├─ docs
+│  ├─ .vuepress
```
ChrisShank, line 8:
```suggestion
│  ├─ .vitepress
```

kiaking: Oops, thanks!

docs/package.json
kiaking: @yyx990803 OK I think I fixed it. Not sure if this is the cleanest way to go. So now we have `docs/package.json` that only contains scripts. This is because if we reference `bin/vitepress.js` from root `package.json`, it tries to optimize node specific modules and crashes.

While we don't have a way to pass `optimizeDev` option to underlining Vite, I went this way.

PR #129, opened 2020-11-17
fix(client): show top part of scrollbar in sidebar

paduszym:
### Before fix:
image

### After fix:
image


kiaking:
Wonderful! Thanks a lot 👍

PR #131, opened 2020-11-19
chore: add undeclared deps

ChrisShank:


kiaking (commented):
Thanks! I think `chalk` is OK but not sure with `rollup`. `rollup` is installed with `vite`, and I think we want the version to be in sync.

Do we have any strong reason why we want to include undeclared deps?

ChrisShank:
@kiaking I was playing around with Yarn 2 and it's pretty strict about undeclared deps so Vitepress does not currently work with Yarn 2 (although there is a userland [workaround](https://next.yarnpkg.com/getting-started/migration#fix-dependencies-with-packageextensions)). Also I think in general it is good to be explicit about what deps we are using. Maybe `"rollup": "*"` would be better?

kiaking:
Ah I see. Yarn 2 makes sense. Thanks for the clarification. OK let's believe we live in the perfect world where semver is working without problem. It shouldn't cause any trouble 👍

yyx990803:
`rollup` should be moved to dev deps because `vitepress` only imports types from it.

PR #137, opened 2020-11-19
refactor: migrate default theme to use script-setup

yyx990803:
This is a PR containing my local changes migrating the default them to use `<script setup>` as proposed in https://github.com/vuejs/rfcs/pull/227.

Specifically I verified that the inline compile mode works as intended. One quirk discovered is that when using `<script setup lang="ts">` it is necessary to explicitly mark type imports with `import type { X } from '...'` otherwise it will be used as a value in the setup return statement. It should be technically possible to avoid this limitation with more sophisticated analysis but for now I don't think it's a deal breaker.

Note this PR isn't "working" yet and is here demonstrating the code simplifications that can be provided by the new RFC. Shipping it depends on the following:

- New release of `@vue/compiler-sfc` for the new syntax AND inline compile mode
- New releases of `vite` and `rollup-plugin-vue` with compatibility updates
- IDE support: I'm using https://github.com/johnsoncodehk/volar which supports the `<script setup>` syntax. In a few cases the binding analysis fails on destructuring declarations (https://github.com/johnsoncodehk/volar/issues/24)


kiaking:
Niiice! Watch out for the conflicts since I'm planning to adjust more on themes 🔥

I think import type issue is not because of <script setup>, I remember I had the same issue before so I thought it was a limitation for Vite 🤔 I agree it wouldn't be a deal breaker though.

yyx990803:
This is now ready. I would recommend enabling [Volar](https://github.com/johnsoncodehk/volar) instead of Vetur for now while working on this codebase.

kiaking:
@yyx990803 Niiice! Let me do the final checks on various example and merge this one 👍  I'll fix the conflicts if we get any.

src/client/theme-default/NotFound.vue
```diff
@@ -2,19 +2,19 @@
...
+<script setup lang="ts">
 const msgs = [
   `There's nothing here.`,
   `How did we get here?`,
   `That's a Four-Oh-Four.`,
   `Looks like we've got some broken links.`
 ]
 
-function getMsg() {
+const getMsg = () => {
```
posva, line 17: does this work without the change? I think having the `function` keyword here is valuable for readability

kiaking: It seemed it doesn't, we must declare function as `const` for it to be exposed. But let me double check 👀

kiaking: Oh sorry, it did work. I was mistaken by something 😅  Reverted the change!

tarkhil:
Suddenly, `<script setup="">` satisfies BOTH vue-mode and sfc-compiler.

PR #139, opened 2020-11-20
Ensure that the base is used for previous and next links

ElMassimo:
### Description 📖 

This pull request fixes next and previous links in sites with a non-default `base` configuration.

### Background 📜 

#130 recently moved the previous and next page link calculation to the client side.

Because links used in the sidebar don't include the base, rendering them directly causes them to have an incorrect URL.

### DIFF
--- a/src/client/theme-default/components/NextAndPrevLinks.vue
+++ b/src/client/theme-default/components/NextAndPrevLinks.vue
@@ -2,13 +2,13 @@
   <div v-if="hasLinks" class="next-and-prev-link">
     <div class="container">
       <div class="prev">
-        <a v-if="prev" class="link" :href="prev.link">
+        <a v-if="prev" class="link" :href="withBase(prev.link)">
           <ArrowLeft class="icon icon-prev" />
           <span class="text">{{ prev.text }}</span>
         </a>
       </div>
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/NextAndPrevLinks.vue
```diff
@@ -35,7 +36,8 @@ export default defineComponent({
     return {
       hasLinks,
       prev,
-      next
+      next,
+      withBase,
```
ElMassimo, line 40: Another option would be to use `withBase` in `getFlatSidebarLinksFromArray`:

```js
    if (item.link) {
      links.push({ text: item.text, link: withBase(item.link) })
    }
```

However, that would also require using `withBase` in `currentPath` for the comparison in `currentIndex` to work as expected.

Calling it in the component seemed consistent with how the logo is currently being rendered.

kiaking: Yeah, I would like to refactor the use of `withBase`, but for now it's being used everywhere so I think this is fine!

kiaking (approved):
Thanks a lot! Totally missed this one 👀 Great work!

PR #142, opened 2020-11-23
fix: code plugins styles

matias-capeletto:
Fixes font sizing and vertical padding styles for highlight lines and line numbers code plugins in the default theme

highlight-lines-style-issue

line-numbers-issue

I added two new vars: `--code-font-size` and `--code-vertical-padding`. The vertical padding is only modifying the plugin's styles after this PR because it requires a bit of refactoring to make the code padding dynamic and I am not sure what is the best way to do it.

After the fix:

code-after


kiaking:
Wonderful! Thanks a lot 🙌

PR #143, opened 2020-11-23
fix: mixed rgba colors for md code plugins

matias-capeletto:
Using `rgba(0,0,0,50%)` was working in dev but it was converted to an invalid color `rgba(0,0%,0%,50%)` when building so the highlighted lines were not visible 

invalid-color

Switching to a number for alpha (as it is in other rgba values in the css) fixes the issue.

I do not know if `rgba(0,0,0,50%)` is a valid color to be supported, I looked a bit and this conversion doesn't seem to be the responsibility of vitepress.

### DIFF
--- a/src/client/theme-default/styles/code.css
+++ b/src/client/theme-default/styles/code.css
@@ -70,7 +70,7 @@ div[class*='language-'] {
 }
 
 .highlight-lines .highlighted {
-  background-color: rgba(0, 0, 0, 66%);
+  background-color: rgba(0, 0, 0, 0.66);
 }
 
 /* Line numbers mode */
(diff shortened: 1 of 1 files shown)


kiaking:
@matias-capeletto Thanks a lot! I think merging #142 made conflicts. Could you resolve this one 🙏

matias-capeletto:
@kiaking should I do a PR porting the markdown extensions page from vuepress to vitepress docs so we can test these plugins? I think we could directly copy that page into vitepress docs.
https://vuepress.vuejs.org/guide/markdown.html

kiaking:
Oh yeah that would be great idea. Would you like to added it to this PR?

matias-capeletto:
Great, I'll look into it. If you do not mind, I prefer to send a new PR for the docs so we can close this one 👍

kiaking:
@matias-capeletto 

> If you do not mind, I prefer to send a new PR for the docs so we can close this one 👍

Got ya. I'll merge this one for now. Please open another PR for the docs 🙌

PR #145, opened 2020-11-24
docs: initial port of markdown extensions guide from VuePress

matias-capeletto:
As discussed with @kiaking, this is an initial port of the extensions markdown docs from VuePress: https://vuepress.vuejs.org/guide/markdown.html

Most of the features were already working. This is a list of changes from the original:

- `<router-link>` to router link as we are no longer using vue-router
- removed badges from headers (like `<Badge text="default theme"/>`)
- removed a link to [vuepress-plugin-container](https://vuepress.github.io/plugins/container/). It is a dead link, the project has moved to another repo.
- option `markdown.extendMarkdown` is called `markdown.config` in VitePress. I think this was a good rename. 
- I added a new folder `/docs/images` to hold the images. They were in public in VuePress.

I commented out with `TODO:`s the things that are not currently working or require discussion. I think the best way forward is to merge this PR and create an issue listing all of these as a checkbox list. I do not know if I can fix all of the things here in this PR and it is better to enable others to work on them in parallel. Some of these things may not be supported in vitepress and we can decide to drop them

TODO:
- markdown.pageSuffix option is not supported
- https://github.com/vuejs/vuepress/tree/master/packages/%40vuepress/markdown/__tests__/fragments
- two links to option descriptions are missing
- port frontmatter guide from vuepress (commented link to it)
- details block not working
- port snippets. In VuePress there were in the test folder in `../@vitepress/markdown/__tests__/fragments/`. I didn't port them because I do not know what folder structure we should follow in VitePress for them


kiaking:
Thanks a lot! I like the idea of first merging this and then tackle the issues one by one. I'll fix few things I found (like broken link to config reference) and then merge this one 👍

PR #151, opened 2020-11-24
docs: port formatter guide from vuepress

matias-capeletto:
Port of https://vuepress.vuejs.org/guide/frontmatter.html adapted to the currently supported features in vitepress

Changed:
  - `$frontmatter` -> `$page.frontmatter`
  - Predefined Variables: `meta` -> `head`

Removed:
  - toml support (I did not find support for it in the code)
  - Predefined Variables: `lang`, `description`, `layout`, `permalink`, `canonicalUrl`
  - Theme: `prev`, `next`, `search`, `tags`

Added:
  - Theme: `editLink`

We can add back prev and next once they are configurable from the frontmatter (I did not find it at least)

Detail not directly affecting this PR: maybe it is not a bad idea to add `meta` as a shortcut as it is easier to write in vuepress
```
meta:
  - name: description
    content: hello
  - name: keywords
    content: super duper SEO
```
against the vitepress version using head
```
head:
  - - meta
    - name: description
      content: hello
  - - meta
    - name: keywords
      content: super duper SEO
```

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -38,6 +38,7 @@ function getGuideSidebar() {
         { text: 'Configuration', link: '/guide/configuration' },
         { text: 'Markdown Extensions', link: '/guide/markdown' },
         { text: 'Customization', link: '/guide/customization' },
+        { text: 'Frontmatter', link: '/guide/frontmatter' },
         { text: 'Deploying', link: '/guide/deploy' }
       ]
     }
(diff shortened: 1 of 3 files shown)


kiaking:
Thanks a lot! Again I'll review this and merge. Many of your point make sense 👍

matias-capeletto:
Great, I am checking `$frontmatter` in Vuepress, and I think it may be a good idea to also expose it in Vitepress (instead of using $page.frontmatter). https://vuepress.vuejs.org/guide/global-computed.html#frontmatter, as it also read better when using it in markdown
```
{{ $frontmatter.title }}
```
I do not know if the idea is to try to keep these globals as small in number as possible. 
I could send a PR adding it if you would like 👍

kiaking:
@matias-capeletto I think Evan mentioned that in the past in README or somewhere that it was intentional. But, at the moment, I imagine having such shortcut wouldn't hurt 👍  So, I would say go ahead if you can make that happen 🙌

PR #153, opened 2020-11-25
feat: add Algolia DocSearch (#40)

posva:
Still waiting for keys to add it to vitepress docs

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -17,6 +17,12 @@ module.exports = {
       placement: 'vuejsorg'
     },
 
+    algolia: {
+      apiKey: 'c57105e511faa5558547599f120ceeba',
+      indexName: 'vitepress'
+      // algoliaOptions: { facetFilters: ['tags:guide,api'] }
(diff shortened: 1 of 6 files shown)


posva:
I'm struggling to migrate this component to script setup because I get this runtime error

Screen Shot 2020-11-26 at 11 11 32

But that's a TS type. I pushed a version with a dynamic import, which I don't think is normal. I tested using a file that exports types and importing them on a component breaks, so I think that's a bug

kiaking:
@posva You might wanna try `import type { ... } from '...'`? 

From https://github.com/vuejs/vitepress/pull/137#issue-524135713

> One quirk discovered is that when using <script setup lang="ts"> it is necessary to explicitly mark type imports with import type { X } from '...' otherwise it will be used as a value in the setup return statement. It should be technically possible to avoid this limitation with more sophisticated analysis but for now I don't think it's a deal breaker.

posva:
This should be good now!

kiaking:
Niiice!

PR #156, opened 2020-11-26
docs: port of assets page from vuepress

matias-capeletto:
Port from https://vuepress.vuejs.org/guide/assets.html bringing the docs for vite from https://github.com/vitejs/vite#asset-url-handling

Removed $withBase global:

> But this is brittle if you ever decide to change the `base`. To help with that, VitePress provides a built-in helper `$withBase` (injected onto Vue’s prototype) that generates the correct path:
> 
> ```vue
> foo
> ```
> 
> Note you can use the above syntax not only in theme components, but in your Markdown files as well.
> 
> Also, if a `base` is set, it’s automatically prepended to all asset URLs in `.vitepress/config.js` options.

I think we should provide this global in vitepress.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -36,6 +36,7 @@ function getGuideSidebar() {
         { text: 'What is VitePress?', link: '/' },
         { text: 'Getting Started', link: '/guide/getting-started' },
         { text: 'Configuration', link: '/guide/configuration' },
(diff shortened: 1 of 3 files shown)


kiaking:
Niiice! You are so true about `$withBase` global. In fact, I'm gonna add it right away so we can add that doc section back in and merge this PR 👍 Just a sec (minutes, hours, maybe

PR #161, opened 2020-11-26
Update basics.md

akauppi:
Lowercasing `description` to be akin to other entries. Shows especially in the navigation pane.


kiaking:
Oops! Thanks a lot! Good catch.

Issue #124, opened 2020-11-14
`_assets` folder not accessible when deploying to GitHub Page


xiaoxian521:
**Describe the bug**
Package and deploy to GitHub, no resource package 404 error found

**To Reproduce**
Steps to reproduce the behavior: Package and deploy to GitHub, no resource package 404 error found

**Expected behavior**
Package and deploy to GitHub, no resource package 404 error found

**System Info**
- vitepress version: 0.7.4
- vite version: 1.20.0
- Node version: 12.13.1
- OS version: windows10

**Additional context**

xiaoxian521:
After packaging, open http-server locally to allow the project to be fully OK, but the deployment of GitHub will report 404, and no resources can be found

as https://xiaoxian521.github.io/zh/standard/

xiaoxian521:
@yyx990803 You da, this problem has troubled me for a long time. Can you solve it and support you forever

kiaking:
@xiaoxian521 I think you need to add `base` option...?

```js
// .vitepress/config.js

module.exports = {
  base: '/standard/',
  ...
}
```

yuxino:
@kiaking @xiaoxian521 i think i can answer this question . 

![image](https://user-images.githubusercontent.com/12481935/99228035-0b19ae00-2827-11eb-8b52-4cd546160bcb.png)

This seems to be a dns issues . If you are in China, you can try to use a global proxy.

xiaoxian521:
The access path is as follows: https://xiaoxian521.github.io/

I am here config.js base: "/" is added to the, but it still doesn't work

xiaoxian521:
@yuxino But there is no problem with vuepress，It should be the same

yuxino:
@xiaoxian521 i can't access any `github.io` site today , i think it's same question not vitepress.u can't access that http://elemefe.github.io/mint-ui, right ?

xiaoxian521:
@yuxino I can open it http://elemefe.github.io/mint-ui

kiaking:
Oh wait, yea yea yea I recall this issue. We can't use underscore as a folder name in GitHub pages. So `_assets` folders doesn't work. We need to add `.nojekyll` file to the root.

@xiaoxian521 As a workaround, could you test placing `public/.nojekyll` file, and then deploy the build?

xiaoxian521:
Perfect solution. Thank you very much

kiaking:
[...]

PR #162, opened 2020-11-26
docs: port deploy instructions for common services from vuepress

matias-capeletto:
Port deploying to common services mini-tutorials from vuepress: https://vuepress.vuejs.org/guide/deploy.html

Deploying works the same way, so not much to change except for vuepress -> vitepress

The only modified section is Vercel, because for vuepress it is pointing to a post in vercel.com. I extracted the deploying guide from that post and included a picture of the wizard base configuration process.

Note: I did not test these instructions in all the platforms, but I assume the docs in vuepress are up to date in this regard.


kiaking (changes requested):
Nice work! For the GitHub deploy, could you add instruction to place `.nojekyll` file in `public` folder. Regarding https://github.com/vuejs/vitepress/issues/124#issuecomment-727821279.

Then I guess we can close #124 by this PR 😳

matias-capeletto:
Ok, I'll add this later 👍 

Are there benefits to use `_assets` instead of `assets` like VuePress does and you proposed in that issue? Because If not, maybe we should change that so users do not trip with this (a lot of people will not read that part of the docs for sure)

kiaking:
I think `_assets` thing is coming from Vite (need to double check). And there's technically nothing wrong with having underscore in URL. It's more of GitHub page issue really.

But, yes, this will trip lots of users, so we might wanna add `.nojekyll` to the `dist` folder on build time. Hmmm... ok let's not close #124 with this issue, and let's provide some feasible solution.

matias-capeletto:
Ok, I added a temporal warning in the mean time as you proposed.

kiaking (approved):

PR #163, opened 2020-11-27
fix: align $title with vuepress (#158)

### LINKED ISSUE #158: $title compat with vuepress
@kiaking when working on https://github.com/vuejs/vitepress/pull/152, I assumed that it was the title that appeared in the sidebar links:
```
$title = $page.title || $site.title
```

But reviewing [the code in Vuepress](https://github.com/vuejs/vuepress/blob/fbf5e5d99b40e3842a0a2566f9eb35e05863eab5/packages/%40vuepress/core/lib/node/ClientComputedMixin.js#L78), it looks like it is actually the title of the document:
```
$title = ( $page.title ? $page.title + ' | ' + $site.title ) : $site.title
```

I think it is more useful as we have it (and maybe add a $documentTitle), but maybe we need to align it to vuepress and have:
```
$title <- same as in vuepress
$sidebarTitle <- current one (or we do not expose this one at all)
```

It will be an interesting walk to keep as much of compat with vuepress but take the opportunity to break what could be improved. I am fine with both options here.

Thoughts?

matias-capeletto:
Fixes #158 

@kiaking I tried to do the localized `pageData.title` change but looks like it is more involved than I thought. I do not have the siteDataByRoute when pageData is created https://github.com/vuejs/vitepress/blob/master/src/node/markdownToVue.ts#L42
I still think that it would be good that do it, but I'll leave this one to someone else to check.
The PR only fixes the bug and aligns $site with the docs and vuepress.

### DIFF
--- a/src/client/app/mixin.ts
+++ b/src/client/app/mixin.ts
@@ -44,7 +44,9 @@ export function mixinGlobalComputed(
 
     $title: {
       get() {
-        return page.value.title || siteByRoute.value.title
+        return page.value.title
+          ? page.value.title + ' | ' + siteByRoute.value.title
+          : siteByRoute.value.title
       }
     },
 


kiaking (approved):
Oh OK cool. Yeah, by the way I also thought, if we were to always add `page.title` at build time, we should do the same for `description` (should always have site.description if page.description is undefined).

I agree this can be a bit tricky, and also, at the moment, both `page.title` and `page.description` is in sync, so I guess that't good.

Let's come back to making `page.title` to have the default value outside of this PR 👍

Issue #157, opened 2020-11-26
Recommended way to use Custom Vue components in .md


patak-dev:
I am looking into porting the using vue in Markdown from vuepress: 
https://vuepress.vuejs.org/guide/using-vue.html

If I understand correctly, vitepress doesn't want to auto register components by convention as Vuepress does. Is this the case? I actually liked this feature, but I understand that vitepress wants to keep the moving parts as small as possible.

What is the recommended way to register the components? I see that in vue-router-next docs they are registered globally inside `enhanceApp`: https://github.com/vuejs/vue-router-next/search?q=HomeSponsors.
Same as with this comment: https://github.com/vuejs/vitepress/issues/92#issuecomment-724645482

Should we document this way in the docs?

Some thoughts about this. It would be great that users that want to use the default theme as is, do not need to learn straight away about enhanceApp to be able to use a vue component in their markdown.

If auto registering by convention in a folder like `.vitepress/components` is not an option, could we import them directly in the markdown?

```markdown
# Docs

This is a .md using a custom component

<CustomComponent />

## More docs

...

<script setup>
  import CustomComponent from '../components/CustomComponent.vue'
</script>
```

Script & style hoisting is working in vitepress: https://vuepress.vuejs.org/guide/using-vue.html#script-style-hoisting, but I tried this example to import a Component and it is not at this point.

kiaking:
> If I understand correctly, vitepress doesn't want to auto register components by convention as Vuepress does. Is this the case?

Yes, at the moment it's. This is a kind of feature that we would want to tell users to use VuePress instead.

A little background on this. It might subject to change in the future, but for now, VitePress is aiming to be smaller VuePress. Like Lumen for Laravel (if you're familiar with PHP 👀). VitePress should have bare minimal features for authoring docs, and for simple markdown based build system for Vue, and all necessary features like Search, Service Workers, built in to the core, where else VuePress has those features enabled by plugins. That's reason why we're merging features like Carbon Ads.

And also to note, VuePress might integrate with Vite in the future. So that user can get benefits from Vite, with all the powerful features in VuePress (we should consider the naming though 😅).

However, maybe we could provide a script that traverse file system and auto include components inside `.vitepress/theme/index.js` (not sure if possible yet), then add that as a recipe in the docs.

> What is the recommended way to register the components? I see that in vue-router-next docs they are registered globally inside `enhanceApp`

[...]

PR #164, opened 2020-11-28
docs: add migrating from vuepress guide

matias-capeletto:
This is a review of current differences in features and config options between VuePress 1.7.1 and VitePress. I was creating these notes as part of porting the docs from VuePress and I think it is helpful that they are part of the formal documentation.

I added the warning from the readme about the WIP status so we do not give the wrong idea by publishing this guide.

At this point, apart from being useful for people that want to play with VitePress, I think it will be good to track the differences and see if the renamings have been done for a reason or can be aligned and if there are features that can be implemented to ease the migration path.

Before 1.0 is released, It would be interesting to _Upgrade Path_ explanations for each of the changes like it is done in https://vuepress.vuejs.org/miscellaneous/migration-guide.html. For the moment, as some of these differences may disappear, I do not think that it is worth adding.

This document should be fairly complete. This a list of things I did not know if Vitepress supports:
- [Multiple sidebars](https://vuepress.vuejs.org/theme/default-theme-config.html#multiple-sidebars)(https://vuepress.vuejs.org/theme/default-theme-config.html#multiple-sidebars)
- [content excerpt](https://vuepress.vuejs.org/theme/writing-a-theme.html#content-excerpt)
- [prepocessors](https://vuepress.vuejs.org/guide/using-vue.html#using-pre-processors)

It also needs to be reviewed if there are any differences in the [i18n support](https://vuepress.vuejs.org/guide/i18n.html)

### DIFF
--- a/docs/guide/migrating-from-vuepress.md
+++ b/docs/guide/migrating-from-vuepress.md
@@ -1,6 +1,6 @@
 # Migrating from VuePress
 
-VitePress follows VuePress config and default theme API where possible to ease the migration path. There are although some features that are not present in vitepress because of differences in the [design goals](../index.md) of the two projects. VitePress aims to have bare minimal features for authoring docs and most features are pushed to the Themes, where else VuePress has those features enabled by plugins.
+VitePress follows VuePress config and default theme API where possible to ease the migration path. However, there are some features that are not present in VitePress because of differences in the [design goals](../index.md) of the two projects. VitePress aims to have bare minimal features for authoring docs and most features are pushed to the Themes, whereas VuePress has those features enabled by plugins.
 
 This is a list of changes and removed features compared to [VuePress v1.7.1](https://github.com/vuejs/vuepress/releases/tag/v1.7.1)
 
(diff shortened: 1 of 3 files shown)


docs/guide/migrating-from-vuepress.md
```diff
@@ -0,0 +1,105 @@
+# Migrating from VuePress
+
+VitePress follows VuePress config and default theme API where possible to ease the migration path. There are although some features that are not present in vitepress because of differences in the [design goals](../index.md) of the two projects. VitePress aims to have bare minimal features for authoring docs and most features are pushed to the Themes, where else VuePress has those features enabled by plugins.
```
fotijr, line 3: > There are although some features that are not present in vitepress because of differences...

Seems to be a typo here, I would go with something like:
> However, there are some features that are not present in VitePress because of...

fotijr: > where else VuePress

guessing you meant:

> whereas VuePress

docs/guide/migrating-from-vuepress.md
```diff
@@ -0,0 +1,105 @@
...
+
+::: warning
+
+Note this is early WIP! Currently the focus is on making Vite stable and feature complete first. It is not recommended to use this for anything serious yet.
+
+Some of these differences may be gone before Vitepress 1.0 is released.
+
+:::
+
+## General
```
fotijr, line 15: I would start each bullet point with a capital letter.

docs/guide/migrating-from-vuepress.md
```diff
@@ -0,0 +1,105 @@
...
+
+- removed support for [toml in frontmatter](https://vuepress.vuejs.org/guide/frontmatter.html#alternative-frontmatter-formats)
+- removed [details block](https://vuepress.vuejs.org/guide/markdown.html#custom-containers)
+- removed [markdown slots](https://vuepress.vuejs.org/guide/markdown-slot.html)
+  guide/using-vue.html#using-components).
+- removed `~` prefix to explicitly specify a url is a [webpack module request](https://vuepress.vuejs.org/guide/assets.html#relative-urls)
+
+## Site Config
+
+- renamed `temp` to `tempDir`
```
fotijr, line 36: It might improve readability if you grouped the changes together, like:

- Renamed:
  - `temp` to `tempDir`
  - `dest` to `outDir`
...
- Removed:
  - `theme` from a dependency
  - `permalink`
  - `port`
...

fotijr (commented):
I'm new to this project so take my comments with a grain of salt. I found your guide very helpful, thanks for putting this together!

docs/guide/migrating-from-vuepress.md
matias-capeletto: Done, thanks a lot!

docs/guide/migrating-from-vuepress.md
matias-capeletto: Done 👍🏼

docs/guide/migrating-from-vuepress.md
matias-capeletto: I agree, changed the PR to this structure. Thanks for the suggestions 💯

matias-capeletto:
@fotijr Thanks for the feedback 🙌🏼 

Once VitePress is stable, we can improve this go further including a migration path for removed features when possible. 
I expect that a lot of the differences that are currently listed will end up disappearing though.

Some examples:

- There is a PR to add [description to frontmatter](https://github.com/vuejs/vitepress/pull/170)
- There is a PR to include all the [missing computed globals](https://github.com/vuejs/vitepress/pull/167) from VuePress, so that section will go away if merged.
- Evan said in a comment that [adding auto registering of components is still on the table](https://github.com/vuejs/vitepress/issues/157#issuecomment-739117124) 

And there are other low hanging fruits that may ease the transition.

I will keep this guide up to date as new features are merged.

kiaking:
Thanks and thanks for the PR! This is looking great, and very helpful. Though I have 2 concerns.

1. Maybe this is really hard to maintain. It might suit better to have this as a blog post or something that outdated content wouldn't have much impact. Maybe we can link it from README or at TOP page of the docs.
2. Maybe we should name it "Differences from VuePress". Since as a design concept, VuePress should have more rich features, so migrating to VuePress should be the path users would take 👀  Ideally, when user moves to VuePress, they wouldn't comeback to VitePress. However this is future idea, so at the moment it makes sense to list the differences between VuePress and VitePress, so it's only about page title.

matias-capeletto:
1. I wouldn't want to share this as a blog post, because if people find it later it may end up confusing more than it helps. But I agree with you that this may not be the time to merge this because of the extra bookkeeping when things are changing a lot. I am fine closing the PR, mainly shared because I had the list at hand 👍🏼 
2. If we decide to close this now, I still think it is useful to have this page up at some point, maybe when VitePress hits v1. I like your idea to change the title to avoid giving the wrong impression. I do not fully understand the migrating from VitePress to VuePress though, I think that VitePress may end up being as powerful as VuePress (but with features implemented in themes instead of plugins) so I do not see myself at least wanting to migrate to VuePress. But maybe I am missing something, the title change would still be good 🙂 

So, if you want, I'll close this, and we can give it another go before v1.

kiaking:
Very good points. Hmmmm this is tough one. But, OK I would say let's add this one.

This content is really useful especially now, since VitePress lacks lots of its docs, and as you mentioned, we should have this before v1 anyway 😅 

>  the title change would still be good 🙂

Yea I would still like to change the title. I think you have very good points, though another reason might be "migrating" sounds like it's a "good" thing. I mean it's not a bad thing, but I think we shouldn't give people too much impression that moving to VitePress from VuePress is GOOD thing to do... you know what I mean?

Like "migrating from Vue 2 to Vue 3", that's something we want people to do.  "Migrating to VitePress" feels to me like you *SHOULD* do it, but I think what we want is actually more like "if you want, this is how what you need to check, but if you decide not to, that's totally fine! Please continue using VuePress 🙌 ".

Maybe I'm having too much preference on the word but, wha do you think? 😆

matias-capeletto:
I think that you are on point about worrying about how we are presenting this guide. I pushed changes in the wording and added a note explicitly stating that there is no need to migrate to VitePress. I think it reads better now, let me know if you would like other changes 👍🏼

kiaking (approved):
Great work! Thank you so much for the write up! 🚀 🚀 🚀

PR #165, opened 2020-11-30
fix: prevLinks and nextLinks config type

matias-capeletto:
Corrected a missing s in the theme config for prevLinks and nextLinks


kiaking:
Thank you so much 👍

PR #167, opened 2020-11-30
feat: $lang and $localePath globals (#166)

### LINKED ISSUE #166: $lang and $localePath computed globals
`$lang`: The language of the current page. Default: en-US.
`$localePath`: The locale path prefix for the current page. Default: /.

These are the only two computed globals from Vuepress that are not currently in Vitepress. 

I think they are good additions. I could send a PR for them if we want to add them.

matias-capeletto:
close #166 

As discussed in #166 

I now see that there several globals that [are exposed](https://github.com/vuejs/vuepress/blob/fbf5e5d99b40e3842a0a2566f9eb35e05863eab5/packages/%40vuepress/core/lib/node/ClientComputedMixin.js#L50) but are not documented.

`$localeConfig` and `$themeLocaleConfig` looks like they are parts of `$siteByRoute` in vitepress
There are also `$siteTitle`, `$canonicalUrl`

In my opinion, there is no need to add any of these four globals that are not documented.

### DIFF
--- a/docs/guide/global-computed.md
+++ b/docs/guide/global-computed.md
@@ -64,6 +64,14 @@ Reference of `$page.frontmatter`.
 }
 ```
 
+## $lang
+
+The language of the current page. Default: `en-US`.
+
+## $localePath
+
+The locale path prefix for the current page. Default: `/`.
+
 ## $title
 
 Value of the `<title>` label used for the current page.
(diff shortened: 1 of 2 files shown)


kiaking:
Thanks a lot and sorry for the late reply! Looking good 👍 👍 👍  And yes, I agree we don't need other globals which are not documented 😅

matias-capeletto:
Hey @kiaking, don't worry 😄
Thanks for merging it 👍🏼

PR #176, opened 2020-12-04
fix: avoid layout shift due to ads

posva:
This is specific to mobile viewports. Since the ad container goes on top of the content, loading it only on client side creates a layout shift:

Screen-Recording-2020-12-04-at-11 25 56


posva:
I think this is the simplest and most maintainable solution: remove the async loading for the CarbonAds component + min-height on the ads container. It does add the CarbonAds to the whole bundle but I would say it's reasonable.

We can also expose the Component to be used by custom themes or build a more complex solution like a placeholder that takes up the space on mobile, which would create duplication.

posva:
Should be good with a container in layout and keeping the component async!

src/client/theme-default/Layout.vue
```diff
@@ -150,3 +154,34 @@ const pageClasses = computed(() => {
...
+    float: right;
+    margin: -8px -8px 24px 24px;
+    width: 146px;
+  }
+}
+
+@media (max-width: 420px) {
+  #ads-container {
+    /* Avoid layout shift */
+    height: 105px;
```
kiaking, line 176: This is tough one, but should it be `min-height` instead? Because depending on the ads, it might exceed 105px...?

posva: but that would also create a layout shift, no? 🤔

kiaking: Well yeah... if the ad has long text, in that case it would cause layout shift for few pixels... not sure how often that occurs... but yeah, I think we do have enough space under the ad so it should be fine!

PR #190, opened 2020-12-23
fix: style pollution on custom theme

antfu:
When using a custom theme, the style from the default theme will be injected and polluting the custom theme (overrides) due to the styles import side effect from re-exporting the defaultTheme here:

https://github.com/vuejs/vitepress/blob/18d18d2eb15a569113ca68ccbb9ba52dfd46c80a/src/client/app/exports.ts#L29

Repro: https://codesandbox.io/s/vitepress-style-pollution-p5kvq?file=/index.md

This PR removed the redirecting as it does not seem to be used and users should always import the theme from `vitepress/dist/client/theme-default` explicitly.

### DIFF
--- a/src/client/app/exports.ts
+++ b/src/client/app/exports.ts
@@ -24,6 +24,3 @@ import { ComponentOptions } from 'vue'
 import _Debug from './components/Debug.vue'
 const Debug = _Debug as ComponentOptions
 export { Debug }
-
-// default theme
-export { default as defaultTheme } from '/@default-theme/index'


antfu:
@knightly-bot build this

matias-capeletto:
I think this export was added after this discussion: https://github.com/vuejs/vitepress/issues/157#issuecomment-739117124
The idea was that users that only need to extend the theme to register a few global components, wouldn't need to also take care of importing the styles:
```ts
// .vitepress/theme/index.js
import { defaultTheme } from 'vitepress'
import MyGlobalComp from './Comp.vue'

defaultTheme.enhanceApp = ({ app }) => {
  app.component('MyGlobalComp', MyGlobalComp)
}
```

But I see your point that it is impossible to use the theme without any styles. Maybe there is a way to opt out with an option in the `theme.enhanceApp()` function?

antfu:
@matias-capeletto The current implementation make importing `vitepress` ** non-side-effect-free**.

```js
import 'vitepress' // will import the css side effects as well
```

I don't think they can be solved without some hacks. I am thinking user should import the default theme explicitly like:

```js
import defaultTheme from 'vitepress/dist/client/default-theme`
```

Which is also recommended by the [Customization Guide](https://vitepress.vuejs.org/guide/customization.html)

----

Maybe we can add exports alias in package.json to achieve something like

```js
import defaultTheme from 'vitepress/default-theme`
```

And I think the `enhanceApp` hook is a good idea, and we can use it regardless the default or custom theme

```js
import { enhanceApp } from 'vitepress'

enhanceApp(({ app }) =>  {

})
```

matias-capeletto:
> Which is also recommended by the Customization Guide

Yes, the change was not documented and I think it was still WIP. Totally agree that importing vitepress should not include the CSS. 
I like your proposal 👍🏼 
```
import defaultTheme from 'vitepress/default-theme`
```

PR #200, opened 2021-01-05
fix: scrollbar when using line highlight

matias-capeletto:
Hides spurious scrollbar in code blocks that appears when using line highlighting
https://vitepress.vuejs.org/guide/markdown.html#line-highlighting-in-code-blocks

image

### DIFF
--- a/src/client/theme-default/styles/code.css
+++ b/src/client/theme-default/styles/code.css
@@ -85,6 +85,7 @@ li > div[class*='language-'] {
   font-family: var(--code-font-family);
   font-size: var(--code-font-size);
   user-select: none;
+  overflow: hidden;
 }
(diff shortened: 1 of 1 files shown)


matias-capeletto:
I just opened https://github.com/vuejs/vitepress/issues/201, the direct link I included in the description was not actually working. Maybe an SSR issue.

@antfu should I send a PR for the styling of the scrollbars too? Not every app does it, GitHub is leaving the default ones for windows. VuePress also leaves the default ones, maybe this is something that users could decide actually.

antfu:
RE: Scrollbar. I think it's better to keep the default style, as documentation tools we shouldn't bring too many opinions on native elements.

PR #203, opened 2021-01-08
fix: resolve page hash case-insenstively, close #202

### LINKED ISSUE #202: Case-insenstive support
**Describe the bug**

When deploying to Netlify, it will redirect the URLs to the lowercase version (which is [forced and no way to opt-out](https://community.netlify.com/t/my-url-paths-are-forced-into-lowercase/1659)). When we serving docs that contain uppercases, the SSG content will be displayed correctly, and after JS loading finished, the whole page will be completely white out.

**To Reproduce**

Example demo: https://5ff76cdd71e66e000a4ef7f0--vueuse.netlify.app/core/useMouse/

image

**Expected behavior**

Resolve the lean.js hash case-insensitively.

**System Info**
- vitepress version: 0.10.7
- vite version: 2.0.0-beta.10
- Node version: v14.15.3
- OS version: macOS v11.1

**Additional context**

With #203: https://vitepress--vueuse.netlify.app/core/useMouse/

antfu:


antfu:
@knightly-bot build this

yyx990803 (changes requested):
I think it's better to directly enforce lowercase entry names [here](https://github.com/vuejs/vitepress/blob/master/src/node/build/bundle.ts#L31).

antfu:
See #202. In Netlify, every url is case-insensitive, which means `/assets/a.js` will be exactly the same as `/assets/A.js`. The problem here is that the hashMap resolving is case-sensitive. For example

```js
const pageToHashMap = {
  hello: '123456789'
}

const hash = pageToHashMap['Hello'] // undefined
const url = `/a.${hash}.js` // /a.undefined.js
```

image


Forcing to generate lowercase output as you suggested might be a good addition but I guess it's not really related to fixing the issue. (or did I misunderstand anything?)

Thanks.

yyx990803 (approved):

PR #204, opened 2021-01-13
snippet plugin root params Missing default value

hcg1023:


yyx990803:
Why does it need this? Making it optional makes it possible for other parts of the codebase to forget it and lead to potential errors.

hcg1023:
> Why does it need this? Making it optional makes it possible for other parts of the codebase to forget it and lead to potential errors.

Now introducing a block of code in Markdown using the @ sign will result in undefined, and when Markdown-it uses SnippetPlugin, no parameters are defined, so provide a default value to keep its behavior consistent with VuePress

yyx990803:
What kind of markdown causes this problem? I'd help to at least have an example. When submitting PRs, it's better to provide enough context upfront since we don't necessarily know what problem you are actually dealing with.

hcg1023:
> 哪种减价会导致此问题？我至少要举一个例子。提交PR时，最好预先提供足够的上下文，因为我们不一定知道您实际上正在处理什么问题。

I'm sorry, this is a case where I had a problem：
https://github.com/hcg1023/vitepress-bug

kiaking:
Oh OK now I see the problem. So `snippetPlugin` is never getting the 2nd argument `root` because it's not passed from anywhere. So instead of setting the default value to the argument, I think we should pass down the `root` value from the `createMarkdownToVueRenderFn` in `markdownToVue.ts`.

```ts
export function createMarkdownToVueRenderFn(
  root: string,
  options: MarkdownOptions = {}
) {
  const md = createMarkdownRenderer(options, root) <- pass root

  ...
```

```ts
export const createMarkdownRenderer = (
  root: string, // <- add
  options: MarkdownOptions = {}
): MarkdownRenderer => {
  const md = MarkdownIt({
    html: true,
    linkify: true,
    highlight,
    ...options
  })

  // custom plugins
  md.use(componentPlugin)
    ...
    .use(snippetPlugin, root) // <- pass in root
    ...
```

@hcg1023 Would you like to update the PR?

hcg1023:
> createMarkdownToVueRenderFn

yes

> Oh OK now I see the problem. So `snippetPlugin` is never getting the 2nd argument `root` because it's not passed from anywhere. So instead of setting the default value to the argument, I think we should pass down the `root` value from the `createMarkdownToVueRenderFn` in `markdownToVue.ts`.
> 
> ```ts
> export function createMarkdownToVueRenderFn(
>   root: string,
>   options: MarkdownOptions = {}
> ) {
>   const md = createMarkdownRenderer(options, root) // <- pass root
> 
>   ...
> ```
> 
> ```ts
> export const createMarkdownRenderer = (
>   root: string, // <- add
>   options: MarkdownOptions = {}
> ): MarkdownRenderer => {
>   const md = MarkdownIt({
>     html: true,
>     linkify: true,
>     highlight,
>     ...options
>   })
> 
>   // custom plugins
>   md.use(componentPlugin)
>     ...
>     .use(snippetPlugin, root) // <- pass in root
>     ...
> ```
> 
> @hcg1023 Would you like to update the PR?

Yes, I have updated my Commit

kiaking (approved):
Thanks for the update and finding the issue! 🙌

PR #207, opened 2021-01-19
fix: render document with standards mode

Wizard67:
This PR fixes the document rendering mode when running `docs-dev` script.

When `<!DOCTYPE html>` is not specified, the browser will use Quirks Mode to render the document, which will cause some third-party libraries to generate wrong logic, such as popperjs.

> In Firefox, select *View Page Info* from the context menu, and look for *Render Mode*.


kiaking (approved):
Thank you so much! 🙌

PR #210, opened 2021-01-21
Make serve url clickable

domnantas:
`vitepress serve` adds a `.` to the logged url, which makes the url non clickable on some terminal emulators.

Replaced `.` with `/` to match Vite server: https://github.com/vitejs/vite/blob/57c03435256b97857700e3df0e45e89fc324ed2c/packages/vite/src/node/server/index.ts#L522


kiaking:
Ah nice catch! Thanks 👍

Issue #195, opened 2021-01-03
locale issues: need different title href URL and home page hero link


ShenQingchuan:
## Problem 1

**Describe the bug**
I need '/' as the title href for English version, and '/zh/' for Chinese version.
But I didn't found any useful suggestions in the documentation.

**To Reproduce**
switch to the '/zh/xxx' page, but the title href is still targeting to '/':
![image](https://user-images.githubusercontent.com/46062972/103481897-94f1fa80-4e18-11eb-8e7b-9f36750eab99.png)

**Expected behavior**
'/zh/xxx' pages' titie href is suppposed to be '/zh/' ( which defines as the key of `themeConfigs.locales`)

**System Info**
- vitepress version: v0.10.5
- vite version: v2.0.0-beta2
- Node version: v12.16.2
- OS version: 64bit Mac OS X 10.15.6 19G2021

# Problem 2

When I toggle the language, the `actionText` in `docs/index.md`  and `docs/zh/index.md` doesn't work well as expected:

Reproduce in my fork for [vitejs-docs Chinese translation repo](https://github.com/ShenQingchuan/vite)

![屏幕录制2021-01-04 13](https://user-images.githubusercontent.com/46062972/103503489-62cfaf80-4e8f-11eb-8186-b22f6f80f85c.gif)

ShenQingchuan:
# Another problem

When I toggle the language, the `actionText` in `docs/index.md`  and `docs/zh/index.md` doesn't work well as expected:
![屏幕录制2021-01-04 13](https://user-images.githubusercontent.com/46062972/103503489-62cfaf80-4e8f-11eb-8186-b22f6f80f85c.gif)

Reproduce in my fork for [vitejs-docs Chinese translation repo](https://github.com/ShenQingchuan/vite)

ShenQingchuan:
seems like the reactivity was lost when destructed....

ShenQingchuan:
# 对问题 2 的解决方案 👆

请 @yyx990803  审阅

## 导致错误的原因：
1. 从 `defineProps()` 结构出来的变量 丢失了响应性
2. `useNavLink` 方法中 `props` 计算属性的闭包函数缓存了该 `item` 值
3. 再从 https://github.com/vuejs/vitepress/blob/a9f75749a683d84bfd2245c989bdd7f67cfb1c55/src/client/theme-default/components/NavLink.vue#L19 解构出去时，props 内的所有属性都将永远不变

## 解决方案：
[...]

PR #212, opened 2021-01-23
fix: HomeHero NavLink reactivity lost (#195)

ShenQingchuan:
Please see https://github.com/vuejs/vitepress/issues/195 for more details... (May Chinese required)

### DIFF
--- a/src/client/theme-default/components/NavLink.vue
+++ b/src/client/theme-default/components/NavLink.vue
@@ -7,16 +7,16 @@
 </template>
 
 <script setup lang="ts">
-import { defineProps } from 'vue'
+import { defineProps, toRefs } from 'vue'
 import type { DefaultTheme } from '../config'
 import { useNavLink } from '../composables/navLink'
 import OutboundLink from './icons/OutboundLink.vue'
 
-const { item } = defineProps<{
-  item: DefaultTheme.NavItemWithLink
-}>()
-
-const { props: linkProps, isExternal } = useNavLink(item)
+const props = defineProps<{
+  item: DefaultTheme.NavItemWithLink,
+}>();
+const propsRefs = toRefs(props);
+const { props: linkProps, isExternal } = useNavLink(propsRefs.item);
 </script>
 
 <style scoped>
--- a/src/client/theme-default/components/NavDropdownLinkItem.vue
+++ b/src/client/theme-default/components/NavDropdownLinkItem.vue
@@ -9,16 +9,16 @@
 </template>
 
 <script setup lang="ts">
-import { defineProps } from 'vue'
+import { defineProps, toRefs } from 'vue'
 import type { DefaultTheme } from '../config'
 import { useNavLink } from '../composables/navLink'
 import OutboundLink from './icons/OutboundLink.vue'
 
-const { item } = defineProps<{
+const props = defineProps<{
   item: DefaultTheme.NavItemWithLink
 }>()
-
-const { props: linkProps, isExternal } = useNavLink(item)
+const propsRefs = toRefs(props);
+const { props: linkProps, isExternal } = useNavLink(propsRefs.item)
 </script>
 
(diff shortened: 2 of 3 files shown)


ShenQingchuan:
# Something Weird

For the code below:
```typescript
const props = defineProps<{
  item: DefaultTheme.NavItemWithLink,
}>();
const propsRefs = toRefs(props);
const { props: linkProps, isExternal } = useNavLink(propsRefs.item);
```

It seems very tedious, right?
but when I changed it to this:
```typescript
const propsRefs = toRefs(defineProps<{
  item: DefaultTheme.NavItemWithLink,
}>());
const { props: linkProps, isExternal } = useNavLink(propsRefs.item);
```

the JS Console print this:

image

May this be a new bug or something wrong with my code ?

src/client/theme-default/components/NavLink.vue
```diff
@@ -7,16 +7,16 @@
...
-const { item } = defineProps<{
-  item: DefaultTheme.NavItemWithLink
-}>()
-
-const { props: linkProps, isExternal } = useNavLink(item)
+const props = defineProps<{
+  item: DefaultTheme.NavItemWithLink,
+}>();
+const propsRefs = toRefs(props);
+const { props: linkProps, isExternal } = useNavLink(propsRefs.item);
```
kiaking, line 19: I think we only need to stop destructuring. So following should work. Could you test with one? It's working on my env.

```suggestion
const props = defineProps<{
  item: DefaultTheme.NavItemWithLink,
}>();

const { props: linkProps, isExternal } = useNavLink(props.item);
```

kiaking (changes requested):
Thank you so much for the PR! I've left a comment so please have a look. I think we only need to change `NavLink.vue` with this change.

ShenQingchuan:
Yes. Without `toRefs` it still works, that's my mistake...

kiaking (changes requested):
I don't think we need `NavDropdownLinkItem.vue ` and `navLink.ts` changes anymore. Can we delete them? 👀

ShenQingchuan:
> I don't think we need `NavDropdownLinkItem.vue ` and `navLink.ts` changes anymore. Can we delete them? 👀

Sure, I'll refactor my commits and repush later...

ShenQingchuan:
**OOPS !! @_@** @kiaking 

 `navLinks.ts` need to be changed...

**Why I need the argument `item` of `useNavLink(item)`  to be a `Ref<DefaultTheme.NavItemWithLink>`**:

Because the value you passed into in `HomeHero.vue` as `item` v-bind: 

```vue
<NavLink
  v-if="hasAction"
  :item="{ link: data.actionLink, text: data.actionText }"
  class="action"
/>
```
is **Not** a `ref` at all, when the locales changed, the data(from frontMatter info) did change, but the object itself didn't change at all, it performs as a constant object.

cut a long story short, it will show a bug when you follow these steps:
1. open a English version doc first
2. click another language locale link
3. and you can see the homehero button's text has changed, that's fine and what we expected.
4. but if you click the button, then we will still go to the English section, not the language we wanted.

ShenQingchuan:
the `item` passed it into `useNavLink` need to be a `ref`, or it will never changed... 

you can set a `watch` and write a function to `console.log(item.link)` for debuging, 

and you'll see the `item.link` string value has never changed to anthoer language (such as 'zh') `/zh/xxx` but still `/xxx`

ShenQingchuan:
I did this PR a long time ago, I just didn't remember some details, feel really embarrassed...

kiaking:
Ah good point! Yap. To note, it's not because of how we pass props, props are reactive because `item.text` is working just fine. It's about how we're passing props to the composable function. Maybe we should have made it call it from the parent component. I might refactor this later.

But for now, thank you so much for the fix!

kiaking (approved):

PR #215, opened 2021-01-26
feat: expose @vitejs/plugin-vue config

### LINKED ISSUE #214: Expose configuration for @vitejs/plugin-vue
**Is your feature request related to a problem? Please describe.**
I am looking for a way to pass some configuration to `@vitejs/plugin-vue`, specifically `template.compilerOptions.isCustomElement`). ([code](https://github.com/vuejs/vitepress/blob/c11055c5fb31ff31502400de3e5f6f05011edd1d/src/node/plugin.ts#L33)). 

**Describe the solution you'd like**
I was think that extracting these options from `.vitepress/config.js` shouldn't be too much work. Are there any other options besides `isCustomElement` that would be worth adding?

**Describe alternatives you've considered**

**Additional context**

ChrisShank:
Fixes #214

### DIFF
--- a/src/node/config.ts
+++ b/src/node/config.ts
@@ -6,6 +6,7 @@ import { resolveAliases, APP_PATH, DEFAULT_THEME_PATH } from './alias'
 import { SiteData, HeadConfig, LocaleConfig } from '../../types/shared'
 import { MarkdownOptions } from './markdown/markdown'
 import { AliasOptions } from 'vite'
+import { Options } from '@vitejs/plugin-vue'
 export { resolveSiteDataByRoute } from './shared/config'
 
 const debug = require('debug')('vitepress:config')
@@ -21,7 +22,7 @@ export interface UserConfig<ThemeConfig = any> {
   alias?: Record<string, string>
   markdown?: MarkdownOptions
   customData?: any
-  isCustomElement?: (tag: string) => boolean | void
+  vueOptions: Options
 }
 
 export interface SiteConfig<ThemeConfig = any> {
@@ -34,7 +35,7 @@ export interface SiteConfig<ThemeConfig = any> {
   alias: AliasOptions
   pages: string[]
   markdown?: MarkdownOptions
-  isCustomElement?: (tag: string) => boolean | void
+  vueOptions: Options
 }
 
 const resolve = (root: string, file: string) =>
@@ -62,7 +63,7 @@ export async function resolveConfig(
     tempDir: path.resolve(APP_PATH, 'temp'),
     markdown: userConfig.markdown,
     alias: resolveAliases(themeDir, userConfig),
-    isCustomElement: userConfig.isCustomElement
+    vueOptions: userConfig.vueOptions
   }
 
   return config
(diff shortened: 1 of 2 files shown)


antfu:
I personally don't feel it's good to be a root-level config option as it loses some context here. Maybe we can have a config entry that forwards the options for `@vitejs/plugin-vue` instead?

ChrisShank:
@antfu Are there any other options from `@vitejs/plugin-vue` that you think would be beneficial to expose?

antfu:
I am thinking to have a top-level option `vueOptions`, where

```ts
 const vuePlugin = createVuePlugin({
    include: [/\.vue$/, /\.md$/],
    ...configs.vueOptions
  })
```

WDYT?

ChrisShank:
Ya I am happy with that! I just didnt know if we wanted the expose all of the vue plugin’s options. Thanks for the feedback, Ill update the PR later today!! 😃

ChrisShank:
@kiaking @antfu Any other feedback for this?

src/node/config.ts
```diff
@@ -6,6 +6,7 @@ import { resolveAliases, APP_PATH, DEFAULT_THEME_PATH } from './alias'
 import { SiteData, HeadConfig, LocaleConfig } from '../../types/shared'
 import { MarkdownOptions } from './markdown/markdown'
 import { AliasOptions } from 'vite'
+import { Options } from '@vitejs/plugin-vue'
```
antfu, line 9: We can alias this to `VuePluginOptions` to provide the context

ChrisShank: Sure!

src/node/config.ts
```diff
@@ -21,6 +22,7 @@ export interface UserConfig<ThemeConfig = any> {
   alias?: Record<string, string>
   markdown?: MarkdownOptions
   customData?: any
+  vueOptions: Options
```
kiaking, line 25: I think this should be optional?

```ts
{
  ...
  vueOptions?: Options
}
```

src/node/plugin.ts
```diff
@@ -24,14 +24,15 @@ const isPageChunk = (
...
+  { configPath, alias, markdown, site, vueOptions }: SiteConfig,
   ssr = false,
   pageToHashMap?: Record<string, string>
 ): Plugin[] {
   const markdownToVue = createMarkdownToVueRenderFn(root, markdown)
 
   const vuePlugin = createVuePlugin({
-    include: [/\.vue$/, /\.md$/]
+    include: [/\.vue$/, /\.md$/],
+    ...vueOptions
```
antfu, line 35: should have a fallback value here

ChrisShank: Are you referring to the case where `vueOptions` is `undefined`? It's valid to spread `undefined` so I didn't think a fallback is necessary.

antfu: Oh what? I thought it's invalid to spread undefined... nvm 😅

antfu (approved):

src/node/plugin.ts
ChrisShank: Ya I think you can also spread `null` too haha! Thanks for the code review 😄

ChrisShank:
Actually just realized that we should print a console warning to restart the dev server if `vueOptions` are updated [similar to when `base` is updated](https://github.com/vuejs/vitepress/blob/master/src/node/plugin.ts#L148).

PR #224, opened 2021-02-05
fix: sidebar 'auto' not working (#178)

### LINKED ISSUE #178: sidebar: 'auto' doesn't create sidebar
**Describe the bug**
Using sidebar auto doesn't produce sidebar at all, even it is considered to be a valid option in https://github.com/vuejs/vitepress/blob/master/src/client/theme-default/config.ts.

**To Reproduce**
1. Create a new project
2. Configure vitepress to use sidebar: 'auto' in .vitepress/config.js
3. Create some pages with headings
4. Start dev server

**Expected behavior**
A sidebar should be auto generated using the headings provided in the markdown files

**System Info**
- vitepress version: v0.9.1
- vite version: v1.0.0-rc.13
- Node version: v15.3.0
- OS version: macOS 10.15.7 (19H2)

kiaking:
fix #178 

Sidebar gets hidden because `auto` was treated as empty. Added helper function to check if the sidebar is empty or not, and refactored a bit on checking logic for better readability 👀 

This adds a breaking change where undefined sidebar now resolves to `auto` where currently it's treated as `false`. Though VuePress treats undefined sidebar as `auto` so this is how it should have been 😓

### DIFF
--- a/__tests__/client/theme-default/support/sideBar.spec.ts
+++ b/__tests__/client/theme-default/support/sideBar.spec.ts
@@ -1,9 +1,19 @@
 import {
+  isSideBarEmpty,
   getSideBarConfig,
   getFlatSideBarLinks
 } from 'client/theme-default/support/sideBar'
 
 describe('client/theme-default/support/sideBar', () => {
+  it('checks if the given sidebar is empty', () => {
+    expect(isSideBarEmpty(undefined)).toBe(true)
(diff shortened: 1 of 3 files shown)


cloydlau:
Is this removed? I got a `TypeError: Cannot read properties of undefined (reading 'some')` with the following config:
```ts
sidebar: {
  'XXX': 'auto',
}
```

PR #230, opened 2021-02-10
fix: ads display

posva:
In mobile:

- better spacing around the ad at the top
- Fix the layout shift due to float

Old:

Screen Shot 2021-02-10 at 09 41 50

Fixed:

Screen Shot 2021-02-10 at 09 43 26

https://deploy-preview-230--vitepress-docs.netlify.app/


kiaking (approved):
Nice 🙌  Now I'm noticing that ads image is squashed a little horizontally but that has nothing to do with this PR, so this fix is just making things better 👍 👍 👍

PR #232, opened 2021-02-10
feat: port CodeGroup & CodeBlock to vitepress

d-pollard:
## This PR brings parity to VitePress in terms of `CodeGroup` and `CodeBlock`

**Some questions I have:**

Where should `CodeGroup` and `CodeBlock` be registered? Currently, I have them in `src/client/app/index.ts` but it feels misplaced; please advise

Also, this still needs a bit of styling so if someone wants to help or point me in the right direction, that would be super helpful

Thanks,
Derek


brc-dd:
@d-pollard The components are missing the styles.

d-pollard:
@posva / @yyx990803 what can I do to help this get merged in?

kiaking:
Thank you so much for your PR! And definitely we want this feature. But since we've started the next-theme development, let's do it overthere. Also, Vue.js website has implemented this too so I'm planning to import this from there 👍

PR #236, opened 2021-02-11
docs: using vue in markdown guide

matias-capeletto:
This is a port of the [VuePress Using Vue in Markdown Guide](https://vuepress.vuejs.org/guide/using-vue.html)

The main difference is that there is no support for auto registering components. We can add that section if the feature gets implemented in VitePress. Instead of that, there is a section that showcases how to register components globally and how to import them in .md files.

@kiaking I respected the order of VuePress sections in the guide, but IMO it would be better to start by showing how to register and import the documents, and leave the first section about ClientOnly for the end (sort of advanced). But maybe it was pushed to the top because people were having issues with this in particular and they wanted to highlighted it. What do you think? (it can also be done later, better to have the guide up for now)

Notes:
Removed the comment about the page Vue component not having private data that I didn't understood.
I did not include a section about using `pug` for templates, because I am unsure about how it works, we can add that later also.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -53,6 +53,7 @@ function getGuideSidebar() {
         { text: 'Configuration', link: '/guide/configuration' },
         { text: 'Asset Handling', link: '/guide/assets' },
         { text: 'Markdown Extensions', link: '/guide/markdown' },
+        { text: 'Using Vue in Markdown', link: '/guide/using-vue' },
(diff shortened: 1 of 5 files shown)


matias-capeletto:
I forgot to add that I tested the features before including them (more testing may be needed though)
There is also a new folder `docs/components` because I wanted that the guide would use the Component in Headers feature, so we can have some free tests.

Issue #209, opened 2021-01-21
Build fails with custom theme and no css


domnantas:
**Describe the bug**
I started building a custom theme and haven't imported any css to `index.js` yet. Running `vitepress build` throws an error:
```
$ vitepress build
vitepress v0.11.4
vite v2.0.0-beta.35
✓ building client + server bundles...
✖ rendering pages...
build error:
 TypeError: Cannot read property 'fileName' of undefined
    at Object.renderPage (/Users/fistmenaruto/Projects/domnantas.lt/node_modules/vitepress/dist/node/build/render.js:53:61)
    at async Object.build (/Users/fistmenaruto/Projects/domnantas.lt/node_modules/vitepress/dist/node/build/build.js:29:17)
error Command failed with exit code 1.
```

`render.js:53` is trying to get `cssChunk.fileName`:
```jsx
<link rel="stylesheet" href="${siteData.base}${cssChunk.fileName}">
```
but it is undefined since there are not css chunks.

The same happens if you try to import an empty .css file to `index.js`.

`vitepress dev` does not throw the error.

**To Reproduce**
- Create an empty custom theme as described in the docs: https://vitepress.vuejs.org/guide/customization.html#customization
(notice how there are no .css imports in the docs either)
- Run `vitepress build`

Reproduction repo: https://github.com/domnantas/domnantas.lt/tree/no-css-chunks

**Expected behavior**
Build should not fail when building without css

**System Info**
- vitepress version: v0.11.4
- vite version: v2.0.0-beta.35
- Node version: v15.5.1
- OS version: MacOS 11.0.1

**Additional context**
```js
`
...
${cssChunk ? `<link rel="stylesheet" href="${siteData.base}${cssChunk.fileName}">` : ""}
...
`
```
would fix this, but maybe there are better ways

kiaking:
Good catch! Thanks for the report. I think your fix is reasonable. If there're no css files, we should just stop requiring it. Would you like to make a PR? 👀

domnantas:
Sure, I can do that

PR #239, opened 2021-02-11
fix: build fails without css chunks (#209)

domnantas:
fixes https://github.com/vuejs/vitepress/issues/209

`renderPage` should probably be dissected to smaller, pure functions to make it testable.


kiaking (approved):
🚀 🚀 🚀

PR #247, opened 2021-02-24
fix(types): `@` symbol not resolving on build files

kiaking:
We have `paths` option set at `tsconfig.json`, however, this path is not resolved during the bundle time and bundled files will contain things like `import {...} from '/@types` as is. Currently, it is not causing major issue because these imports are only importing types.

But when we try to create custom theme that imports for example `useSiteData`, it throws type error because it can't find `/@types`.

This PR replace import path alias to the actual relative path to resolve the issue.

We might should remove all aliases except `vitepress`, but in this PR, I'm just removing the paths that actually causing issue at the moment.

### DIFF
--- a/src/shared/tsconfig.json
+++ b/src/shared/tsconfig.json
@@ -1,10 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "paths": {
-      "/@types/*": ["../../types/*"]
-    }
+    "baseUrl": "."
   },
   "include": [
     ".",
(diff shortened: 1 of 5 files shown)


src/shared/tsconfig.json
```diff
@@ -1,10 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "paths": {
-      "/@types/*": ["../../types/*"]
-    }
+    "baseUrl": "."
```
kemila, line 4: 111

yyx990803:
This seems outdated.

PR #256, opened 2021-03-05
fix(css): reuse css vars

posva:
I found some hardcoded colors that I adapted to existing variables. Found this while inspecting the existing docs of vite.

I also added a `--c-bg-accent` color

### DIFF
--- a/src/client/theme-default/components/HomeHero.vue
+++ b/src/client/theme-default/components/HomeHero.vue
@@ -132,20 +132,20 @@ const hasAltAction = computed(() => data.value.altActionLink && data.value.altAc
   line-height: 44px;
   font-size: 1rem;
   font-weight: 500;
-  color: #ffffff;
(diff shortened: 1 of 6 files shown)


src/client/theme-default/components/HomeHero.vue
```diff
@@ -132,20 +132,20 @@ const hasAltAction = computed(() => data.value.altActionLink && data.value.altAc
...
-  color: #ffffff;
+  color: var(--c-bg);
   background-color: var(--c-brand);
   border: 2px solid var(--c-brand);
   transition: background-color 0.1s ease;
 }
 
 .action.alt :deep(.item) {
-  background-color: #fff;
+  background-color: var(--c-bg);
```
posva, line 142: I wonder if we shouldn't just do
```suggestion
  background-color: transparent;
```

PR #267, opened 2021-03-23
docs: keyword vuepress change vitepress

qdbest:
This should be vitepress.


antfu:
I guess this was intended? Maybe instead of replacing, we can append "vitepress" to the list

qdbest:
I guess that it was copied from the project of vuepress, but was forgot to modify.

qdbest:
There are two mistakes in the file markdown.md.

antfu:
Don't feel the change to highlight.ts is necessary, reverted and merging the rest. Thanks

PR #276, opened 2021-04-01
fix(locales): use correct lang

posva:
- Correctly changes the lang attribute when switching languages
- Removes the limitation of prod only lang attribute
- Allows multi languages searches with DocSearch

### DIFF
--- a/src/client/app/index.ts
+++ b/src/client/app/index.ts
@@ -1,5 +1,12 @@
 import 'vite/dynamic-import-polyfill'
-import { App, createApp as createClientApp, createSSRApp, h } from 'vue'
+import {
+  App,
+  createApp as createClientApp,
+  createSSRApp,
+  h,
+  onMounted,
+  watch
+} from 'vue'
 import { inBrowser, pathToFile } from './utils'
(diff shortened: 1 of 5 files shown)


src/client/app/index.ts
```diff
@@ -15,6 +22,20 @@ const NotFound = Theme.NotFound || (() => '404 Not Found')
...
   name: 'VitePressApp',
   setup() {
+    const siteData = useSiteDataByRoute()
+
+    // change the language on the HTML element based on the current lang
+    onMounted(() => {
+      watch(
+        () => siteData.value.lang,
+        (lang: string) => {
+          console.log('lang', lang)
```
kiaking, line 32: Should we remove this 👀 ?

kiaking (commented):
Niiiice! Just made a comment!

src/client/app/index.ts
posva: Yeah, thanks 😆

kiaking (approved):
:tada: :tada: :tada:

posva:
BTW here is an example https://github.com/algolia/docsearch-configs/pull/3906

PR #277, opened 2021-04-01
feat(route): expose decoded hash

posva:
This PR has 2 things

- Exposing the **decoded** hash on the route. It could also be exposed encoded and I think this deserves some discussion:
  - The encoded version is more flexible but requires more work from the user
  - The decoded version works out of the box when using a `querySelector()` or `getElementById` (which we should be using if we decide to have this instead `querySelector()`)
- Always scrolling to a hash. This is to imitate native behaviour: when clicking an anchor twice, it should scroll to the position. We still avoid adding an entry to the history stack


kiaking:
Closing due to inactivity.

PR #283, opened 2021-04-08
fix(locales): use correct lang

posva:


posva:
Let's wait for the config to be merged on Vue Router next and make sure it works

kazupon:
I same issues vue-i18n-next docs.
I sent PR docsearch-configs (ref: https://github.com/algolia/docsearch-configs/pull/3957#pullrequestreview-635550867)

posva:
Tested with Vue router next, Pinia, and Vite

kazupon (approved):
~~LGTMです！~~
sorry, my mistake. NVM

src/client/theme-default/components/AlgoliaSearchBox.vue
```diff
@@ -54,11 +58,21 @@ function update(options: any) {
...
   docsearch(
     Object.assign({}, userOptions, {
       container: '#docsearch',
 
-      searchParameters: Object.assign({}, userOptions.searchParameters),
+      searchParameters: Object.assign({}, userOptions.searchParameters, {
+        // pass a custom lang facetFilter to allow multiple language search
+        // https://github.com/algolia/docsearch-configs/pull/3942
+        facetFilters: facetFilters.concat(
+          (userOptions.searchParameters || {}).facetFilters || []
```
meteorlxy, line 73: What about `userOptions.searchParameters?.facetFilters || []` ?

posva: I would have used it but the syntax is not transpiled and has very recent support only

meteorlxy: Not transpiled? I thought it would be 🤔

posva: oh, it's actually transpiled when building with vitepress! Not when building vitepress itself. It should be fine then

posva, line 73:
```suggestion
          userOptions.searchParameters?.facetFilters || []
```

kiaking (approved):
👍 👍 👍 👍 👍

PR #284, opened 2021-04-09
fix: include emoji text in nav link to match toc

sampullman:
I moved the emoji plugin after anchor/toc to avoid a mismatch between nav link and toc generation. The bug can be seen by clicking the `Emoji` link in the rendered TOC here: https://vitepress.vuejs.org/guide/markdown.html#table-of-contents

The desired behavior might actually be to strip the emoji text from the nav link, but that's a (slightly) more complicated change and breaks existing links.


posva (approved):
Thanks!

PR #288, opened 2021-04-18
fix(utils): add logic for image links

Nodreame:
### Description：
$withBase make 'base' before my img link
截屏2021-04-18_下午3_31_15

### Solution
use VuePress Logic：
截屏2021-04-18_下午4_00_50

### Result
截屏2021-04-18 下午4 03 40

### DIFF
--- a/src/client/app/utils.ts
+++ b/src/client/app/utils.ts
@@ -6,6 +6,7 @@ export { inBrowser }
  * Join two paths by resolving the slash collision.
  */
 export function joinPath(base: string, path: string): string {
+  if (path.charAt(0) !== '/') return path
   return `${base}${path}`.replace(/\/+/g, '/')
 }
 


antfu:
Probably better to handle here
https://github.com/vuejs/vitepress/blob/08b9e17a93cd56529a9cd7cbc73b64505e99a544/src/client/theme-default/composables/url.ts#L6-L6

Nodreame:
> Probably better to handle here
> https://github.com/vuejs/vitepress/blob/08b9e17a93cd56529a9cd7cbc73b64505e99a544/src/client/theme-default/composables/url.ts#L6-L6

Thank you for your reply~ 

I met this problem when I writing [Homepage](https://vitepress.vuejs.org/config/homepage.html) for my post blog. The sample code is shown below：

``` md
---
home: true
heroImage: http://img.nodreame.cn/2020-3-23-21-27-49
heroAlt: Logo image
heroText: Hero Title
---
```

Then I found the template of HomePage and found "$withBase"，so I modify "joinPath" below it.

image-20210529000817970

I read the code for Vitepress to follow your advice, then found that "withBase" was only called by "useNavLink".

Then I added ```console.log(path)``` for two files，"withBase" was not called.

image-20210529001452657

So I think its should be handle in "joinPath"，or just in "$withBase". 

Looking forward to your reply~ THX~

PR #290, opened 2021-04-24
fix: detect public folder for dead link

YunYouJun:
For example:

```md
[test](/test.html)
```

It will check dead link automatically. `[test](/test.html)` need `docs/test.md`, otherwise it will fail to build.

But we also can place `test.html` in `docs/public`, `docs/public/test.html`.
So we can add a rule for it. Just check it too.

```ts
path.resolve(dir, 'public', `${resolved}.html`)
```

### DIFF
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -56,7 +56,10 @@ export function createMarkdownToVueRenderFn(
             ? url.slice(1)
             : path.relative(root, path.resolve(dir, url))
         )
-        if (!pages.includes(resolved)) {
+        if (
+          !pages.includes(resolved) &&
+          !fs.existsSync(path.resolve(dir, 'public', `${resolved}.html`))
+        ) {
           console.warn(
             chalk.yellow(
               `\n(!) Found dead link ${chalk.cyan(
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -25,7 +25,11 @@ export function createMarkdownToVueRenderFn(
   const md = createMarkdownRenderer(root, options)
   pages = pages.map((p) => slash(p.replace(/\.md$/, '')))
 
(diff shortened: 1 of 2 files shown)


src/node/markdownToVue.ts
```diff
@@ -56,7 +56,10 @@ export function createMarkdownToVueRenderFn(
             ? url.slice(1)
             : path.relative(root, path.resolve(dir, url))
         )
-        if (!pages.includes(resolved)) {
+        if (
+          !pages.includes(resolved) &&
+          !fs.existsSync(path.resolve(dir, 'public', `${resolved}.html`))
```
antfu, line 61: Better to use `config.publicDir`

YunYouJun: Yeah, thanks. But i can not find how to use vite config in vitepress.
Can you give me a hint?

YunYouJun: How can i get `resolvedViteConfig` in `markdownToVue.ts`?
It seems that there is not a global configuration exported.
Should there be a better solution to get some global configuration in the function?

https://github.com/vuejs/vitepress/blob/77f144a0c75d6a238abd4a486d20c1e9a6888ab5/src/node/build/bundle.ts#L83

https://github.com/vuejs/vitepress/blob/77f144a0c75d6a238abd4a486d20c1e9a6888ab5/src/node/markdownToVue.ts#L20

antfu: You can get it using https://vitejs.dev/guide/api-plugin.html#configresolved in plugin.ts and pass it over

YunYouJun: Thanks. I submitted a new commit that uses `config.publicDir` instead of the `'public'` string.

YunYouJun:
@antfu Hello, is there any problem with this PR? Can it be merged?

antfu (approved):

PR #292, opened 2021-04-27
fix(algolia): avoid creating multiple algolia searches

posva:
Workaround until a destroy method is added

More at https://github.com/algolia/docsearch-configs/pull/3967#issuecomment-827565200

### DIFF
--- a/src/client/theme-default/Layout.vue
+++ b/src/client/theme-default/Layout.vue
@@ -7,7 +7,6 @@
             v-if="theme.algolia"
             :options="theme.algolia"
             :multilang="!!theme.locales"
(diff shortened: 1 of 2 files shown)


meteorlxy:
Is it possible to support different `placeholder` in different locales without re-creating docsearch instance?

https://github.com/algolia/docsearch/blob/bf18c7374e66581bd0b08b19f7a61b4a30bdcef3/packages/docsearch-react/src/DocSearch.tsx#L22

posva:
Maybe by passing an object and modifying it but at this point it's just too much to expect it to use the same object reference all the time 😆

PR #308, opened 2021-06-04
docs: # default theme change ##

cabbage9:
I think `#` should be a careless little mistake, change to `##` so that sidebar will be correct.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -32,10 +32,6 @@ module.exports = {
       {
         text: 'Release Notes',
         link: 'https://github.com/vuejs/vitepress/releases'
-      },
-      {
-        text: 'Translations',
-        items: [{ text: '中文', link: 'https://cn-vitepress.netlify.app/' }]
       }
     ],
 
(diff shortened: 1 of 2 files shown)


cabbage9:
I have translate whole docs about vitepress and I really want this will be added.

docs/.vitepress/config.js
```diff
@@ -32,6 +32,10 @@ module.exports = {
       {
         text: 'Release Notes',
         link: 'https://github.com/vuejs/vitepress/releases'
+      },
+      {
+        text: 'Translations',
+        items: [{ text: '中文', link: 'https://cn-vitepress.netlify.app/' }]
```
posva, line 38: Can you remove this from this PR and coordinate with the people behind https://cn.vitejs.dev/ to follow the same practices? I noticed a few problems with your translation in the landing page already:

- cn-vitepress instead of Vitepress
- Wrong copyright notice

posva (changes requested):
Thanks!

docs/.vitepress/config.js
antfu: /cc @ShenQingchuan @QC-L

cabbage9: OK, thanks for your reminding and help. I have removed the redundant codes and contacted @ShenQingchuan @ShenQingchuan.

PR #317, opened 2021-06-19
feat: Add before/after router hooks

khalwat:
Add before/after router hooks so that can be added via `enhanceApp()`, to allow for code to be executed when new pages are routed to

My specific use-case is to be able to fire Google Analytics tracking events after a new page is routed to, but there could be any number of cases where it might be useful to have these basic before/after router hooks

Closes: https://github.com/vuejs/vitepress/issues/318

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -9,9 +9,15 @@ export interface Route {
   component: Component | null
 }
 
+export interface RouterHooks {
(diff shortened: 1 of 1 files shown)


posva:
Thanks for the PR. Please see https://github.com/vuejs/vitepress/issues/318#issuecomment-864915561

PR #321, opened 2021-06-21
fix(nav): display nav if locales are present

m4dz:
This PR fix how the NavBar is shown. Right now, a `themeConfig.repo` key is mandatory to display the NavBar, regardless `nav` or `locales` config are defined.

### DIFF
--- a/src/client/theme-default/components/NavLinks.vue
+++ b/src/client/theme-default/components/NavLinks.vue
@@ -9,7 +9,7 @@ import NavDropdownLink from './NavDropdownLink.vue'
 const { theme } = useData()
 const localeLinks = useLocaleLinks()
 const repo = useRepo()
-const show = computed(() => theme.value.value || repo.value)
+const show = computed(() => theme.value.nav || repo.value || localeLinks.value)
 </script>
 
 <template>


m4dz:
Also, may be related to #322

src/client/theme-default/components/NavLinks.vue
```diff
@@ -9,7 +9,7 @@ import NavDropdownLink from './NavDropdownLink.vue'
 const { theme } = useData()
 const localeLinks = useLocaleLinks()
 const repo = useRepo()
-const show = computed(() => theme.value.value || repo.value)
+const show = computed(() => theme.value.nav || repo.value || localeLinks)
```
posva, line 12: It needs to read the `.value`, otherwise this would always be true

```suggestion
const show = computed(() => theme.value.nav || repo.value || localeLinks.value)
```

posva (changes requested):

src/client/theme-default/components/NavLinks.vue
m4dz: Damned, you're right! Thanks for the point!

posva (approved):
Thanks!

PR #328, opened 2021-06-30
fix: skip external URLs in `withBase`

zouhangwithsweet:
## Reason
I found that `withBase` will add path for all path, if I want to use some web resource. It's no need. So I add some path check for some url

## More
I have no idea for that there is necessary to check all `path` which is been  wrapped `withBase`.

### DIFF
--- a/src/client/app/utils.ts
+++ b/src/client/app/utils.ts
@@ -1,5 +1,5 @@
 import { siteDataRef } from './data'
-import { inBrowser } from '../shared'
+import { inBrowser, EXTERNAL_URL_RE } from '../shared'
 
 export { inBrowser }
 
@@ -11,7 +11,9 @@ export function joinPath(base: string, path: string): string {
 }
 
 export function withBase(path: string) {
-  return joinPath(siteDataRef.value.base, path)
+  return EXTERNAL_URL_RE.test(path)
+    ? path
+    : joinPath(siteDataRef.value.base, path)
 }
 
 /**
--- a/src/shared/shared.ts
+++ b/src/shared/shared.ts
@@ -8,6 +8,8 @@ export type {
   Header
 } from '../../types/shared'
 
+export const EXTERNAL_URL_RE = /^https?:/i
+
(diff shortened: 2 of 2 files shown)


posva:
Maybe is worth of [withBase](https://github.com/vuejs/vitepress/blob/master/src/client/app/utils.ts#L13) to skip paths that start with `http`.

BTW, in a regex you can write `c?` instead of `c{0,1}`

kiaking:
Yeah I also think `withBase` skipping URL starting `https?://` sounds a better idea 👀

zouhangwithsweet:
> Maybe is worth for [withBase](https://github.com/vuejs/vitepress/blob/master/src/client/app/utils.ts#L13) to skip paths that start with `http`.
> 
> BTW, in a regex you can write `c?` instead of `c{0,1}`

OK, sir

zouhangwithsweet:
> Yeah I also think `withBase` skipping URL starting with `https?://` sounds a better idea 👀 I wouldn't think it will break anything, if anything, it's breaking URLs for people so 😅

I will make changes according to your suggestions. :)

kiaking (approved):

src/client/app/utils.ts
```diff
@@ -11,6 +11,7 @@ export function joinPath(base: string, path: string): string {
 }
 
 export function withBase(path: string) {
+  if (/^(https?:)?\/\//.test(path)) return path
```
posva, line 14: let's move the regex to a const outside `const EXTERNAL_URL_RE = /^https?:/i`. Which should be enough (using `//external.com` is an antipattern and should be avoided

Then use a ternary: `return re.test(path) ? path : join(...)

posva (changes requested):

posva:
@kiaking I think we should move the RE in a shared file. I saw multiple occurrences of it across the codebase

src/client/app/utils.ts
zouhangwithsweet: Great

posva (approved):
Thanks!

PR #338, opened 2021-07-20
chore: update vue

posva:
Updating Vue to test what is and isn't working

- Needed a `reactive` call to unwrap the `useData()` object. It was otherwise erroring on cyclic dependencies


yyx990803:
Turns out I was stringify raw computeds in `Debug` and didn't notice. It is now throwing an error because we adjusted the TS codegen so `effect` is now exposed as an enumerable property on computed refs which in this case leads to circular references.

While the fix works, I think it's better to make `toDisplayString` handle raw refs (see 68e8c2fd)

posva:
Great! I updated vue and removed the use of reactive and everything works

PR #340, opened 2021-07-21
fix(edit-link): let frontmatter overwrite global editLink

### LINKED ISSUE #334: Edit link can't be disabled via frontmatter
### Describe the bug

When edit link is enabled in the config, disabling it via frontmatter on a page is not possible.
This seems to be caused here https://github.com/vuejs/vitepress/blob/master/src/client/theme-default/composables/editLink.ts#L19.

Intending to open a PR if the expected behaviour is agreed upon.

### Reproduction

- Set `editLink` to `true` in the config
- Add `editLink: false` to the frontmatter of a documentation page
- The `Edit this page` link still shows up

### Expected behavior

Expecting that when either `editLink: false` or `editLink: "false"` is passed in via frontmatter, that the edit link is not rendered on that page.

### System Info

```shell
System:
  OS: Linux 5.10 Manjaro Linux
  CPU: (8) x64 Intel(R) Core(TM) i7-8809G CPU @ 3.10GHz
  Memory: 19.23 GB / 31.28 GB
  Container: Yes
  Shell: 5.1.4 - /bin/bash
Binaries:
  Node: 16.0.0 - ~/.config/nvm/versions/node/v16.0.0/bin/node
  Yarn: 1.22.5 - ~/.yarn/bin/yarn
  npm: 7.10.0 - ~/.config/nvm/versions/node/v16.0.0/bin/npm
npmPackages:
  vitepress: 0.15.6 => 0.15.6
```

### Additional context

_No response_

### Validations

dirkdev98:
Allows overwriting the `editLinks` setting via the frontmatter `editLink` property.

Closes #334

### DIFF
--- a/src/client/theme-default/composables/editLink.ts
+++ b/src/client/theme-default/composables/editLink.ts
@@ -16,7 +16,10 @@ export function useEditLink() {
       editLinks
     } = theme.value
 
-    const showEditLink = frontmatter.value.editLink || editLinks
+    const showEditLink =
+      frontmatter.value.editLink !== undefined
+        ? frontmatter.value.editLink
+        : editLinks
     const { relativePath } = page.value
 
     if (!showEditLink || !relativePath || !repo) {
--- a/src/client/theme-default/composables/editLink.ts
+++ b/src/client/theme-default/composables/editLink.ts
@@ -17,7 +17,7 @@ export function useEditLink() {
     } = theme.value
 
     const showEditLink =
-      frontmatter.value.editLink !== undefined
+      frontmatter.value.editLink != null
         ? frontmatter.value.editLink
         : editLinks
     const { relativePath } = page.value


src/client/theme-default/composables/editLink.ts
```diff
@@ -16,7 +16,10 @@ export function useEditLink() {
       editLinks
     } = theme.value
 
-    const showEditLink = frontmatter.value.editLink || editLinks
+    const showEditLink =
+      frontmatter.value.editLink !== undefined
+        ? frontmatter.value.editLink
+        : editLinks
```
posva, line 22:
```suggestion
    const showEditLink =
      frontmatter.value.editLink != null
        ? frontmatter.value.editLink
        : editLinks
```

posva (approved):
Thanks!

Issue #346, opened 2021-07-29
margin left/right is zero when `home: true` in default theme


gcnyin:
### Describe the bug

margin left/right is zero when `home: true`
![WechatIMG2403](https://user-images.githubusercontent.com/53973962/127427203-5a7be0f7-505f-4428-a3a4-f0f6ce9be371.jpeg)

### Reproduction

1. start a new vitepress project
2. in `docs/index`, set `home: true`
3. open in mobile

### Expected behavior

should have left and right margin

### System Info

```shell
System:
    OS: macOS 11.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 293.35 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.22.1 - ~/.nvm/versions/node/v12.22.1/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v12.22.1/bin/yarn
    npm: 7.17.0 - ~/.nvm/versions/node/v12.22.1/bin/npm
  Browsers:
    Chrome: 92.0.4515.107
    Firefox: 84.0
    Safari: 14.1.2
  npmPackages:
    vitepress: ^0.15.6 => 0.15.6
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #347, opened 2021-07-29
fix(css): remove 720px breakpoint in home layout

gcnyin:
Close https://github.com/vuejs/vitepress/issues/346


posva (approved):
This does make the spacing consistent with the rest of the content on the home page

PR #348, opened 2021-08-04
feat: allow displaying fallback when ads are blocked

posva:
The idea behind this is to enable a way to remind developers to safelist docs in adblockers like in Nuxtjs or JSFiddle:

Screen Shot 2021-08-04 at 17 58 34

Currently, this exposes a slot named `carbon-ads-blocked` at the Layout level to let the user customize how this is displayed.


kiaking:
We might wanna need this but it has been quite some time and the code has changed a lot so let's start over 😅

PR #364, opened 2021-08-14
fix: to make config hmr work in window

CHOYSEN:


posva:
Do tou have a repro?

CHOYSEN:
Just use Vitepress in Windows that can repro it.

The following line will resolve configPath that use backslash in Windows:`c:\test-vitepress\docs\.vitepress\config.js`
https://github.com/vuejs/vitepress/blob/e74c5f06d1d5890fad6dd728df9bf85dcfda87d1/src/node/config.ts#L102

The `file` is normalized by Vite:`c:/test-vitepress/docs/.vitepress/config.js`, so they never equal in Windows.
https://github.com/vuejs/vitepress/blob/e74c5f06d1d5890fad6dd728df9bf85dcfda87d1/src/node/plugin.ts#L204

PR #365, opened 2021-08-14
fix: use useData() instead of $site

CHOYSEN:
All global mixin properties (e.g. $site) except $frontmatter are removed. Always use useData() to retrieve VitePress data in Vue components.


posva (approved):
Thanks!

PR #366, opened 2021-08-14
fix: fallback to base url when visit a non-exist page at first

CHOYSEN:


posva:
Right now we show a 404 page, that's much more desirable behavior. Maybe it should be added in dev too.

CHOYSEN:
When you first visit a normal page and then jump to a non-existent page, 404 pages will indeed be displayed. However, if you visit a non-existent page at the beginning, the whole application will display a white screen.

posva:
Then let's display the 404 in dev mode instead.
In production, it has to be configured on the server anyway

Issue #330, opened 2021-07-02
Sidebar not showing anything.


hadihammurabi:
### Describe the bug

This is my site with empty sidebar.
![image](https://user-images.githubusercontent.com/16364286/124286102-cbff5980-db78-11eb-90cd-b40c1007bc9e.png)

### Reproduction

- Create a new vitepress project with some directories and md files.
- Run `npx vitepress`

### Expected behavior

Sidebar displays all pages in a tree or how they should be.

### System Info
![image](https://user-images.githubusercontent.com/16364286/124286525-40d29380-db79-11eb-9451-836c8fd8637f.png)
![image](https://user-images.githubusercontent.com/16364286/124286680-65c70680-db79-11eb-8788-d2ad929b8983.png)

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

sosnail:
@hadihammurabi

I found that the sidebar must have a secondary title.

fz6m:
> @hadihammurabi
> 
> I found that the sidebar must have a secondary title.

Yes, I don't think it's friendly enough
Hope support the display of deeper titles （ like `sidebarDepth` ）

newur:
>     * Create a new vitepress project with some directories and md files.

This sounds like you expect vitepress to show these files / directories automatically - at least that is what I expected. However, vitepress requires a configuration in `.vitepress/config.js` and (mostly) manual setup of the files/dirs. See https://github.com/vuejs/vitepress/blob/c1ba483e587eab1cb7b1dd828c340700d9c01532/docs/.vitepress/config.js#L38 for an example.

I wrote a little plugin to have an automatic way of resolving this, inspired by https://github.com/hannoeru/vite-plugin-pages

Would love to bring my code upstream, but not sure how to best address this, as it is a bit.. rough?

Btw I found a video/stream recording that shows how others struggle with this a well #144 

https://www.youtube.com/watch?v=Gx2aMCwQ3Rg (not affiliated in any way)

PR #384, opened 2021-09-08
feat(theme) add example with file-based-sidebar plugin

newur:
Draft PR for a plugin that auto-generates the sidebar from files / dirs. Inspired by the excellent https://github.com/hannoeru/vite-plugin-pages

The current manual approach is sometimes a bit confusing, see also my comment here https://github.com/vuejs/vitepress/issues/330#issuecomment-915058408

Note: I tried my best to understand how rollup / vite / vitepress work together. However, I did not work with this particular stack before, so the solution might be hacky a bit (or a lot, don't know). So please consider it as a working showcase. :)

### DIFF
--- a/examples/pages/docs/plugin/vitepress-pages.ts
+++ b/examples/pages/docs/plugin/vitepress-pages.ts
@@ -1,0 +1,110 @@
+import fg from 'fast-glob'
+import * as fsAsync from 'fs/promises'
+import path from 'path'
+import { resolveSiteData } from 'vitepress/dist/node/config'
+import { HmrContext } from 'vite'
+
+let root = ''
+
+export default function vitepressPages(docDir?: string) {
+  const virtualFileId = '/@siteData'
+  return {
+    name: 'vitepress-pages', // required, will show up in warnings and errors
+    resolveId(id) {
+      if (id === virtualFileId) {
+        return virtualFileId
+      }
+    },
+    async load(id) {
+      if (id === virtualFileId) {
+        root = docDir ?? 'docs'
+        const conf = await resolveSiteData(root)
+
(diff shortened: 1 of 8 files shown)


examples/pages/docs/plugin/vitepress-pages.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  async function findFirstH1(filepath: string) {
+    const file = await fsAsync.readFile(filepath, 'utf8')
+    const data = file.toString()
+      .split('\n')
+      .filter(l => l.startsWith('# '))
+
+    return data[0]
```
stefnotch, line 73: This will break when a markdown file doesn't have a header. I recommend replacing it with `return data[0] ?? "Missing header";` or similar

examples/pages/docs/plugin/vitepress-pages.ts
```diff
@@ -0,0 +1,110 @@
...
+      // not sure if this can plugin can support hot reloading
+    }
+  }
+
+  async function resolvePages() {
+    const mdFiles = findMdFiles()
+    const dirToFiles = createMultiMap(mdFiles)
+
+    const sidebarItems = []
+    for (let [dir, files] of dirToFiles) {
```
stefnotch, line 40: Is the sorting order for this something one can rely on? If not
```js
const dirToFilesSorted = [...dirToFiles.entries()];
    dirToFilesSorted.sort((a, b) => a[0].localeCompare(b[0]));
    for (let [dir, files] of dirToFilesSorted) {
```

kiaking:
Closing due to inactivity. Thanks for the PR!

PR #387, opened 2021-09-08
docs: add sidebar docu with examples

newur:
I was a bit lost  - and according to youtube and some issue here, others as well - how to populate the sidebar. So I mostly copy and pasted the vitepress config from this repo into the existing docs.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -1,5 +1,7 @@
 # Configuration
 
+## Basics
+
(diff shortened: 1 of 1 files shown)


brc-dd:
Hi! We have rewritten the docs. Check them out here: https://vitepress.vuejs.org/guide/theme-sidebar.html. If you find there is something missing, please let us know. PRs are welcome too.

Issue #241, opened 2021-02-16
Ability to extend vite config


TheDutchCoder:
**Is your feature request related to a problem? Please describe.**
Currently there seems to be no way to extend the Vite config used by vitepress (or at least it's not clear how to do it).
It would be great to have that ability to specify aliases that are used inside of components that are not part of VitePress.

In our case we have a component library in `src`, and the docs in `docs`, but components imported in VitePress use `#components/button` imports (for example) and VitePress can't resolve those as there's no way to set up aliases for it.

**Describe the solution you'd like**
A way to extend vite config and/or specify aliases.

**Describe alternatives you've considered**
I tried creating a vite config in the root, but that prevents TS based SFC's from compiling properly (I think it overwrites the internal config).

**Additional context**
A quick screenshot of our setup:
![image](https://user-images.githubusercontent.com/1668207/108101779-b79d4300-7055-11eb-9af8-0b74a4382327.png)

`docs/.vitepress` hosts VitePress, but the root is Vite itself (library mode).

TheDutchCoder:
I do see an alias option in here: https://github.com/vuejs/vitepress/blob/master/src/node/config.ts which might be what I'm looking for, though I haven't been able to get it to work yet.

yyx990803:
If you put a `vite.config.js` in the VitePress (docs) root it will be picked up by VitePress.

TheDutchCoder:
Yeah I discovered that, but then it'll fail to compile SFC's. I assume because it needs some additional config for that, but wouldn't I at that point be rewriting whatever VitePress has built in?

yyx990803:
Hmm, not sure. It works as expected in the Vue blog: https://github.com/vuejs/blog/blob/master/vite.config.js

TheDutchCoder:
🤔 strange. Let me experiment with this some more.

Should it just merge the configs though?
In my use case I wanted the aliases to work (VitePress and Vite have slightly different configs for that) but I couldn't get it to work (Vite works fine with `resolve.alias`, but VitePress does not).

kiaking:
@TheDutchCoder Do you have any updates on this? 👀  If not, maybe we could close this one for now and open another issue if you encounter the issue again.

antfu:
Maybe due to this https://github.com/vitejs/vite/pull/2847

kiaking:
Oh OK cool. Let's see how it goes with that issue then.

kiaking:
[...]

PR #394, opened 2021-09-16
feat: support user-provided vite plugins

fnlctrl:
Currently plugins defined in `.vitepress/config.js` are ignored, e.g.
```js
module.exports = {
  vite: {
    plugins: [
      myPlugin() // doesn't work
    ]
  }
}
```

Therefore, `userViteConfig.plugins` is appended to `createVitePressPlugin` result to support custom vite plugins

### DIFF
--- a/src/node/plugin.ts
+++ b/src/node/plugin.ts
@@ -273,5 +273,5 @@ export function createVitePressPlugin(
     }
   }
 
-  return [vitePressPlugin, vuePlugin]
+  return [vitePressPlugin, vuePlugin, ...(userViteConfig.plugins || [])]
 }


lencx:
@fnlctrl I found the solution here https://github.com/vuejs/vitepress/issues/241#issuecomment-780167185

example: [lencx/code-snippets/docs/vite.config.js](https://github.com/lencx/code-snippets/blob/main/docs/vite.config.js)

```
[root]
|- [docs]
|	 |- [.vitepress]
|	 |- vite.config.js
|	 `- ...
```

fnlctrl:
@lencx Thanks for the workaround! Didn't know it was possible.
 
The `vite` option was introduced in https://github.com/vuejs/vitepress/commit/3737b1055dc1145dc70b10994564c6d83affd15d so I think plugins defined in it should also work

PR #395, opened 2021-09-20
On Dead Link, Add info about them

romain130492:
## Description
* When having deadlinks I thought It would be good to know which links are dead... otherwise it's quite a burden to find them .
* I added 2 logs:
  * One to get the deadLink
  * One to get the deadLink's file path 

@posva 😀


posva:
Sorry I don't remember where the initial conversation was but I don't think we need this anymore.

PR #399, opened 2021-09-24
fix: tolerant invalid hash

antfu:
In some cases, users might have custom logic to control the hash and possibly make the hash an invalid selector. This PR makes the router tolerant in those cases.


posva (approved):
LGTM!
I think this is good but I wrote about possible outcomes when using querySelector vs getElementById at https://github.com/vuejs/vue-router-next/blob/master/src/scrollBehavior.ts#L85 in case it's an interesting topic

PR #400, opened 2021-09-24
fix(default-theme): navLink matches active but lacks withBase

kt3721:


yyx990803:
Should be fixed by #342

PR #410, opened 2021-10-08
docs: add config.ts example

YunYouJun:
Related to https://github.com/vuejs/vitepress/commit/d3b1521ebef831e0d0307b3b12e4fc1f6ce4721a.

I have been looking forward to this feature for a long time, and I think it would be better if it can also be written in the documentation.


YunYouJun:
Is there anything else that needs to be modified?

kiaking:
Thanks for the PR! This is already documented in `next` branch 👍

Issue #424, opened 2021-10-22
build errors when file or folder named in Chinese and used in link


fuchunhui:
### Describe the bug

In docs folder, some files named in Chinese, that make errors, when `yarn build `.

```
API Extractor completed successfully
vitepress v0.20.0
⠙ building client + server bundles...
(!) Found dead link /guide/%E5%A4%8D%E5%88%B6 in file /Users/user/github/vitepress/docs/index.md
⠙ building client + server bundles...[vitepress] One or more pages contain dead links.
✖ building client + server bundles...
build error:
 Error: One or more pages contain dead links.
    at Object.renderStart (/Users/user/github/vitepress/dist/node/serve-6eb63a16.js:36672:15)
    at /Users/user/github/vitepress/node_modules/rollup/dist/shared/rollup.js:22620:25
    at async Promise.all (index 0)
    at async Bundle.generate (/Users/user/github/vitepress/node_modules/rollup/dist/shared/rollup.js:15901:13)
    at async handleGenerateWrite (/Users/user/github/vitepress/node_modules/rollup/dist/shared/rollup.js:23400:23) {
  code: 'PLUGIN_ERROR',
  plugin: 'vitepress',
  hook: 'renderStart'
}
```

### Reproduction

yes, just change the file name or folder name to Chinese.

1. copy a file, named to Chinese. `cp docs/guide/api.md docs/guide/复制.md`
2. open `docs/index.md`, append a link.
```
## Chinese test
[测试中文](/guide/复制)'
```
3. `yarn docs-build`.

### Expected behavior

build completed successfully, no errors.

### System Info

```shell
System:
  OS: macOS 10.15.3
  CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
  Memory: 137.25 MB / 8.00 GB
  Shell: 5.7.1 - /bin/zsh
Binaries:
  Node: 16.8.0 - /usr/local/bin/node
  Yarn: 1.22.10 - /usr/local/bin/yarn
  npm: 7.21.0 - /usr/local/bin/npm
Browsers:
  Chrome: 95.0.4638.54
  Safari: 13.0.5
```

### Additional context

vitepress@0.20.0

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #425, opened 2021-10-22
fix: Chinese file link build failed (#424)

fuchunhui:


Xu-pixel:
but in v0.22.3, it is still not work.
image

fuchunhui:
Just tried this and didn't reproduce your problem. you can [try here](https://github.com/fuchunhui/test-vitepress)
I guess that the possible reason is the path problem of your file.

Here are some tips:
* Ask questions needed in the issue area.
* Provide to a demo that can be reproduced so that others can help you solve it more quickly.
* More detailed description of the problem, step by step.

Demo for you.
image

Xu-pixel:
Thank you for your reply.It can work well in dev environment but error in build.

I downloaded your project and used yarn docs:build:
image

Xu-pixel:
Thank you !

PR #429, opened 2021-10-28
fix: fix render __VP_STATIC_START__

huyanhai:
fix: fix render __VP_STATIC_START__


yyx990803:
This doesn't look like the right fix.

PR #448, opened 2021-11-20
feat: user configurable `outDir`

### LINKED ISSUE #171: How to configure the build directory of vitepress ?

Mando75:
Hi, this is my first time opening a PR in Vitepress, so please let me know if I need to change anything. I did my best to follow the contributing guidelines. 

## Changes
- Adds a new `outDir` option to the UserConfig in .vitepress/config.js.
   - User-provided `outDir` is resolved the same way as `srcDir` to keep things consistent.
   - The existing behavior is used as fallback if no `outDir` is provided by the user.
   - I chose to name it `outDir` instead of `dest` (from Vuepress) because the `outDir` option already existed on the internal SiteConfig type
- Updates the GitLab deployment docs to use the new `outDir` to publish files to the top-level `public` directory in accordance with [GitLab pages deployment requirements](https://docs.gitlab.com/ee/user/project/pages/getting_started/pages_from_scratch.html#specify-the-public-directory-for-output) 
- Adds a .gitignore item for `.idea`, the `.vscode` equivalent for JetBrains IDE's (Webstorm, PHPStorm, etc.). I am a WebStorm user, and I figured it may make contributing easier for other developers using JetBrains products.
- Fixes a typo in a comment I noticed when making these changes. 

## Motivation

- There was an issue requesting this feature recently updated that shows there is interest. #171
- A similar option [exists in Vuepress](https://vuepress.vuejs.org/config/#dest)
- This would make my life easier for deploying my Vitepress site to GitLab :wink: 

## Notes
I included both the feature and doc changes in a single PR since I thought they should be updated together. If you would prefer them split into multiple PRs for better tracking of feat/doc changes, I can do that. 

I looked for any tests I should update, but didn't find any. If I missed them, please let me know and I can add a test case. 

I have tested it in my Vitepress project using `npm link`, and it works great. I recorded a video demo that I can provide if you would like to see a demo. 

close #171


src/node/config.ts
```diff
@@ -95,6 +96,7 @@ export async function resolveConfig(
   const [userConfig, configPath] = await resolveUserConfig(root, command, mode)
   const site = await resolveSiteData(root, userConfig)
   const srcDir = path.resolve(root, userConfig.srcDir || '.')
+  const outDir = path.resolve(root, userConfig.outDir || resolve(root, 'dist'))
```
yyx990803, line 99:
```suggestion
  const outDir = userConfig.outDir ? path.resolve(root, userConfig.outDir) : resolve(root, 'dist')
```

Issue #464, opened 2021-12-11
[Feature Request] Improve TypeScript support for config file.


ulivz:
### Is your feature request related to a problem? Please describe.

Make configuration easier.

### Describe the solution you'd like

Same as [vite](https://vitejs.dev/config/#config-intellisense) :

```ts
import { defineConfig } from 'vitepress'

export default defineConfig({
   lang: 'en-US',
   // ...
}) 
``` 

But we need to improve the support for theme config:

By default, `defineConfig` helper leverages the theme config type from default theme:

```js
import { defineConfig } from 'vitepress'

export default defineConfig({
  themeConfig: {
     // Type is `DefaultTheme.Config` 
  }
})
```

If you use a custom theme, you'll be able to pass the generics type for your custom theme, and you need overload it with the second parameter of `defineConfig` helper:

```js
import { defineConfig } from 'vitepress'
import { ThemeConfig } from 'your-theme'

export default defineConfig<ThemeConfig>({
  themeConfig: {
    // Type is `ThemeConfig` 
  }
}, true); // declare `usingCustomTheme` and discard usage of the default theme.
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/master/.github/contributing.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #465, opened 2021-12-11
feat: improve typescript support for config file

ulivz:
Close: #464


types/index.d.ts
```diff
@@ -1,4 +1,4 @@
 export * from './shared'
+export * from './default-theme'
 export * from '../dist/node/index'
 export * from '../dist/client/index'
-export * from '../dist/client/theme-default/config'
```
aamayamaa, line 4: Remove this path since it does not exists: https://unpkg.com/browse/vitepress@0.20.4/dist/client/

src/client/theme-default/config.ts
```diff
@@ -1,146 +1 @@
-export namespace DefaultTheme {
```
aamayamaa, line 1: Move this file to shared types since we need it between server and client.

src/node/config.ts
```diff
@@ -82,7 +85,12 @@ const resolve = (root: string, file: string) =>
 /**
  * Type config helper
  */
-export function defineConfig(config: RawConfigExports) {
+ export function defineConfig<T extends ThemeConfig = ThemeConfig>(
+  config: UserConfig<T>,
+  usingCustomTheme: true
```
aamayamaa, line 90: Motivation here: TypeScript cannot overload a function type inferring based on the number of generic type parameters, so I added an additional function parameters here.

aamayamaa:
@yyx990803 Could you help review this pull request?

yyx990803:
An extra argument to `defineConfig` feels too implicit, so I changed to export a separate method `defineConfigWithTheme`.

PR #486, opened 2022-01-06
chore(LastUpdated): ISO 8601 format, <time>

simon04:
"Last Updated:2022-01-06 10:22:35 UTC"


kiaking:
Maybe it might be nice to make this bit configurable, but I think it's better to have default set to `localeString` for now 👍

Issue #219, opened 2021-02-03
Configurable 'clean url' in client router


jalt007:
**Is your feature request related to a problem? Please describe.**

In production it may be desirable to strip the `.html` extension in urls. Of course this isn't a direct concern of vitepress, and typically static hosts provide a turnkey setting to do this e.g; https://vercel.com/docs/configuration#project/clean-urls

However the client side router will append `.html` to the history push state when navigating:

- https://github.com/vuejs/vitepress/blob/master/src/client/app/router.ts#L47

This will result in an extra server redirects when e.g. sharing urls, doing hard reloads; that ideally could be avoided.

**Describe the solution you'd like**

To be able to configure this `.html` appending behaviour on the client router :)

kiaking:
Thanks for the feedback! Interesting. Sounds nice, but the concern would be it might be misleading for people who are not using advanced hosting such as Vercel. If we have option like `cleanUrl: true`, then people might think VitePress is going to generate clean url directly structure.

```bash
/index.html
/guide/getting-started/index.html
```

I guess we need to be extra careful when designing spec if we were to add this feature 🤔

GamearooCodes:
This would be a nice feature to be able to have it as /about instead of /about.html

talawahtech:
FYI the issue mentioned by @jalt007 happens with Cloudflare Pages, but unlike vercel, there is no way to run it off. It also causes web analytics numbers to be duplicated.

https://developers.cloudflare.com/pages/platform/serving-pages#route-matching

PR #488, opened 2022-01-06
feat(cleanUrls): CleanUrls feature - No .html in navigation bar

georges-gomes:
This feature is meant to be working with SPA and MPA mode so it's a little bit bigger than just rename the browser location.

- New option: `cleanUrls` (default false)
- Generates `/foo.md` -> `/foo/index.html` (This works best for MPA mode with most static site hosting services)
- Internal links in `.md` files are not transformed to `.html`
- When on SPA mode, the client router doesn't rename `/foo` into `/foo.html` anymore.

First PR on Vitepress so please send comments, I will adjust.

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -17,6 +17,7 @@ import anchor from 'markdown-it-anchor'
 import attrs from 'markdown-it-attrs'
 import emoji from 'markdown-it-emoji'
 import toc from 'markdown-it-table-of-contents'
+import { SiteConfig } from 'config'
 
 export interface MarkdownOptions extends MarkdownIt.Options {
   lineNumbers?: boolean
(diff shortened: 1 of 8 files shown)


georges-gomes:
Fixes #219

src/node/markdown/markdown.ts
```diff
@@ -48,7 +49,7 @@ export interface MarkdownRenderer extends MarkdownIt {
 export type { Header }
 
 export const createMarkdownRenderer = (
-  srcDir: string,
+  siteConfig: SiteConfig,
```
yyx990803, line 52: `createMarkdownRenderer` is in fact an exported API used in some custom themes, so this would be a breaking change. I think we can move `cleanUrls` as an optional 3rd parameter instead.

yyx990803 (changes requested):

georges-gomes:
Hi @kiaking 
Since you seem to be working for "next" release.
Do you want this PR for it?
I have the feeling this is an important feature for some people.
(including me obviously)
Let me know what should be done.
Thanks

kiaking:
@georges-gomes Yes! That would be super cool 🙌  I think this is a nice feature to have 👍

georges-gomes:
@kiaking should I try to rebase on `next` branch?

kiaking:
Yes please! Let's see if it works. No change in theme file, so it might just work fine 👀

georges-gomes:
Replaced by #869

Issue #483, opened 2022-01-04
`dist\client\theme-default\config.d.ts` is not getting bundled


brc-dd:
### Describe the bug

There is no `config.d.ts` available here: https://cdn.jsdelivr.net/npm/vitepress@0.20.10/dist/client/theme-default/

It is required by: https://cdn.jsdelivr.net/npm/vitepress@0.20.10/types/index.d.ts (line 4)

### Expected behavior

`config.d.ts` to be present in `dist/client/theme-default/` so that we can reference `DefaultTheme` in `.vitepress/config.ts`.

### Additional context

Something needs to be changed with `api-extractor` as `tsc` (first part of `build-types-client` script) is generating the files perfectly in `dist/temp`.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
Just noticed that #465 will also fix this. #484 can be closed without merging if the former is merged.

brc-dd:
@yyx990803 `theme.d.ts` (the one at project root for `vitepress/theme`) is still not in `package.json`'s `files` array. It isn't being bundled too.

PR #489, opened 2022-01-07
fix: export default theme type

Estelle00:


brc-dd:
Fixes: https://github.com/vuejs/vitepress/issues/483#issuecomment-1006273672

yyx990803:
I realized this file is no longer necessary - you can just do `import { DefaultTheme } from 'vitepress'`.

brc-dd:
> I realized this file is no longer necessary - you can just do `import { DefaultTheme } from 'vitepress'`.

@yyx990803 That's just the `DefaultTheme` interface, not the actual defaultTheme.

yyx990803:
Thanks, #494 is close but the `Theme` interface is meant as a constraint type instead of the actual type of the default theme. See https://github.com/vuejs/vitepress/commit/eabf6d2aa69d2a5452042bbb59edbbbc95aece87

PR #496, opened 2022-01-13
fix(theme): color for info custom block

edimitchel:
Seen on Vitest docs:
On dark, info custom block is light, but the text color remains light (as the body color is light).

Capture d’écran 2022-01-13 à 12 05 43
https://vitest.dev/config/#globalsetup


patak-dev (approved):

patak-dev:
We can actually fix this in the Vitest theme @edimitchel, the official theme for VitePress will change soon and properly support dark mode.

PR #498, opened 2022-01-15
feat: rendering meta tags for preview images (twitter/og)

jbandi:
For some use-cases it might be desired to have a preview image for social media.

With this PR a preview image can be specified on site level (.viteress/config.ts) and also per page (frontmatter).

The images have to be added to `public/images`, vitepress is not putting them there.

resolves: #504

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -4,6 +4,7 @@ export default defineConfig({
   lang: 'en-US',
   title: 'VitePress',
   description: 'Vite & Vue powered static site generator.',
+  banner: './images/vite.png',
 
   head: [
     [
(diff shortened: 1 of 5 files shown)


docs/.vitepress/config.ts
```diff
@@ -4,6 +4,7 @@ export default defineConfig({
   lang: 'en-US',
   title: 'VitePress',
   description: 'Vite & Vue powered static site generator.',
+  banner: './images/vite.png',
```
yyx990803, line 7: Should this be `socialImage` instead?

src/node/config.ts
```diff
@@ -33,6 +33,7 @@ export interface UserConfig<ThemeConfig = any> {
   base?: string
   title?: string
   description?: string
+  socialImage?: string
```
yyx990803, line 36: I don't think we need a site-wide config option for this, since you can already do that using `head`.

yyx990803 (changes requested):

kiaking:
Closing in favor of https://github.com/vuejs/vitepress/issues/551

PR #501, opened 2022-01-19
fix: append base to links

### LINKED ISSUE #252: Base paths (and links) not generating correctly for nested pages.
**Describe the bug**

See https://github.com/songololo/vitepress-debug for reproduction of this issue:

Run a dev server: `yarn docs:dev`.

Note that the base-path `/my-base-path/` is set in the config at `.vitepress/config.js`.

The base path works for root files, i.e. `index.md` is served from `/my-base-path/`.

However, the base path is missing for nested files, i.e. `content/index.md` is served from `/content/` and `content/hello.md` is served from `/content/hello.html`.

This doesn't break things in development mode because the links are also missing the base path.

However, it breaks things in production because the links are missing the base path.

**System Info**
- vitepress 0.12.2
- vite/2.0.4
- darwin-x64
- node-v15.8.0

This happens for both my local system and for a github actions build.

brc-dd:
fixes #252

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -49,7 +49,8 @@ export type { Header }
 
 export const createMarkdownRenderer = (
   srcDir: string,
-  options: MarkdownOptions = {}
+  options: MarkdownOptions = {},
+  base: string
 ): MarkdownRenderer => {
   const md = MarkdownIt({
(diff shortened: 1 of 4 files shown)


brc-dd:
It is fixing that linked issue but causing some others.

Issue #458, opened 2021-12-03
Links to pages outside "base" path scope do not work


aantipov:
### Describe the bug

I have the following site structure

![image](https://user-images.githubusercontent.com/427621/144564393-262a3544-c93f-40c4-a18b-dbadb37a9997.png)

`/blog` is powered by VitePress (I set `base: '/blog/'` in `/blog/.vitepress/config.js`)

The problem is that links from `/blog` pages to other pages of the website do not work

![image](https://user-images.githubusercontent.com/427621/144565176-08b33186-cf68-4a7b-bf4e-1efedfae25ea.png)

I checked the console and saw 404 for js bundles. And no document requests

### Reproduction

Described above

### Expected behavior

Links to pages pages outside `/<basePath>/` should work.

As I understood, VitePress treats all the links under `/` as internal and applies routing and prefetching.
This is the root of the problem.
Scoping routing and prefetching to `/<basePath>/` should fix the problem.

Here is where an additional check for `basePath` is needed for Router
https://github.com/vuejs/vitepress/blob/main/src/client/app/router.ts#L125
![image](https://user-images.githubusercontent.com/427621/144566953-51ba7668-dcde-433d-ab63-93a92768baf0.png)

Here is where an additional check for `basePath` is needed for preFetch https://github.com/vuejs/vitepress/blob/main/src/client/app/composables/preFetch.ts#L83 
![image](https://user-images.githubusercontent.com/427621/144567247-1f7fd1d3-e413-41cd-acc4-10518e9e2fa4.png)

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 323.38 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.1.4 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 95.1.31.88
    Chrome: 96.0.4664.55
    Firefox: 90.0.2
    Safari: 15.1
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
[...]

PR #502, opened 2022-01-20
fix: append base to links

### LINKED ISSUE #252: Base paths (and links) not generating correctly for nested pages.

brc-dd:
fixes #252

As mentioned at https://github.com/vuejs/vitepress/issues/252#issuecomment-1000024504, relative links are still working. If you guys want to keep it that way only and map `/foo` to `scheme://path/foo` instead of `scheme://path/base/foo` then the current external URL testing needs to be changed as `/foo` may not point to valid internal link if `base` is set (ref: #458). If this is merged, then #458 can be closed as won't fix, i.e. `scheme://path` is required if linking to any external link (incl. links outside the base).

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -49,7 +49,8 @@ export type { Header }
 
 export const createMarkdownRenderer = (
   srcDir: string,
-  options: MarkdownOptions = {}
+  options: MarkdownOptions = {},
+  base: string
 ): MarkdownRenderer => {
   const md = MarkdownIt({
     html: true,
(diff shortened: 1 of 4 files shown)


ST-DDT:
Could you change this to default to `''`?

https://github.com/vuejs/vitepress/commit/804954cf4d5417b1abcba9854ed5f064348292c5#diff-332834eb76757b8e31926e411cef34f7207564d0bf5fd00a185b41d97b3403bfR53

PR #511, opened 2022-01-29
docs: fix typo

ilkome:


brc-dd:
The original one is correct. The apostrophe in "Vuepress' little brother" is not a typo. It is interpreted like "little brother of Vuepress". When a word ends with "s", you (generally) don't write "foos's" and instead write "foos' ". Both mean the same.

PR #514, opened 2022-01-31
fix: normalize image source

### LINKED ISSUE #450: Assets resolution inconsistency in Dev and Build modes
### Describe the bug

When a markdown file has sibling assets files
```
- index.md
- image.png
```
and assets are referenced as
```
![An image](image.png)
```
then those assets
- are resolved just fine in DEV mode
- are NOT resolved during BUILD and BUILD fails with an error message like
  
image

Referencing assets using relative paths `An image` fixes the problem 

### Reproduction

Described above

### Expected behavior

Dev and Build modes should behave the same way.
If assets should be referenced using relative paths, then BOTH modes should fail at resolving absolutely referenced assets

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 126.73 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 95.1.31.88
    Chrome: 96.0.4664.55
    Firefox: 90.0.2
    Safari: 15.1
  npmPackages:
    vitepress: ^0.20.1 => 0.20.1
```

### Additional context

_No response_

### Validations

brc-dd:
fixes #450

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -11,6 +11,7 @@ import { hoistPlugin } from './plugins/hoist'
 import { preWrapperPlugin } from './plugins/preWrapper'
 import { linkPlugin } from './plugins/link'
 import { headingPlugin } from './plugins/headings'
+import { imagePlugin } from './plugins/image'
 import { Header } from '../shared'
 import anchor from 'markdown-it-anchor'
 import attrs from 'markdown-it-attrs'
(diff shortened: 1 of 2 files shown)


meteorlxy:
I'm afraid it's not an ideal fix. Two problems:

- The raw html `` tag is not handled
- It makes the `vite.resolve.alias` fail to work, because the plugin will prepend `./` to the alias.

meteorlxy:
IMO, the original behavior is consistent with vite + vue DX, so we might not need this plugin.

PR #523, opened 2022-02-08
docs: update links

brc-dd:


brc-dd:
PS: some changes were done by linter/prettier.

PR #532, opened 2022-02-14
fix(docs) - broken "custom theming" link target

JanKruft:
Currently the link target is broken. (see https://vitepress.vuejs.org/guide/api.html#global-components). I guess the correct link should be referring to custom theming . I also changed the link target to a relative one.


brc-dd:
Should be fixed when #523 gets merged, along with many other broken targets.

JanKruft:
@brc-dd awesome i ll leave it with you and close the PR

PR #533, opened 2022-02-14
docs: make users install `vue` too

### LINKED ISSUE #531: Your application tried to access vue, but it isn't declared in your dependencies; this makes the require call ambiguous and unsound
### Describe the bug

use yarn2 build docs

### Reproduction

yarn 

yarn add vitepress -D
vitepress serve docs

### Expected behavior

build success

### System Info

```shell
❯ node -v
v16.13.2
❯ npm -v
8.4.1
❯ yarn -v
3.1.1

"vitepress": "^0.22.1"
```

### Additional context

Xnip2022-02-14_14-47-47

### Validations

brc-dd:
fixes #531

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -17,15 +17,15 @@ $ mkdir vitepress-starter && cd vitepress-starter
 Then, initialize with your preferred package manager.
 
 ```bash
-$ yarn init -2
+$ yarn init
 ```
 
 ## Step. 2: Install VitePress
 
-Add VitePress and Vue as (dev-)dependencies for the project.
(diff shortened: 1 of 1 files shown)


brc-dd:
Updated the branch. Explicitly installing Vue is required in Yarn v2/v3.

kiaking:
Thanks! Should we change package manager to be `npm` 🤔  And then add `info` box or something to mention about `yarn 2/3`. We're using `pnpm` here but... it should work in npm without probrems too so 👀

brc-dd:
If we merge #707, then also we would need to add vue explicitly (in pnpm/npm too).

Issue #456, opened 2021-12-02
failed to start server. error:  Error [ERR_PACKAGE_PATH_NOT_EXPORTED]:


skyroam:
### Describe the bug

Hi, when I served the documentation site in the local server with the command 'yarn docs:dev`, a error happened.

```
[root@calm-spin-2 doc.xxx.org]# yarn docs:dev
yarn run v1.22.17
warning package.json: No license field
$ vitepress dev docs
vitepress v0.20.1
failed to start server. error:
 Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/vue.runtime.esm-bundler.js' is not defined by "exports" in /home/wwwroot/doc.xxx.org/node_modules/vue/package.json
    at applyExports (internal/modules/cjs/loader.js:490:9)
    at resolveExports (internal/modules/cjs/loader.js:506:23)
    at Function.Module._findPath (internal/modules/cjs/loader.js:634:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:952:27)
    at Function.resolve (internal/modules/cjs/helpers.js:78:19)
    at resolveAliases (/home/wwwroot/doc.xxx.org/node_modules/vitepress/dist/node/serve-3fdbda2c.js:12333:28)
    at resolveConfig (/home/wwwroot/doc.xxx.org/node_modules/vitepress/dist/node/serve-3fdbda2c.js:13374:12)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Object.createServer (/home/wwwroot/doc.xxx.org/node_modules/vitepress/dist/node/serve-3fdbda2c.js:36633:18) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[root@calm-spin-2 doc.wandersky.org]# node --version
v12.18.3
```
Why did this happen? Should I update the node?

### Reproduction

I install the vitepress according to [this tutorial](https://vitepress.vuejs.org/guide/getting-started.html) totally. Until this command: `yarn docs:dev`, the error occurred. 

### Expected behavior

installed succussfully

### System Info

```shell
centos7 

npx: installed 1 in 1.808s

  System:
    OS: Linux 4.10 CentOS Linux 7 (Core)
    CPU: (2) x64 QEMU Virtual CPU version (cpu64-rhel6)
[...]

PR #546, opened 2022-02-22
fix: require at least node v14

brc-dd:
https://github.com/vuejs/vitepress/issues/456#issuecomment-1047407711


victorlmneves:
It doesn't work with node 14.7.0
But I can say it works with 14.15.0

PR #564, opened 2022-03-03
docs(config): DocSearch upgrade to v3 configuration

godxiaoji:
The vitepress uses algolia [DocSearch v3](https://github.com/algolia/docsearch/). In this version, `appId` is now required.

See: [Migrating from DocSearch v2](https://docsearch.algolia.com/docs/migrating-from-v2).


godxiaoji:
There is already a better pr.

PR #569, opened 2022-03-05
docs: edge comment about details

mwoodpatrick:
Edge after the switch to using chromium engine (version 98) does support details block

### DIFF
--- a/docs/guide/markdown.md
+++ b/docs/guide/markdown.md
@@ -140,7 +140,7 @@ This is a dangerous warning
 :::
 
 ::: details
-This is a details block, which does not work in Internet Explorer or Edge.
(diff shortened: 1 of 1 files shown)


brc-dd:
Actually, it's supported since v79. [Ref.](https://caniuse.com/details)

docs/guide/markdown.md
```diff
@@ -140,7 +140,7 @@ This is a dangerous warning
 :::
 
 ::: details
-This is a details block, which does not work in Internet Explorer or Edge.
+This is a details block, which does not work in Internet Explorer or Edge (before version 98)
```
posva, line 143:
```suggestion
This is a details block, which does not work in Internet Explorer or old versions of Edge.
```

Issue #551, opened 2022-02-22
API to change head content


aboutsimon:
### Is your feature request related to a problem? Please describe.

Currently, there is no documented way to programmatically change or add `head` tags to the page at build time, that i'm aware of. I know that i can add `head` tags in the config or via `frontmatter`, but in some cases that is not sufficient. For example, when i want to add head tags based on existing meta data, like generating an `og:image` based on `frontmatter` ([gatsby-remark-twitter-cards](https://www.npmjs.com/package/gatsby-remark-twitter-cards) is an example in react land).

### Describe the solution you'd like

A clean API to programmatically change or add `head` tags, prior to SSG.

### Describe alternatives you've considered

Obviously i could dynamically add `head` tags to the page, but for some tags that is not a viable solution. For example, social media sites usually don't execute JS on the page, so any dynamically added Open Graph tags will not be considered.

Another option would be hacking something together around [rollup's transform hook](https://rollupjs.org/guide/en/#transform), but i would rather use a more dedicated vitepress API directly.

### Additional context

Maybe i'm missing something here and this is already possible - then please point me in the right direction.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

jimbo2150:
The social media tags were added to Frontmatter for the [VitePress website](https://raw.githubusercontent.com/vitejs/vite/v3.1.6/docs/blog/announcing-vite3.md), but they don't appear to be applying (either on the VitePress site or mine) and [don't appear to be documented](https://vitepress.vuejs.org/config/frontmatter-configs). Is that a bug?

brc-dd:
https://vitepress.dev/reference/site-config#transformhead

PR #588, opened 2022-03-25
feat(seo): head content is configurable per pages

georges-gomes:
## Background

This changes has been created to improve quality of meta tags in `head` for social links and SEO.

A number of tags are required in `<head>`.
- Meta for OG
- Meta for Twitter
- Canonical URL
...

Some of this content is page specific like title and description but also potentially images.

Currently, some of this information is filled by VitePress, some can be added to head via frontmatters
https://vitepress.vuejs.org/guide/frontmatter.html#head
But the first one has precedence to the other one and the notation is quickly very verbose.

## Solution here

The solution here is to support a function in the `head` configuration
`HeadConfig[] | (page: DataPage) => HeadConfig[]`

So every page is rendered with this function.

It let you have:
- `og:title` or `twitter:title` with the title of the page without the extra Title decoration
- `og:description` or `twitter:description`
- create a `canonical` for each page
- generate or link to meta images appropriatly
- add tags in frontmatters of the page and transform them into the right head tags
- dynamically change anything in the meta tags based on the page and page frontmatters

## Not supported

Page routing doesn't update head tags.


georges-gomes:
Example of use -> https://github.com/vuejs/vitepress/pull/588/files#diff-3def678deb1b1d5a53948eb8491817dde4dc881e032e785fc61cc93d334eefcd

georges-gomes:
fixes #551

georges-gomes:
Hi @kiaking 
Let me know if you like that.
I think this feature is better than adding individual tags support in frontmatters.
We use it in production already.
At your service

kiaking:
@georges-gomes Thanks for the PR and this is indeed cool feature. But let me think bit more deeper on this.

It's super flexible and good, but maybe most of these meta should be injected automatically. From that perspective, maybe there might be simpler way to customize for users...? I don't know, but I'll think about this a bit more!

kiaking:
I'll add #551 to 1.0.0 milestone so we don't forget.

georges-gomes:
@kiaking Great. Happy to revamp another implementation if you come across a better idea 👍

docs/.vitepress/config.ts
```diff
@@ -92,3 +98,34 @@ function getConfigSidebar() {
...
+  return [
+        // Twitter
+        ['meta', { name: 'twitter:card', content: 'summary_large_image' }],
+        ['meta', { name: 'twitter:site', content: '@vuejs' }],
+        ['meta', { name: 'twitter:title', content: pageData.title }],
+        ['meta', { name: 'twitter:description', content: pageData.description || description }],
+
+        // Open Graph
+        ['meta', { property: 'og:type', content: 'website' }],
+        ['meta', { property: 'og:locale', content: lang }],
```
xfq, line 116: `og:locale` uses [underscore](https://ogp.me/#optional) instead of the BCP 47 [hyphen](https://www.rfc-editor.org/rfc/rfc5646.html#page-4) in HTML [lang](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes) to separate the subtags, so language tags like `en-US` should be converted to `en_US`.

treystout:
@georges-gomes thank you for this PR, hoping it's merged in some form soon. Could you explain what you mean by "Page routing doesn't update head tags." Does this mean crawlers would see the right tags but developers would not?

brc-dd:
> Does this mean crawlers would see the right tags but developers would not?

The tags will be there in the generated HTML (yeah, crawlers would see the right tags) but it won't be updated when a user moves from one page to another.

treystout:
@brc-dd thank you for clarifying. Is the team still deciding if this PR is relevant? The only thing I can think of that would make it more expressive for end-users is ensuring @georges-gomes `getHead` has sufficient context. But adding arguments to it later would be easy and backward compatible. Anything I can do to help move either this PR forward or a another approach? Meta/Head issues are stopping my team from using vitepress in production, and it seems like something almost everyone would need once hosting in production. Thank you!

brc-dd:
> The only thing I can think of that would make it more expressive for end-users is ensuring getHead has sufficient context.

It still has sufficient context, you can access nearly everything from `pageData` itself.

> Is the team still deciding if this PR is relevant?

The PR is obviously relevant and much demanded TBH (needs rebase though). We're thinking of implementing some more compact API though, but even in its current form the PR is cool. We'll try to get this feature landed as soon as possible.

georges-gomes:
@brc-dd How do you feel about this PR? Should I rebase it or you are not sure about it?

georges-gomes:
> @georges-gomes thank you for this PR, hoping it's merged in some form soon. Could you explain what you mean by "Page routing doesn't update head tags." Does this mean crawlers would see the right tags but developers would not?

Sorry for the late reply, missed the notification.
Yes, the head tags are good at "arrival" and Google will see them.
But when a user route to another page the head tags are not updated with the values of the new page.
Doesn't affect Google but could be considered as "not good enough" for Vitepress.

We are using this as a patch in production with dynamic meta tags on https://backlight.dev/docs/ and https://backlight.dev/mastery/ if you want to see it in action.

brc-dd:
> How do you feel about this PR? Should I rebase it or you are not sure about it?

@georges-gomes It will need rebase, but not right now. Let's merge #1339 first. Likely there will be conflicts after that. But resolving them on this PR will be easier. Also, I wanted to ask, isn't [`transformHead`](https://vitepress.vuejs.org/config/app-configs#transformhead) behaves similar to what we have in this PR?

kiaking:
Closing due to inactivity. I think this is useful feature, though maybe we might wanna discuss in issues first to figure out the best API for it 👍

brc-dd:
People coming to this later, please try transformHead hook. IMO it does the same thing as this PR.

PR #597, opened 2022-04-04
ESM instead of CommonJs

ikmrgrv:
Modified the CommonJs style export of `config.js` to the ESM style export.

Vite is all about advancement of tools. Why not reflect it in documents, by using ESM instead of CommonJs.


WebMechanic:
I believe this is where the voyage goes like with the rest of the Vueniverse:
```js
// .vitepress/config.js
import {defineConfig} from 'vitepress'
export default defineConfig({ ... });
```

yyx990803:
We actually ship `defineConfig` and `defineConfigWithTheme` now, both with full type support, which we should document at some point.

Issue #593, opened 2022-04-01
export to pdf 


laoshaw:
### Is your feature request related to a problem? Please describe.

not really

### Describe the solution you'd like

a plugin or built-in feature of vitepress that I can export the document to PDF, to annotate offline

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

kiaking:
I think this is too heavy feature to be built in to the core. Having such an 3rd party lib might be really cool though 👍

posva:
You can also use a browser extension: https://chrome.google.com/webstore/detail/save-as-pdf/kpdjmbiefanbdgnkcikhllpmjnnllbbc

yadunandanbhat:
> I think this is too heavy feature to be built in to the core. Having such an 3rd party lib might be really cool though 👍

There is! See [markdown-pdf](https://github.com/alanshaw/markdown-pdf/), But it might be a hassle integrating it.

ourongxing:
try to use https://github.com/busiyiworld/web-printer/tree/main/packages/vitepress

condorheroblog:
you can use this https://github.com/condorheroblog/vitepress-export-pdf

Born specifically for `VitePress`, convert your desired web page into a PDF.

![](https://github.com/vuejs/vitepress/assets/47056890/d12e2ef4-2372-4854-9cdd-b61ceea5b5d7)

PR #612, opened 2022-04-25
feat: export page to pdf

kecrily:
Use [`window.print`](https://developer.mozilla.org/en-US/docs/Web/API/Window/print) and `@media print`. No external libraries were imported.

image

close https://github.com/vuejs/vitepress/issues/593


kiaking:
I think having css for Print style is nice! But adding button might be too much 🤔

kecrily:
Ok. I will create a PR that only includes print style

brc-dd:
Closing this in favor of #755

PR #615, opened 2022-04-26
feat: use faster git timestamp implementation if possible

Brooooooklyn:
Decrease the `docs-build-only` time from `1.6s` to `1.42s`.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -70,6 +70,10 @@ function getGuideSidebar() {
         {
           text: 'Differences from Vuepress',
           link: '/guide/differences-from-vuepress'
+        },
+        {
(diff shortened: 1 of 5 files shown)


yyx990803:
Tested on my M1 mac and the difference seems to be less noticeable (1.59s vs 1.55s)
In the meanwhile this pulls in an extra 2.4mb of install weight (also a dep that can't be pre-packaged)

Maybe it should be a plugin that can be separately installed 🤔

Brooooooklyn:
@yyx990803 moved into `peerDependencies` and added some documentation about how to use it.

kiaking:
Closing for now. I think the time difference is too small to add in extra complecity 👀

PR #619, opened 2022-05-06
docs: fix missed argument in snippet section

codecalm:
In docs should be:

```
<<< @/snippets/snippet-with-region.js#snippet{1}
```

instead of:

```
<<< @/snippets/snippet-with-region.js{1}
```


kiaking:
Looks like this is fixed in the latest doc 👍  Thanks for the PR!

PR #625, opened 2022-05-18
fix: remove `configureServer()` hack and use Vite 3 instead

brillout:
This PR removes a hack that was needed to inject the Vitepress middleware. After we merge https://github.com/vitejs/vite/pull/8217 we won't need that hack anymore.

This is a draft to show how to make Vitepress work with https://github.com/vitejs/vite/pull/8217.

We can merge this PR after merging https://github.com/vitejs/vite/pull/8217.


brillout:
This PR doesn't work for URLs ending with `.html`.

I'm closing this since the updated PR doesn't need any Vitepress changes anymore.

PR #639, opened 2022-05-22
fix(theme): header blur incomplete

kecrily:
The `content` class doesn't occupy 100% of the width, so the header's blur is not complete and appears abrupt.

### DIFF
--- a/src/client/theme-default/components/VPNav.vue
+++ b/src/client/theme-default/components/VPNav.vue
@@ -28,6 +28,7 @@ provide('close-screen', closeScreen)
 @media (min-width: 960px) {
   .VPNav {
     position: fixed;
+    backdrop-filter: saturate(50%) blur(8px);
   }
 }
 </style>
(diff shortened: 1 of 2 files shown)


kiaking:
Thanks for the PR! Yap we need this, but we need a bit more. When sidebar is open, this gives extra blur on top left 😓 

Screen Shot 2022-05-23 at 9 52 46

We need a way to crop this blue? Do you think you can tackle this issue? 👀

kiaking:
Maybe we could switch the element having blur by checking `has-sidebar`.

```ts
// in header
import { useSidebar } from 'vitepress'

const { hasSidebar } = useSidebar()
```

<template>
 <div :class="{ 'has-sidebar': hasSidebar }>
</template>

kecrily:
It should be fine now.

kiaking (approved):
Thanks a lot! I've styled the code format just a bit, but this is working like a charm 👍

PR #642, opened 2022-05-23
fix(locale): use site lang as lastUpdated locales

RainKolwa:
The last updated time string is hard coded as "en-US", it would be better if using site config `lang`.


kiaking:
Done in `main` branch 👍

PR #644, opened 2022-05-23
perf(a11y): add aria-checked attribute to switch

kecrily:
Use the dark switch provided by vueuse and add the aria-checked attribute to the switch


kiaking:
Thanks for the pr!

@yyx990803 Do you remember why we stop using Vueuse for Dark mode control? I remember initially, we're using Vueuse dark mode feature, and then we swtiched to creating the original one in Vue.js website.

Was it due to trying to reduce dependencies? If so, we're already using Vueuse for `useMediaQuery` so maybe we could reuse Vueuse for the dark theme as well? 🤔

kiaking:
@kecrily OK, let's not use vueuse for now. We can simply apply the aria label with current `useAppearance` function. Would you mind updating the code 👀

kiaking:
@kecrily OK, let's not use vueuse for now. We can simply apply the aria label with current `useAppearance` function. Would you mind updating the code 👀

kiaking (changes requested):
This will fail on Build 😢 Because `window` and `localStorage` will be undefined during the build time. We need to do a bit more complicated stuff here.

PR #654, opened 2022-05-25
feat: use shiki instead of prismjs for syntax highlight

### LINKED ISSUE #627: use `shiki` instead of `prism` as the default syntax higlighting engine
The same is already done in Vue.js website.

ryo-gk:
close #627

Use [shiki](https://github.com/shikijs/shiki) instead of prismjs for syntax highlighting.

The shiki's theme can be passes as `UserConfig.markdown.theme` like below. 
And if not be passed, then `material-palenight` is used as a default, which is the same as [vuejs/docs](https://github.com/vuejs/docs).

```js
export default defineConfig({
  lang: 'en-US',
  //...
  markdown: {
    theme: 'nord'
  },
  //...
})
```

Maybe something is missing because this is my first PR in Vitepress, please let me know then!


kiaking:
Thanks a bunch! I'll takeover to style things like line highlighting and documentations 👍

kiaking (approved):

Issue #473, opened 2021-12-17
Copy to clipboard button in code snippets


jd-solanki:
### Is your feature request related to a problem? Please describe.

It will be great to add copy to clipboard button in code snippets for longer code snippets where dragging requires window scroll

### Describe the solution you'd like

Add copy to clipboard button in default theme

### Describe alternatives you've considered

None

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/master/.github/contributing.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
Any proposal for the UI? Because currently the top-right of a code block shows language.

WebMechanic:
This is available as a [plugin for Prism](https://prismjs.com/download.html#themes=prism-funky&languages=markup+css+clike+javascript) (need to scroll down the page), but it's unclear to me how they could be added to the configuration inside VitePress. 

I'd also love to have this feature and potentially use other Prism plugins.

rogier-stegeman:
Would love this feature!

> Any proposal for the UI? Because currently the top-right of a code block shows language.

How about moving the language to the bottom right? See any page on the [Vuetify docs](https://vuetifyjs.com/en/features/accessibility/) for an example.
Copy button could also fade in on hover like it does on GitHub.

WebMechanic:
> How about moving the language to the bottom right? 

I put this in my custom theme stylesheet:
```css
div[class*="language-"]::before {
  top: initial; /* or unset */
  bottom: .6em;
}
```

kiaking:
This is cool idea and we should do it! Although, I think this can be done after 1.0.0 release as a feature addition, so probably I'll not be working on with this anytime soon.

If someone wants to jump in, that's super welcome!

WebMechanic:
I don't believe the single "copy" feature should be a built into Vitepress.
Rather allow for a simple integration of **other (official) Prism plugins** so ppl can add whatever they want and place it wherever they want in their code blocks. Maybe even make use of Prism's themes.
https://prismjs.com/plugins/

brc-dd:
@WebMechanic We won't be using prism anymore from the next release. #627

WebMechanic:
@brc-dd ok. Thanks for the heads up.

[...]

PR #658, opened 2022-05-25
feat: add copy code element (#473)

brc-dd:
fixes: #473

### DIFF
--- a/src/client/theme-default/components/VPDoc.vue
+++ b/src/client/theme-default/components/VPDoc.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { computed } from 'vue'
 import { useData } from 'vitepress'
+import { useCopyCode } from '../composables/copy-code'
 import { useSidebar } from '../composables/sidebar'
 import VPDocAside from './VPDocAside.vue'
 import VPDocFooter from './VPDocFooter.vue'
@@ -11,6 +12,8 @@ const { hasSidebar } = useSidebar()
 const pageName = computed(() => {
   return page.value.relativePath.slice(0, page.value.relativePath.indexOf('/'))
 })
+
+useCopyCode()
 </script>
 
 <template>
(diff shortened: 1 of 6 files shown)


src/client/theme-default/components/VPDoc.vue
```diff
@@ -11,6 +12,8 @@ const { hasSidebar } = useSidebar()
 const pageName = computed(() => {
   return page.value.relativePath.slice(0, page.value.relativePath.indexOf('/'))
 })
+
+useCopyCode()
```
kiaking, line 16: Maybe we should use this in `VPContent` instead? Because people can use `layout: page` and still have code block 👀

kiaking (commented):
Niiiiice! This looking slick man! Awesome 🎉 

I made one comment. After resolving that, I'll modify styles just a bit to make it fit my obsession, but definetly going to merge this one 👍

src/client/theme-default/components/VPDoc.vue
brc-dd: @kiaking All of our styles regarding code blocks need parent to have `vp-doc` class (all references to `[class*='language-']` are present only inside `vp-doc.css` file). When `layout: page` is set, we only have syntax highlighting, no other thing like language descriptor at the top or that curved-rectangular container is there.

But since `useCopyCode` just adds the event handlers not the styles, there isn't any issue in adding it to `VPContent` itself. Developers will need to add content to `span.copy` in order to interact with the element and copy in page layouts.

kiaking: Ahhh good point. I did miss out one that part 😅 

But thanks! I think your call of moving the code to `VPContent` still make sense, where the purpose of `layout: page` is to provide same markdown pasring feature, but without styles 👍

As you mentioned, users can define their own styles for this button inside page layout.

kiaking (approved):

PR #659, opened 2022-05-26
feat: migrate to ESM

antfu:
This shouldn't have much impact on the user side as Vitepress is mostly a CLI tool

## Changes

- `"type": "module"` and now `.js` are all ESM
- Use `rollup-plugin-dts` instead of api-extractor for simplity
- Use `picocolors`, `fast-glob` and `prompts` to align with vite


src/client/tsconfig.json
```diff
@@ -1,10 +1,12 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../../tsconfig.json",
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "../../dist/client",
     "target": "esnext",
     "module": "esnext",
+    "declaration": true,
+    "declarationDir": "../../dist/client-types",
```
kiaking, line 9: This should be `client` instead of `client-types`?

```suggestion
    "declarationDir": "../../dist/client",
```

kiaking: Wait not really, we need `client-types` in rollup config... However, this raises

```
[!] (plugin cleanup) Error: ENOENT: no such file or directory, stat '/Users/kia/Code/vitepress-next/dist/client-types'
```

When running `pnpm run docs`.

kiaking: OK this is happening because on `dev` command, client build executes, and then node build (rollup) executes in parallel and `client-types` folder gets deleted becaue of `cleanup` hook in rollup config. So we should first check for the folder existense and then perform delete on the folder I guess.

So maybe we just do this?

```ts
// rollup.config.ts
plugins: [
  dts(),
  {
    name: 'cleanup',
    async closeBundle() {
      try {
        await fs.rm(r('dist/client-types'), { recursive: true })
      } catch {}
    }
  }
]
```

kiaking:
@antfu Currently, because we only have es build, the consumer of VitePress have to define `"type": "module"`.

Maybe we need `cjs` build as well? Since if someone is building none type module lib and if they want to create docs with VitePress, are they doomed? 🤔

kiaking:
I think I'll add cjs support. Tested locally with Vite and it will work with/without `"type": "module"`.

kiaking:
Opened PR that targets this PR at #666.

kiaking (approved):

patak-dev:
This is awesome! Related https://github.com/vitejs/vite/pull/8348#issuecomment-1138724151. I think we should be able to test VitePress with the new ESM SSR build after this PR is merged.

kiaking:
Cool 👀 I'm resolving conflict now and will merge it soon 👍

Issue #521, opened 2022-02-07
Vitepress's own docs have hydration mismatch


JessicaSachs:
### Describe the bug

When working on fakerjs's Vitepress docs we noticed that all content entered after the last header on a page would be removed.

We opened up the console and saw there was a hydration mismatch. At first, we thought the issue was that we were using a `[[toc]]` element at the top of our document https://github.com/vuejs/vitepress/issues/388 but that was not the case.

After we removed the TOC, there was no change and all content after the last `### tag` header would still be removed when refreshing the page.

### Reproduction

Visit https://vitepress.vuejs.org/guide/markdown.html#advanced-configuration and see that all content underneath the "Advanced Configuration" header does not exist.

### Expected behavior

All content should render with no hydration mismatch.

### System Info

```shell
System:
    OS: macOS 11.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 393.57 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.0 - ~/.nvm/versions/node/v14.17.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.13 - ~/.nvm/versions/node/v14.17.0/bin/npm
  Browsers:
    Chrome Canary: 100.0.4868.0
    Firefox: 95.0
    Safari: 14.1.1
  npmPackages:
    vitepress: ^0.21.6 => 0.21.6
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
> After we removed the TOC, there was no change and all content after the last ### tag header would still be removed when refreshing the page.

Any reproduction for this particularly? For me, if I'm commenting out `[[toc]]` in the docs, the advanced configuration section is appearing properly in production.

I suspect that there might be some other component in your file that is causing hydration mismatch.

kiaking:
Maybe fixed by #521. It can't be reproduced now. I'll close this issue in a week or so if don't get any additional feedback 👍

kiaking:
Closing the issue since it seems like we don't have this problem now 👍

Issue #201, opened 2021-01-05
Direct link to page showing duplicated sections in the content


patak-dev:
Starting from https://vitepress.vuejs.org, you can navigate to the `Markdown Extensions` section and it properly works:

![image](https://user-images.githubusercontent.com/583075/103703486-e67cbf80-4fa7-11eb-96f7-74a2159490d1.png)

But reloading the page, or directly visiting https://vitepress.vuejs.org/guide/markdown.html is showing duplicated content:

![image](https://user-images.githubusercontent.com/583075/103703407-ccdb7800-4fa7-11eb-9f39-ca3c3dfc6c4c.png)

kiaking:
Oh yeah... hmmm... we must look into this. We're getting `Hydration completed but contains mismatches.` error in the console. I can only find this error in https://vitepress.vuejs.org/guide/markdown.html so markdown example should be the one causing it...? Need to to investigation.

sampullman:
I'm not sure if it's the exact same problem, but I get weird issues using `[[toc]]` where some sections are moved around or removed entirely. Similarly, it only happens when I refresh or visit directly, and only in the final build (not when served from `vitepress dev docs`

sampullman:
You can see this behavior in the Vitepress docs as well: https://vitepress.vuejs.org/guide/markdown.html#table-of-contents

When that page is loaded directly you get the "hydration mismatch" error (and the TOC is missing), but loading another page and navigating to the markdown section works fine.

stefandesu:
I have this problem as well. All pages that contain JavaScript show "Hydration completed but contains mismatches." when loaded directly. For example: https://gbv.github.io/jskos-vue/dev/components/ItemDetails.html

Interestingly, it works when navigating to the page from another site, but not if I simply go back and forth. Only the page that got initially loaded stays broken.

JeremyWuuuuu:
@patak-js Any idea what causes this? Having the same issue of duplicated sections and component lost.

seonglae:
` Hydration completed but contains mismatches.` error occurs when I deployed vitepress to vercel with custom dns

blurgyy:
Building with mpa mode (e.g. `yarn vitepress build --mpa`) seems to solve this issue, as well as mentioned in #388

seonglae:
In my case, mpa mode not working

https://intuiter.seongland.com/

vmihailenco:
`--mpa` helps but it breaks client code that uses JS...

PR #660, opened 2022-05-26
fix: switch to `markdown-it-toc-done-right`

### LINKED ISSUE #388: Adding [[toc]] generates div inside p element and causes a hydration mismatch
### Describe the bug

Adding `[[toc]]` generates `div` inside `p` element and causes a hydration mismatch

### Reproduction

```md

---
title: Testing
---

[[toc]]

## Test 1

test

## Test 2

test

```
this generates:
```html
...
<p>
<div class="table-of-contents"><ul><li><a href="#test-1">Test 1</a></li><li><a href="#test-2">Test 2</a></li></ul></div>
</p>
...
```
and a `div` element cannot be inside a `p` element so it causes a hydration mismatch when directly visiting the page

### Expected behavior

the `div` should not be inside the `p`

### System Info

```shell
System:
    OS: Linux 5.4 Alpine Linux
    CPU: (6) x64 Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
    Memory: 1.05 GB / 5.80 GB
    Container: Yes
    Shell: 1.32.1 - /bin/ash
  Binaries:
    Node: 16.8.0 - /usr/local/bin/node
    npm: 7.21.0 - /usr/local/bin/npm
  npmPackages:
    vitepress: ^0.16.1 => 0.16.1
```

### Additional context

_No response_

### Validations

brc-dd:
Fixes #388

Maybe #201 #521 too. Need investigation. Both are no longer reproducible.

### DIFF
--- a/docs/config/app-configs.md
+++ b/docs/config/app-configs.md
@@ -121,8 +121,8 @@ interface MarkdownOptions extends MarkdownIt.Options {
     allowedAttributes?: string[]
   }
 
-  // markdown-it-table-of-contents cplugin options
-  // https://github.com/Oktavilla/markdown-it-table-of-contents
+  // markdown-it-toc-done-right plugin options
+  // https://github.com/nagaozen/markdown-it-toc-done-right
   toc?: any
 
   // Configure the Markdown-it instance to fully customize
--- a/docs/guide/markdown-extensions.md
+++ b/docs/guide/markdown-extensions.md
@@ -403,8 +403,8 @@ module.exports = {
       permalink: anchor.permalink.headerLink()
     },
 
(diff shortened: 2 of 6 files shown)


brc-dd:
What should we keep the default list type in TOC? Ordered or unordered? Currently, VitePress has unordered. This PR changes that to ordered (not explicitly set, that's what toc-done-right has by default). We can change it back to unordered by passing `listType: 'ul'` to toc options.

kiaking:
I think it should be "unordered", cause I think that's more common in general 👍

kiaking (approved):
Awesome! 🚀

PR #663, opened 2022-05-26
feat: allow collapsible sidebar items

### LINKED ISSUE #640: [next] Add Collapsable feature to SidebarGroup

User experience is bad when I have too many SidebarGroups.  The Collapsable feature of SidebarGroup (similar to VuePress@Next) would be very necessary, which should be quite commonly required.

I understand VitePress might intend to provide a mininal default theme, rather than create another vuepress.  But still, I would strongly recommend you to add this feature.

_No response_

### Additional context

_No response_

### Validations

brc-dd:
fixes #640


kiaking:
Thanks a lot! As always, I'll adjust a bit on styles since I actually have the design for this toggle feature 😅 

But before doing that, I would like to discuss a bit on config option!

First, let's make `collapsible` -> `collapsable`, just because VuePress config is this way 👀 

Next, I would like to have defaults to open/closed option. So `collapsable: true` means sidebar group is collapsable. But I think it would be helpful to have option to set if the collapsable group is open by default, or closed by default.

So how about we add 2 options?

```ts
export interface SidebarGroup {
  text: string
  items: SidebarItem[]

  // If `false`, toggle button is hidden.
  // Default: true
  collapsible?: boolean

  // If `false`, collapsible group is collapsed by default
  // Default: true
  open?: boolean
}
```

What do you think?

brc-dd:
> First, let's make collapsible -> collapsable, just because VuePress config is this way 👀

Its `collapsible` in their docs: https://v2.vuepress.vuejs.org/reference/default-theme/config.html#sidebar

> So how about we add 2 options?

Fine.

kiaking:
Oh sorry for that. I was looking at [VuePress v1 doc](https://vuepress.vuejs.org/theme/default-theme-config.html#sidebar-groups) 😓  Let's go with `collapsible` 👍 

> Also, lets keep collapsible false by default. VuePress also has the same.

Gotcha. Sounds perfect 👍

brc-dd:
Changed the interface to this:

```ts
  export interface SidebarGroup {
    text: string
    items: SidebarItem[]
    // If `true`, toggle button is shown.
    // Default: false
    collapsible?: boolean
    // If `true`, collapsible group is collapsed by default.
    // Default: false
    collapsed?: boolean
  }
```

There is a certain issue with current implementation. When a group is collapsed by default, how do we get the items height for smooth transition? Currently, I am setting `collapsed` on mount, which is causing a flash on reload.

kiaking:
> Changed the interface to this:

Nice! I like `collapsed` much better 👍 

> There is a certain issue with current implementation. When a group is collapsed by default, how do we get the items height for smooth transition? Currently, I am setting collapsed on mount, which is causing a flash on reload.

I was thinking to not have smooth animation and just pop open/close 😅 I think it's not possible (at least in an easy way). So I would say for now go with no animation... until we find out some cool technique.

kiaking (approved):
I've swapped the icon and adjusted the styles.

I went with no style for now. I think we can animate it by using [FLIP](https://aerotwist.com/blog/flip-your-animations/) technique, but it's a bit of work and we can always add it later 👍

brc-dd:
@kiaking You removed `props.collapsible` checks too. Now, even if someone hasn't set `collapsible` (or has set false), one can just click the title to toggle.

kiaking:
Woops, will fix right away 👍

kiaking:
Ah... good point. Hmmm, yeah let's only add `role="button"` also when assigning collapsible too.

kiaking:
Fixed! 🎉  Thank you so much for your review! 🙏

brc-dd:
One last one 😅: These horizontal borders are not aligned when scrollbar is shown:

image

Setting this (inside `src\client\theme-default\components\VPSidebar.vue:116`),  should fix this:

```css
@media (min-width: 960px) {
  .group {
    width: calc(var(--vp-sidebar-width) - 64px);
  }
}
```

kiaking:
Good catch! Thanks 😳 I'll push the fix now 👍

hyp530:
When a user enters a specific page, the corresponding SidebarGroup should open **automatically** (with other SidebarGroups collapsed), and even scroll to the exact location of sidebar.  Otherwise, a user would definitely get lost in the sidebar, when there're a lot of pages.

Wondering if you could add this feature? @kiaking

Issue #637, opened 2022-05-21
[next] Hide title if logo is specified


sifferhans:
### Is your feature request related to a problem? Please describe.

I want to use a logo for my documentation site, but when I add a logo in the config file, the page title is pushed to the side of the logo. I am using the `next-theme`, which I know isn't finished :)

See the image below.

![image](https://user-images.githubusercontent.com/18753964/169671366-2dc6dfcd-a18f-4cac-abb4-a1abf0d089ae.png)

### Describe the solution you'd like

It would be nice to hide the title in the header if there is a logo specified - or at least have the option to hide it.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

kiaking:
We should definitely have this feature 👍

kiaking:
Done!

PR #677, opened 2022-05-28
feat: option to hide title in case of logo

kecrily:
close https://github.com/vuejs/vitepress/issues/637

### DIFF
--- a/types/default-theme.d.ts
+++ b/types/default-theme.d.ts
@@ -8,6 +8,11 @@ export namespace DefaultTheme {
     logo?: string
 
     /**
+     * Custom site title in navbar
+     */
+    siteTitle?: string | false
+
+    /**
      * The nav items.
      */
     nav?: NavItem[]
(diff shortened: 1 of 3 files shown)


types/default-theme.d.ts
```diff
@@ -7,6 +7,11 @@ export namespace DefaultTheme {
      */
     logo?: string
 
+    /**
+     * Hide title when logo is avaiable
+     */
+    hideTitle?: boolean
```
kiaking, line 13: How about we make this `siteTitle?: string | false` and disable when it's `false`. This way, we can let user customize even the text as well.

- When it's `string` - Use it as nav title.
- When it's `false` - Do not display title.
- When iT's `undefined` - use `config.title` as title.

kiaking (commented):
Thanks for the PR! Added a comment. PLease have a look 🙏

kiaking (approved):
Thanks! I've changed condition to not care about `logo` property. `siteTitle` will always act in a same way w/o logo.

PR #680, opened 2022-05-29
fix: menu link don't show active status

kecrily:


kiaking (approved):
Nice! Thank you for the fix 👍

Issue #653, opened 2022-05-25
[next] Scroll to anchor not working on vitepress dev


brc-dd:
### Reproduction

- Clone and checkout `next-theme`.
- Run  `pnpm i && pnpm build && pnpm docs-dev`.
- Head to `http://localhost:3000/guide/markdown-extensions.html#custom-containers` in a new tab.

### Expected behavior

Links having target anchor should scroll to corresponding header.

### Additional context

Works on production build or if the tab has already the site opened, i.e. doesn't work only on dev mode when opening a new tab.

---

OT: Why is there a `ci-docs` script in package.json? It builds the project twice. `docs-build` should be sufficient in CI envs too.

kiaking:
Confirmed, yeah it's not working 🤔 I think this would be low priority since it works on prod env, but would be really nice if we can fix it.

> OT: Why is there a ci-docs script in package.json? It builds the project twice. docs-build should be sufficient in CI envs too.

Thanks! I don't know too. Seems like we're not using it in our GitHub actions but maybe we're using it in external service...? Not sure. I'll check when we setup VitePress alpha website online and if we don't need it, we can delete it.

PR #688, opened 2022-05-30
fix(next): scroll to anchor (#653)

RainKolwa:


brc-dd:
Hi. Thanks for the PR. But can you explain what's being done here? i.e. what was wrong with `nextTick`?

kiaking:
So, what is not working here? Scroll to anchor is working, isn't it? 🤔 Could you discribe what issue you're facing at the moment?

RainKolwa:
This PR fixes #653 .

RainKolwa:
I did some test like:
```js
setTimeout(() => console.log(document.querySelector('#app').innerHTML));
nextTick(() => console.log(document.querySelector('#app').innerHTML));
```
and it turned out that the `nextTick` line just printed empty, whereas `setTimeout` works.

kiaking:
Hmmm, strange. I'm not sure if I like ``setTimeout being used here. Though I don't have any other way of doing it, but... At least is's working on Prod env, I would like to know bit more on why this issue is happening.

PR #702, opened 2022-06-02
feat: support custom content in home layout

antfu:
Since v1, it's not possible to have custom content/components on the home page:

```md
---
layout: home
---

All the following content will be ignored
```

which was quite useful in v0.x to provide additional info on the home page.

This PR brings back the ability, but not so sure if it's a ideal solution

### DIFF
--- a/src/client/theme-default/components/VPHome.vue
+++ b/src/client/theme-default/components/VPHome.vue
@@ -13,13 +13,7 @@ import VPHomeFeatures from './VPHomeFeatures.vue'
     <VPHomeFeatures />
     <slot name="home-features-after" />
 
-    <div class="content">
-      <div class="content-container">
-        <main class="main">
-          <Content class="vp-doc" />
-        </main>
-      </div>
-    </div>
+    <Content />
   </div>
 </template>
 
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPHome.vue
```diff
@@ -12,6 +19,14 @@ import VPHomeFeatures from './VPHomeFeatures.vue'
     <slot name="home-features-before" />
     <VPHomeFeatures />
     <slot name="home-features-after" />
+
+    <div class="content">
+      <div class="content-container">
+        <main class="main">
+          <Content class="vp-doc" :class="pageName" />
```
brc-dd, line 26: We probably don't need this `pageName` feature here. People usually don't have multiple home pages. This was added in docs so that people may customize css for specific page.

kiaking (approved):
Thanks a lot! I've simplified the DOM structure and removed `vp-docs` style for now. Since I think usually when you want to customize Homepage, default styles might be too much to have. And it's easier to opt-in by adding `vp-docs` style than opt-out from it so 👀

Issue #570, opened 2022-03-05
Font normalization makes user font size override in browser impossible


Volker-E:
### Describe the bug

Hey, 
Volker E. from Wikimedia Foundation here. We're gladly using VitePress in our new [Codex](https://github.com/wikimedia/design-codex/) [component demos](https://doc.wikimedia.org/codex/main/). 

Using absolute `px` value units instead of relative CSS value units hinders user setting overrides in browser for font size choice. 
That's one of the [most popular accessibility features](https://medium.com/@vamptvo/pixels-vs-ems-users-do-change-font-size-5cfb20831773).

Original bug is downstream in https://phabricator.wikimedia.org/T303100

### Reproduction

Update font size setting in your browser to a larger font size and browse several pages. With `px` the font size remains the same. 

### Expected behavior

A relative unit (`rem`) should be used instead.

### System Info

```shell
VitePress v0.21.6
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #703, opened 2022-06-02
perf(a11y): make font size follow user brower settings

kecrily:
<table><tr>

<td width="500px">

## Before:

image

</td>

<td width="500px">

## After:

image

</td></tr></table>

Basically, change the px from `font-size` `line-height` `width` `height` to rem. Then fixed some very obvious problems.
close https://github.com/vuejs/vitepress/issues/570

### DIFF
--- a/src/client/app/components/Debug.vue
+++ b/src/client/app/components/Debug.vue
@@ -27,11 +27,11 @@ watch(open, (value) => {
 .debug {
   box-sizing: border-box;
   position: fixed;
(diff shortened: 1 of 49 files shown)


kiaking:
Thanks! Though I'm planning to do this at my side being a bit more careful. Since I don't know if it's a good idea to change things like border-width 🤔 And if we were to change the `width` of the content with `rem`, I think media queries should also follow that.

PR #707, opened 2022-06-03
feat: support esm ssr build

antfu:
In Vite v3 we now default to emit ESM build for SSR. This PR enables the ability it uses ESM build when available and fallback to CJS build when it's in Vite v2.

So we don't need this workaround in Vite 3
https://github.com/vitejs/vite/blob/f8c92d1f4c1bddb6c29cfd4753079157c4d96922/docs/vite.config.ts#L4-L6


src/node/plugin.ts
```diff
@@ -215,6 +215,17 @@ export async function createVitePressPlugin(
             delete bundle[name]
           }
         }
+
+        // in Vite v3
+        // we generate a fake package.json to make the output `.js` as ESM
+        // @ts-ignore
+        if (config.ssr?.format === 'esm') {
```
antfu, line 222: Blocked by https://github.com/vitejs/vite/pull/8455

src/node/build/bundle.ts
```diff
@@ -54,7 +54,10 @@ export async function bundle(
     ),
     // @ts-ignore
     ssr: {
-      noExternal: ['vitepress']
+      noExternal: [
+        'vitepress',
+        '@docsearch/css'
```
brc-dd, line 59: Also need to put `@vueuse/core` and `body-scroll-lock` here...?

antfu:
close in favor of #856

PR #709, opened 2022-06-03
feat: provide build end hook

### LINKED ISSUE #706: provide build end hook

There are situations where information about the build process is required before vitepress finish, which cannot be done from a Vite plugin. For example, adding a PWA including the generated pages in the service worker manifest.

The current problem is that we have to include scripts after the vitepress build so that it loads the PWA plugin manually and regenerates the PWA on the distribution folder and it is a very repetitive arduous job.

Adding a `build-end` hook providing the Vite config and the routes generated by Vitepress.

For example, using `Astro` integrations, you can just add a `astro:build:done` hook, `Astro` just call this hook once SSG process finish and before finishing the build process (provides some info about the SSG result like routes and urls):

https://docs.astro.build/en/reference/integrations-reference/#quick-api-reference

I can make a PR for the build end hook.

_No response_

### Additional context

_No response_

### Validations

userquin:
should also be included on next ;)

closes #706


userquin:
It seems Vitepress is not resolving the root Vite config, only expose the inlined Vite configuration and not resolving it via Vite `resolveConfig`.

I need to do this to make `vite-plugin-pwa` work:
```ts
import { resolveConfig } from 'vite'
import type { VitePluginPWAAPI } from 'vite-plugin-pwa'
import { VitePWA } from 'vite-plugin-pwa'

const PWA = VitePWA({...})
...
export default defineConfig({
  lang: 'en-US',
  ...
  async buildEnd() {
      const viteConfig = await resolveConfig({
          plugins: [PWA],
        },
        'build',
        'production',
      )
      const pwaPlugin: VitePluginPWAAPI = viteConfig.plugins.find(i => i.name === 'vite-plugin-pwa')?.api
      const pwa = pwaPlugin && !pwaPlugin.disabled
      await optimizePages(pwa)
      if (pwa)
        await pwaPlugin.generateSW()
    },
})

```

PR #721, opened 2022-06-04
feat: setup devtools and remove debug component

meteorlxy:
image
image


src/client/app/index.ts
```diff
@@ -80,6 +74,11 @@ export function createApp() {
     })
   }
 
+  // setup devtools in dev mode
+  if (import.meta.env.DEV) {
+    setupDevtools(app, router, data)
```
antfu, line 79: I think this API is better to be dynamically imported for better code-splitting and tree-shaking in production.

meteorlxy: I've tested and the tree-shaking seems to work properly.

antfu: But the side-effect from `./devtools` will remain (if any). Even if there aren't any currently, it's still better to split it into another chunk to prevent leaking side effects from future changes.

kiaking: OK, so adding `__VUE_PROD_DEVTOOLS__` will also strip off `import` statement from the production build as well...? 👀  If so, then we're ready to merge this sweet PR right?

meteorlxy: > OK, so adding `__VUE_PROD_DEVTOOLS__` will also strip off `import` statement from the production build as well...? 👀 If so, then we're ready to merge this sweet PR right?

Oh, it's another thing,  `__VUE_PROD_DEVTOOLS__` is used by vue core, and pre-defined by vite plugin-vue.

In fact, I think the `import.meta.env.DEV` has already stripped it from prod build. I'm not quite sure about what's "the side-effect from `./devtools`" said by @antfu  😢

meteorlxy: In fact, other tools like vue-router does not use dynamic import, neither (although it is packaged by rollup):

https://github.com/vuejs/router/blob/047851299c4c438cc977f6d5d7da455eb69195db/src/router.ts#L1233-L1235

kiaking: Side effect means that even the `if` condition is stripped away if `import { setupDevtools } from './devtools'` is kept, then anything that fire within that `devtools.ts` file will still be included in the build.

So if `devtools.ts` contains some side effect like loading custom css, then it will be loaded even we don't call `setupDevtools` function 👀 

We could check this by simply adding `console.log('Something')` in `devtools.ts` and see  if it gets logged. If yes, then it means the file is still imported I think.

meteorlxy: Yes, I know what is side-effect 👍 , but I'm not quite sure why this file might introduce any side-effects. I had checked the dist file by searching the keyword `org.vuejs.vitepress` (I think it could identify the file, just like your console.log way 😃 ) and confirmed it has been entirely stripped.

antfu: > Stripping off the import statement should be the default behavior of rollup/vite in most cases (I think so).

Only when the module does NOT contain side-effects. You might want to 

```ts
import 'some-polyfill'
```

This won't be stripped away even tho it imports nothing.

> I had checked the dist file by searching the keyword org.vuejs.vitepress

Yes, the `setupDevtools` function is side-effect-free so it got stripped. But it doesn't necessarily mean the entire `devtools.ts` has been stripped. I have just checked, if you put a top-level console.log in `devtools.ts` it will remain in the production bundle. **I am talking about defensive programming**, even if it's side-effect free at this moment, it won't prevent us to leak the side-effect in future iterations unintentionally. - Or do you see any reason we can't use dynamic import for that module?

meteorlxy: I got your point. As I understand it, defensive programming is to defend unpredictable user input and runtime environment, but not to defend developer commits - code review does it. I think it's unnecessary to try to defend against problems that do not exist now (they are endless).
Of course I could change it to dynamic import if you prefer it 😃

antfu (approved):

PR #723, opened 2022-06-05
perf(style): beautify scrollbar

alex8088:
screenshot

### DIFF
--- a/src/client/theme-default/components/VPSidebar.vue
+++ b/src/client/theme-default/components/VPSidebar.vue
@@ -120,5 +120,25 @@ watchPostEffect(async () => {
   .group + .group {
     margin-top: 24px;
   }
+
+  .VPSidebar:hover::-webkit-scrollbar {
+    display: block;
+  }
+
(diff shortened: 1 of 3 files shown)


kiaking:
Hmmmm I don't know, I don't know man. I'm kinda thinking we should keep default styles for this kind of stuff 🤔? It does look good though.

@alex8088 Is that screen shot taken on Windows?

alex8088:
> 



> Hmmmm I don't know, I don't know man. I'm kinda thinking we should keep default styles for this kind of stuff 🤔? It does look good though.
> 
> @alex8088 Is that screen shot taken on Windows?

No, on Ubuntu, a bit blurry. I think it will make the overall visual effect better.

meteorlxy:
As I remember, it will lower the score of a11y (I personally like the style, though)

kiaking:
K let's close this one for now. Personally, I like the style, and thank you for the PR! But I'm still not sure if really want to interact with this kind of thing in terms of a11y and such.

But this might be a cool trick to have as a custom style recipe or something like that 👍

PR #724, opened 2022-06-05
feat: don't add .html to urls

brc-dd:
Taking over #488.

The router will no longer force `.html` in urls. There is no change in the generated directory structure.


kiaking:
We need to have a bit more care on this one.

Having `.html` prefixed make the resulting html work without JS. Some services can handle the URL without `.html` like Netlify, but not all of them. If we want to have a pretty URL, we should build a directory that converts `page.md` into `page/index.html`. This means we might have to also modify the `manifest.json`.

Also ideally, we would want it to work with/without a trailing slash, (`/page` and `/page/`) but I am not 100% sure what is the standard here 🤔 

Seems like in Netlify, it redirects `/page` to `/page/`. [Netlify Docs](https://docs.netlify.com/routing/redirects/redirect-options/#trailing-slash)

georges-gomes:
My change in #488 generate the output with `page/index.html`  and rename the href accordingly so it also works in `mpa` mode.

I didn't want to support it on the router only for the warnings reported by @kiaking .

I was planning to port #488 to `next` branch in the next few days.

PR #743, opened 2022-06-07
fix(next): use base url on title

stafyniaksacha:
Use `site.base` instead of `/` prefix in `VPNavBarTitle`


brc-dd:
Hi! This was fixed in #718. PS: `next-theme` is no longer active development branch. After the `alpha` release, we are back to `main`.

PR #744, opened 2022-06-07
fix(styles): show close icon when nav menu is open

brc-dd:


brc-dd:
This also fixes an issue when hovering anywhere on the navbar was triggering the menu icon to change.

brc-dd:
And, also fixes the one where hovering over button was not causing the icon to change.

kiaking (approved):

PR #745, opened 2022-06-07
feat: support themable images for logo and hero

antfu:
This allows the logo and hero icon to be provided with light and dark versions.

#### config.ts

```ts
export default {
  theme: {
    logo: {
      light: '/logo-light.svg',
      dark: '/logo-dark.svg'
    }
  }
}
```

#### index.html

```yaml
---
layout: home
hero:
  image:
    light: /logo-light.svg
    dark: /logo-dark.svg
---
```

PR #757, opened 2022-06-08
fix(next): use withbase for images in hero and navbar

stafyniaksacha:
Append base path to user provided images path (home and nav logo)


brc-dd:
Use relative links for now (`./foo.png`), we'll fix this in #745 itself. Thanks for pointing this out!

PR #767, opened 2022-06-09
fix: use official version of Inter fonts

### LINKED ISSUE #759: Proper italic/cursive rendering
### Describe the bug

As of now, if you format text as italic/cursive (with either `_` or `*`) it gets transpiled to HTML as an `<em>` element, instead of `<i>`. This works in chromium-based browser (as far as I've seen), but for example doesn't show as italic in the latest Firefox versions.

### Reproduction

In a Markdown file, wrap any text in a single `_` or `*` (make it cursive), compile and search that text in the page with Firefox.

### Expected behavior

Render it to an `<i>` tag so it is properly formatted in both Chromium-based and Gecko-based (Firefox & others). Alternatively, a CSS could be included by default that formats `<em>` as italic/cursive.

### System Info

```shell
System:
  OS: Linux 5.15 Pop!_OS 21.04
  CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  Memory: 5.75 GB / 15.48 GB
  Container: Yes
  Shell: 5.1.4 - /bin/bash
Binaries:
  Node: 16.4.1 - ~/.nvm/versions/node/v16.4.1/bin/node
  npm: 8.9.0 - ~/.nvm/versions/node/v16.4.1/bin/npm
Browsers:
  Chromium: 102.0.5005.61
  Firefox: 96.0
npmPackages:
  vitepress: ^1.0.0-alpha.1 => 1.0.0-alpha.1
```

### Additional context

_No response_

### Validations

brc-dd:
fixes #759

On latest version of Chrome/Firefox, only `Inter.var.woff2` will be loaded. On slightly older ones (2-3 years older versions), `Inter-roman.var.woff2` will be loaded, `Inter-italic.var.woff2` will get loaded only if there are any italic text. On much older ones (those which do not support variable fonts) normal woff2/woff files will be loaded.


brc-dd:
Doesn't appear to be working on Safari.

Issue #765, opened 2022-06-09
Markdown formatting in custom container titles


noClaps:
### Is your feature request related to a problem? Please describe.

N/A

### Describe the solution you'd like

Allow custom titles of containers to be manipulated with Markdown, such as **bold**, _italics_, or hyperlinks.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
Hi! Can you explain more on what are you trying to do here?

noClaps:
I'm trying to do something like:

```
::: info [Lorem ipsum](https://example.org)
This is some example text
```

and have the title of the container be a hyperlink like [Lorem Ipsum](https://example.org). This is especially useful for quotes as I can link it back to the source.

PR #768, opened 2022-06-09
fix: allow title in containers to have markdown content

brc-dd:
fixes: #765

### DIFF
--- a/src/node/markdown/plugins/containers.ts
+++ b/src/node/markdown/plugins/containers.ts
@@ -1,4 +1,5 @@
 import MarkdownIt from 'markdown-it'
+import { RenderRule } from 'markdown-it/lib/renderer'
 import Token from 'markdown-it/lib/token'
 import container from 'markdown-it-container'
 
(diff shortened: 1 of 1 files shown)


src/node/markdown/plugins/containers.ts
```diff
@@ -15,31 +16,23 @@ export const containerPlugin = (md: MarkdownIt) => {
...
+type ContainerArgs = [typeof container, string, { render: RenderRule }]
 
 function createContainer(klass: string, defaultTitle: string): ContainerArgs {
   return [
     container,
     klass,
     {
-      render(tokens, idx) {
+      render(tokens, idx, options, env) {
+        const md = new MarkdownIt().set(options)
```
meteorlxy, line 27: I think we can passing the `md` instance into the `createContainer` function, instead of creating a new one here. It will create a new instance for every container.

---

Oh I remember it. The fifth arg of the render function is the renderer. You can also consider using it instead

```ts
render(tokens, idx, options, env, slf) {
  slf.renderInline()
}
```

meteorlxy (changes requested):

src/node/markdown/plugins/containers.ts
brc-dd: It needs a token array to be passed, instead of text I guess.

PR #774, opened 2022-06-11
fix: switch to h2 for feature headers

Demivan:
Theoretically Google is smart enough to work with multiple h1 headers. But using h2 for hero features, I think, makes more sense.


brc-dd:
Thanks!

PR #777, opened 2022-06-11
fix: add italic fonts

brc-dd:
fixes: #759


HipyCas:
When will this come out as a release on NPM?

kecrily:
Is it really necessary to have italic font built in? Why not fallback to other fonts.

kecrily:
Is it really necessary to have italic font built in? Why not fallback to other fonts.

brc-dd:
@kecrily Falling back doesn't appear to be working here. We had `"Inter, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif"` as font family. Setting `font-style: italic` on an element was not using any fallback font and the text was rendering with normal font style.

HipyCas:
When will this be merged???

kiaking (approved):

PR #778, opened 2022-06-11
refactor: improve page data parsing

meteorlxy:
Put `JSON.parse()` in the codegen, avoiding calling `JSON.parse()` everywhere.

### DIFF
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -169,9 +169,9 @@ const defaultExportRE = /((?:^|\n|;)\s*)export(\s*)default/
 const namedDefaultExportRE = /((?:^|\n|;)\s*)export(.+)as(\s*)default/
 
 function genPageDataCode(tags: string[], data: PageData) {
-  const code = `\nexport const __pageData = ${JSON.stringify(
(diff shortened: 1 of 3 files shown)


src/node/markdownToVue.ts
```diff
@@ -169,9 +169,9 @@ const defaultExportRE = /((?:^|\n|;)\s*)export(\s*)default/
 const namedDefaultExportRE = /((?:^|\n|;)\s*)export(.+)as(\s*)default/
 
 function genPageDataCode(tags: string[], data: PageData) {
-  const code = `\nexport const __pageData = ${JSON.stringify(
+  const code = `\nexport const __pageData = JSON.parse(${JSON.stringify(
```
brc-dd, line 172: Can't this be simplified to:

```ts
  const code = `\nexport const __pageData = ${JSON.stringify(data)}`
```

meteorlxy: JSON parsing is faster than object literals, and that's way we are using `JSON.stringify(JSON.stringify())` before 😆

meteorlxy: FYI: https://github.com/GoogleChromeLabs/json-parse-benchmark

PR #779, opened 2022-06-11
refactor: handle page data hmr in the same file

meteorlxy:
**Before**

We are loading page data in `router.ts`, but handling page data HMR in `index.ts`, which is bad for maintainability and readability.

**After**

Moving the page data HMR code into `router.ts` as well.


kiaking:
@meteorlxy Sorry seems like there have been conflict. Would you mind fixing it 🙏

meteorlxy:
@kiaking Done 👌

kiaking:
Thanks a lot!

PR #799, opened 2022-06-16
docs(migration): Added image migration guide for vuepress

jd-solanki:


kiaking (approved):
very precise, easy to follow, good documentation! Awesome!

meteorlxy:
Well, I wonder if the `withBase` is still necessary. 🤔

kiaking:
Well 👀 

> For dynamic images you still need `withBase` as shown in [Base URL guide](/guide/asset-handling.html#base-url).

meteorlxy:
Yeah I've read that. I just found that it's Vite who handle the base, but not VitePress. So I wonder if the base is not necessary. After some tests I know the reason, but I found a issue of Vite... I'm not sure if it has already been reported 🤣

kiaking:
Hah. Well it could be super cool if we can removed it completely though 👀

kecrily:
I think the second failure was in line with expectations. After you set base, everything you do should be done under base by default.

meteorlxy:
> I think the second failure was in line with expectations. After you set base, everything you do should be done under base by default.

Yes, but the behaviors of dev and build differs, so I thought it should be considered as a bug

meteorlxy:
In fact, there might be another edge case.

- The base of foo site is set to `/foo/`
- We have another bar site deployed under `/bar/` in the same domain.
- Then how to reference `/bar/logo.png` in foo site without add the domain name? I thinks the only way is to use dynamic path for now 🤔

meteorlxy:
> Is the second one available in dev? Its availability should be a bug

Yes it is.

kecrily:
> * Then how to reference `/bar/logo.png` in foo site without add the domain name?

Maybe it can be like this `../bar/logo.png`

Issue #809, opened 2022-06-18
when using `Multiple Sidebars` the sidebar not being updated


userquin:
### Describe the bug

just read the title

### Reproduction

https://github.com/antfu/vite-plugin-pwa/tree/userquin/feat-new-theme

you can also disable pwa, just add `disable: true` option on `VitePWA` plugin here: https://github.com/antfu/vite-plugin-pwa/blob/userquin/feat-new-theme/docs/vite.config.ts#L53

from root folder run `pnpm install && pnpm run build`  then `pnpm run docs` to start dev or `pnpm run docs:preview`  to start preview

### Expected behavior

The sidebar should update

### System Info

```shell
Also failing on Netlify preview: https://github.com/antfu/vite-plugin-pwa/pull/304


System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 42.87 GB / 63.95 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 102.0.5005.115
    Edge: Spartan (44.19041.1266.0), Chromium (102.0.1245.41)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    vitepress: ^1.0.0-alpha.2 => 1.0.0-alpha.2
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
Fix: `collapsed` should be a computed ref:

https://github.com/vuejs/vitepress/blob/107724ac6f24e5272964d3bdbff54169fa4d91ae/src/client/theme-default/components/VPSidebarGroup.vue#L15-L21

But what will be the `toggle` then?

userquin:
add a watcher or just add a setter to the computed: https://vuejs.org/guide/essentials/computed.html#writable-computed

userquin:
try this:
```ts
watchEffect(() => {
  if (props.collapsible)
    collapsed.value = props.collapsed
})
```

PR #811, opened 2022-06-18
fix: watch for change in props

brc-dd:
fixes: #809


userquin:
@brc-dd keep initilize the ref, the `watchEffect`  will not run first time

brc-dd:
@userquin I guess you are looking at a different documentation. See this: https://vuejs.org/guide/essentials/watchers.html#watcheffect

> `watchEffect()` allows us to perform a side effect immediately while automatically tracking the effect's reactive dependencies.

userquin:
> @userquin I guess you are looking at a different documentation. See this: https://vuejs.org/guide/essentials/watchers.html#watcheffect

you0re right, I was looking here: https://vuejs.org/api/reactivity-core.html#watcheffect

PR #831, opened 2022-06-21
chore: handle undefined hmr

sapphi-red:
Actually this PR does nothing when Vite is v2.

After Vite v3, `import.meta.hot.accept` is more accurately typed. (https://github.com/vitejs/vite/pull/8698/files)
This PR fixes the type error when Vite is upgraded to v3 which will fix vite-ecosystem-ci.


kiaking (approved):
Thanks! 👍 👍 👍

PR #836, opened 2022-06-22
fix: compatible with `--base` flag

### LINKED ISSUE #835: `vitepress build docs --base /xxx/` does not work as expected
### Describe the bug

It is passed as `build.base` instead of root level `base` option

```log
"build.base" option is deprecated. "base" is now a root-level config option.
```

### Reproduction

N/A

### Expected behavior

No warning and work

### System Info

```shell
System:
    OS: macOS 13.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.91 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/Library/Caches/fnm_multishells/985_1655820447617/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/985_1655820447617/bin/yarn
    npm: 8.12.1 - ~/Library/Caches/fnm_multishells/985_1655820447617/bin/npm
  Browsers:
    Chrome: 102.0.5005.115
    Firefox: 101.0.1
    Safari: 16.0
  npmPackages:
    vitepress: ^1.0.0-alpha.2 => 1.0.0-alpha.2
```

### Additional context

Working on a fix.

### Validations

JounQin:
close #835

### DIFF
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -9,14 +9,14 @@ import { bundle, okMark, failMark } from './bundle'
 
 export async function build(
   root: string,
-  buildOptions: BuildOptions & { base?: string; mpa?: string } = {}
+  buildOptions: BuildOptions & { mpa?: string } = {}
 ) {
   const start = Date.now()
 
   process.env.NODE_ENV = 'production'
   const siteConfig = await resolveConfig(root, 'build', 'production')
 
-  if (buildOptions.base != null) {
+  if (buildOptions.base) {
     siteConfig.site.base = buildOptions.base
     delete buildOptions.base
   }
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -9,7 +9,7 @@ import { bundle, okMark, failMark } from './bundle'
 
 export async function build(
   root: string,
-  buildOptions: BuildOptions & { mpa?: string } = {}
+  buildOptions: BuildOptions & { base?:string; mpa?: string } = {}
 ) {
   const start = Date.now()
 
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -9,12 +9,17 @@ import { bundle, okMark, failMark } from './bundle'
 
 export async function build(
   root: string,
-  buildOptions: BuildOptions & { mpa?: string } = {}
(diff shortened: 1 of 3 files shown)


brc-dd:
`vitepress serve` will break in this case. It requires the base to be in site config itself.

JounQin:
> `vitepress serve` will break in this case. It requires the base to be in site config itself.

@brc-dd 

So we should support `vitepress serve --base /xxx/`?

brc-dd:
We can, but I don't feel any need to introduce support for this. We'll need to change this:

https://github.com/vuejs/vitepress/blob/baf8083e668ce50a9d26d9398609e196237ebc95/src/node/serve/serve.ts#L26

to this:

```ts
  const base = trimChar(options.base ?? site?.site?.base ?? '', '/')
```

Then:

```sh
pnpm docs-build-only -- --base /foo/
pnpm docs-serve -- --base /foo/
```

will work.

JounQin:
@brc-dd Done!

src/node/build/build.ts
```diff
@@ -9,14 +9,14 @@ import { bundle, okMark, failMark } from './bundle'
 
 export async function build(
   root: string,
-  buildOptions: BuildOptions & { base?: string; mpa?: string } = {}
```
JounQin: There is no `base` available in `BuildOptions` because it is deprecated.

brc-dd:
Should we support this in dev too?

JounQin:
> Should we support this in dev too?

It seems vite does support this, but this is uncommon usage for myself.

src/node/server.ts
```diff
@@ -10,7 +10,7 @@ export async function createServer(
 
   return createViteServer({
     root: config.srcDir,
-    base: config.site.base,
+    base: serverOptions.base ?? config.site.base,
```
brc-dd, line 13: Yeah, this won't work, because the rest of VitePress needs it to be in site config itself. This option is to pass that to Vite only.

JounQin: Changed as your suggestion.

brc-dd (approved):

PR #837, opened 2022-06-22
fix(theme): don't show footer without sidebar on doc layout

FuZhouJohn:
When I don't use sidebar, the doc layout will show footer.

### DIFF
--- a/src/client/theme-default/components/VPFooter.vue
+++ b/src/client/theme-default/components/VPFooter.vue
@@ -1,13 +1,15 @@
 <script setup lang="ts">
 import { useData } from 'vitepress'
+import { computed } from 'vue'
 import { useSidebar } from '../composables/sidebar'
 
-const { theme } = useData()
+const { theme, frontmatter } = useData()
 const { hasSidebar } = useSidebar()
+const showFooter = computed(() => theme.value.footer && (frontmatter.value.layout === 'home' || frontmatter.value.layout === 'page'))
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


kiaking:
I like how it look with footer for example in this page.
https://main.vitejs.dev/plugins/

So, I like to keep footer in sidebarless doc page 🤔 

I haven't check but is this misunderstanding comes from some where in the docs?

FuZhouJohn:
> 

https://vitepress.vuejs.org/guide/theme-footer.html
> Note that footer will not be displayed when the page layout is set to doc.

FuZhouJohn:
> 我喜欢它的外观与页脚，例如在这个页面。 https://main.vitejs.dev/plugins/
> 
> 所以，我喜欢在无侧边栏文档页面中保留页脚🤔
> 
> 我没有检查，但这种误解是否来自文档中的某些地方？

@kiaking
 
[https://vitepress.vuejs.org/guide/theme-footer.html](https://vitepress.vuejs.org/guide/theme-footer.html)

> Note that footer will not be displayed when the page layout is set to doc.

FuZhouJohn:
> I like how it look with footer for example in this page. https://main.vitejs.dev/plugins/
> 
> So, I like to keep footer in sidebarless doc page 🤔
> 
> I haven't check but is this misunderstanding comes from some where in the docs?

@kiaking 

[https://vitepress.vuejs.org/guide/theme-footer.html](https://vitepress.vuejs.org/guide/theme-footer.html)

> Note that footer will not be displayed when the page layout is set to doc.

kiaking:
Oh I see, we should change that docs instead saying that it gets hidden when there is sidebar 😅

PR #838, opened 2022-06-22
fix: markdown plugin 内部调用render时__data数据被清空问题

Estelle00:


meteorlxy:
That's a reasonable fix, but we are planing to change the non-standard usage of markdown-it. Thanks anyway.

Issue #638, opened 2022-05-22
Support multi-line indented vue codes in markdown files


hyp530:
### Is your feature request related to a problem? Please describe.

This works in Vuepress@Next, but doesn't work in VitePress (treated as an **Indented Code Block**):

```html
<a-collapse>
    <a-collapse-item>
        <div> Content </div>
    </a-collapse-item>
</a-collapse>
```

This works in the current version of VitePress :

```html
<a-collapse>
<a-collapse-item>
<div> Content </div>
</a-collapse-item>
</a-collapse>
```

### Describe the solution you'd like

Suggests to support: (i) multi-line indented vue codes in vitepress, (ii) provide an option in confit.ts to disable **Indented Code Blocks**, which would trigger lots of troubles in a markdown file. 

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

kiaking:
Seems like we can't have 4 indent. So this also works.

```html
<a-collapse>
  <a-collapse-item>
  <div> Content </div>
  </a-collapse-item>
</a-collapse>
```

When ever we have more than 4 spaces, it gets treated as code block by markdown parser 👀  Not sure how to fix it but we should.

brc-dd:
Is this present on the next-theme branch? I am not able to reproduce this:

![image](https://user-images.githubusercontent.com/40380293/170556481-6e9ee897-2410-42b8-add4-da385d36a65f.png)

brc-dd:
These changes need to be made at `src\node\markdown\plugins\component.ts` in order to fix this.

https://github.com/vuepress/vuepress-next/blob/f91651c3c7248fdabfa19cb9bdcdba4c28622eec/packages/@vuepress/markdown/src/plugins/customComponentPlugin/htmlBlockRule.ts#L17-L30

kiaking:
Well done! Thank you so much @brc-dd

meteorlxy:
I propose to extract the markdown package / plugins to a standalone project, so that we don't need to fork the code like this 🤣 

I'd say I've fixed lots of markdown rendering issues like this one when development vuepress-next, and add lots of unit tests for those fixes.

Issue #607, opened 2022-04-22
My solution for custom components accessing relative assets  -- Vite's static asset handling and enhanced Markdown hoisting


kaceo:
### Is your feature request related to a problem? Please describe.

Currently Markdown is first-class in Vitepress in terms of asset-handling.  Vue, frontmatter are second class.

E.g. create a markdown page:

~~~md
<!-- index.md -->
This is an image
![](./image/myimage.png)
~~~

Vitepress parses into a Vue component, plus the image file is also added to the bundle pipeline for asset handling.

Compare with a user-defined Vue component:

~~~vue
//MyImage.vue
<template>
<img :src="withPath(src)">
</template>

<script setup>
import { withPath } from 'vitepress'
const props = defineProps({ src: String })
</script>
~~~

~~~md
<!-- index.md -->
This is an image defined as Vue component.
<script setup>
import MyImage from './components/MyImage.vue'
</script>

<MyImage src="./image/myimage.png" />
~~~

The second case, Vitepress generates a target filename that is original source path with Base prepended.  **But the filename is not the same as the Markdown-generated filename**.  Also the image file is not added to the build.

Same problem also happen to images referenced in frontmatter are also unprocessed by asset pipelines, e.g. HeroImage on the Home page.

New users of Vitepress are always confused about this --- e.g. see #505 and #301

### Describe the solution you'd like

Of course not every property or frontmatter string are assets, only some of them.  

Hence the user should have a choice of which strings should be hashed and added to the asset pipeline.

Vitepress already provides a function "withBase" for customising URLs, but this is not enough for image assets. 

I would like to see user-callable functions that:
- resolves the URL correctly in both Dev and Build cases
- adds the file to the hash/build asset pipeline
- and behaves correctly with the rest of Vite and Rollup 

This will allow custom components to behave identically with Markdown, which is the original promise of Vitepress (i.e. "every Markdown page is translated into a Vue SFC component")

### Describe alternatives you've considered

1) The usual fallback of putting all images into "public" directory.  Not an elegant solution.

2) Defining path prefix constants in themeConfig.  Does not resolve all user cases.

3) Extending the Vite config to process extra directories.  Two moving targets instead of one.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
[...]

PR #843, opened 2022-06-22
refactor: externalize markdown component plugin

meteorlxy:
As I commented here https://github.com/vuejs/vitepress/issues/638#issuecomment-1156004999

This PR externalizes the markdown-it component plugin.

The `@mdit-vue/plugin-component` package is extracted from vuepress-next, with full tests and some minor bug fixes (like #638).

Another fix is that allowing `@click` on native HTML tags, e.g. `<button @click="onClick">`, which will fail in current vitepress.


brc-dd:
Based on the usage part here: https://github.com/mdit-vue/mdit-vue/tree/main/packages/plugin-component#usage, do we need to add `{ html: true }` here in VitePress by default?

kiaking:
This is a cool idea and I like the direction. But, if we're going to externalize I think it's better to discuss this before actually using it in VitePress.

For example, it might be better to have it under `@vue/` organization. And probably generalize it so that users can actually use it for more general use.

Downside of externalizing the package is that it gets harder to tweak the code in dev. For example, if we have issues in VitePress that require to modify the package, we need to link it locally, publish the package first, etc... It's way easier to maintain if the code is directly under the VitePress project.

So for example if vite-plugin-vue-markdown will not use it for some reason, and if it's not that attractive to be used on other projects, it only leaves the downside of having more complexity, which I would like to avoid.

Let's discuss this with Vue team and see what would be the best fit.

Overall to be clear, I'm not against the approach and I think it's really good idea to have generic Vue Markdown parsing plugin. But If we were to have one, and for VitePress to depend on it, I would like it to be more well thought 👍

stafyniaksacha:
What about using MDC syntax introduced by Nuxt content? 
- https://content.nuxtjs.org/guide/writing/mdc
- https://github.com/nuxt/content/tree/main/src/runtime/markdown-parser

kiaking:
> What about using MDC syntax introduced in Nuxt content?

I would say not for now. Because VitePress is all about simplicity. If you want more powerful features, you should use just Nuxt Content instead.

But, it would be cool if users can opt-in MDC somehow. Like having `vitepress-mdc` package or something.

meteorlxy:
> For example, it might be better to have it under @vue/ organization. And probably generalize it so that users can actually use it for more general use.

IMO, it might be more "markdown-it scoped" than "vue scoped". As for now I don't think it's properly to put it under `@vue` orgs  🤔 

> Downside of externalizing the package is that it gets harder to tweak the code in dev .... it only leaves the downside of having more complexity

We are not going to externalize all markdown plugins. Some plugins are coupled with specific tool, which are rarely useful in other projects. The externalized plugins should be solid and decoupled enough, and could be useful as a standalone markdown-it plugin (even without vitepress).

@kiaking So should we move our discussion to Discord or keep it here?

kiaking:
> So should we move our discussion to Discord or keep it here?

Yeah to get the general concensus let's do it briefly on Discord. At the moment this this is more of a poltical issues like where the package should belong, who should be mainteiners, publish access, and so on. Probably bringing this up in online MTG once should conclude the dicission 👍

antfu:
I would also concern a bit about externalizing it - since it's only a one-file module and not really complex to maintain within this repo. While externalizing it will debugging harder and releasing cycles longer.

If we are talking about reusability and testing, I  would suggest we keep them under this repo as monorepo so we could have their separate tests and publish them for other plugins to reuse.

kaceo:
If the intention is to separate the markdown logic into external packages, I hope you consider adding my suggestions about hoistng multiple <script> <style> blocks (instead of only one block) per markdown file
https://github.com/vuejs/vitepress/issues/607

Vitepress is basically a preprocessor for Vite, and transparent MD parse to SFC is a core principle of Vitepress.  Hence it should support the full possibilities of Vue/Vite, including extending SFC syntax into other types of markups, like route blocks, graphql, mermaid markup, music markup, etc.

meteorlxy:
@kaceo  It's already supported for sure.

https://github.com/mdit-vue/mdit-vue/tree/main/packages/plugin-sfc#customblocks

PR #850, opened 2022-06-24
refactor(theme-default): CSS to SCSS

jd-solanki:
This PR:
- [x] Refactors theme-default code from CSS to SCSS
- [ ] Allows `.vp-raw` class to isolate vp-doc styles


kiaking:
No, we're not using SCSS. I like the idea of adding `.vp-raw` but please refrain from adding SCSS. I would be OK to add postcss plugin for nesting syntax though.

PR #855, opened 2022-06-25
fix: ESM SSR

sapphi-red:
@brc-dd

Now SSR uses ESM. So dynamic import is used instead of `require`.
This means it now requires `await`.

refs https://github.com/vitejs/vite/issues/8759


brc-dd:
Thanks a lot!! 🙏

PR #860, opened 2022-06-25
fix(theme): fix blockquote overflow (with many code childs)

tarikjabiri:
Hi

This is a fix for the blockquote overflow with many code elements: 👇
bug-overflow
The result after the fix: :tada:
fix-overflow

You can see the problem directlty in [my docs](https://dxf.vercel.app/guide/tables.html#style-record), or use this markdown snippet:

```md
> `Element example`, `Element example`, `Element example`, `Element example`, `Element example`, `Element example`, `Element example`, `Element example`, 
```

Thank you for this amazing project I like it.

### DIFF
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -235,6 +235,8 @@
 }
 
 .vp-doc :not(pre) > code {
+  display: inline-block;
+  line-height: initial;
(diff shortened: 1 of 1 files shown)


brc-dd:
After this PR:

image

Before this PR:

image

I don't think that this is a proper fix. Also in your case the overflow is because you have no space after commas.

PR #862, opened 2022-06-26
docs: add ignoreDeadLinks

xfq:
This PR alphabetizes the configs in https://vitepress.vuejs.org/config/app-configs.html and adds `ignoreDeadLinks`.


brc-dd:
Thanks!

PR #865, opened 2022-06-26
docs: modify ts language and enable https links

zihaoyy:
Modified ts language type and enabled more secure https links


brc-dd:
The docs state:

> These values are used to call carbon CDN script as shown below.
> `//cdn.carbonads.com/carbon.js?serve=${code}&placement=${placement}`

You added the protocol but we don't do that to prevent mixed content issues.

https://github.com/vuejs/vitepress/blob/baf8083e668ce50a9d26d9398609e196237ebc95/src/client/theme-default/components/VPCarbonAds.vue#L18

PR #867, opened 2022-06-26
fix(theme): change sponsor link class name to bypass adblock

### LINKED ISSUE #866: `sponsor-link` hidden when `ad block` present
### Describe the bug

When adding `VPTeamMembers` with members with sponsor entry, all sponsor links will be hidden if the `ad block` addon present in the browser.

### Reproduction

The branch on Vitest repo: https://github.com/vitest-dev/vitest/tree/userquin/docs-update-theme

If https://github.com/vitest-dev/vitest/pull/1547 merged, just use the main branch.

The `ad block` addon must be installed in the browser to reproduce the issue.

Check also last comments on previous PR, failing in Chrome in MacOS.

Remove the `SponsorLinkFix` plugin from `docs/vite.config.ts`, then:
- run from root folder `pnpm install`
- run from root folder `pnpm run docs:serve`

### Expected behavior

The `sponsor-link` should be always visible.

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz   
    Memory: 46.29 GB / 63.95 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 102.0.5005.115
    Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.37)
  npmPackages:
    vitepress: ^1.0.0-alpha.4 => 1.0.0-alpha.4
```

### Additional context

The `SponsorLinkFix` plugin in the Vitest repo just rename any class with the  `sponsor`  key to `xsponsor` in `VPTeamMembersItem` component: the solution should avoid using the `sponsor` key.

### Validations

userquin:
closes #866

### DIFF
--- a/src/client/theme-default/components/VPTeamMembersItem.vue
+++ b/src/client/theme-default/components/VPTeamMembersItem.vue
@@ -39,9 +39,9 @@ defineProps<{
         </div>
       </div>
     </div>
-    <div v-if="member.sponsor" class="x-sponsor">
-      <VPLink class="x-sponsor-link" :href="member.sponsor" no-icon>
-        <VPIconHeart class="x-sponsor-icon" /> Sponsor
+    <div v-if="member.sponsor" class="vp-sponsor">
+      <VPLink class="vp-sponsor-link" :href="member.sponsor" no-icon>
+        <VPIconHeart class="vp-sponsor-icon" /> Sponsor
       </VPLink>
     </div>
   </article>
@@ -186,7 +186,7 @@ defineProps<{
   height: 56px;
 }
 
-.x-sponsor-link {
+.vp-sponsor-link {
   display: flex;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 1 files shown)


brc-dd:
`x-sponsor-link` is also likely to be hidden by some ad blockers. Lets change it to something like `sp-link`.

---

We should also track Carbon Ads (separately), currently this is what I get on my browser (using uBlock Origin):

image

I was hoping we could have something similar to Nuxt here.

image

There was an effort for this #348, but that PR went stale. The fix seemed quite straightforward. If you wish (and have some time to implement), we can have that here itself or maybe a separate PR by anyone. Either works for me.

userquin:
upps, so we should ommit the `sponsor` key, fixing it...

brc-dd:
Yeah, this one is fine. Using `vp-sponsor-foo` should also not cause issues with most of the ad-blockers, but that will probably conflict with our sponsors components.

brc-dd (approved):

Issue #861, opened 2022-06-26
Add support for cs files! Yes I do use vitepress to document C# libraries 😁


emersonbottero:
### Is your feature request related to a problem? Please describe.

Although C# is one of the available languages it does not recognize cs extension while trying to use <<< to import code.

```bash
[vite] Internal server error: No language registration for cs
```

### Describe the solution you'd like

I'm not sure how it work internally but it seems in only missing the recognition of cs files as C# language!

### Describe alternatives you've considered

If it can be done by default great.
If this needs an configuration maybe updating the docs would suffice.

### Additional context

I have support for other languages would improve the usage (and popularity) of vitepress 🤓 

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

xfq:
Did you try `c#` instead of `cs`?

emersonbottero:
> Have you tried `csharp` / `c#` instead of `cs`?
> 
> The currently supported languages are documented in https://github.com/shikijs/shiki/blob/main/docs/languages.md#all-languages

I can't replace the file extensions I'm using in the library.

I guess you misunderstood... 
I'm not trying to use ``` sintax

I'm using <<<

xfq:
Sorry, I misunderstood. You're right, VitePress doesn't support the `<<< @/test.cs` syntax indeed.

brc-dd:
Thanks for pointing this out. We are probably just adding the extension as the language in our snippet plugin. We should instead be mapping the extension to language the way linguist does. Also, even in fenced code blocks, we don't currently show language on top right for C# I guess. We have a hard coded list in our CSS to do that. But I guess this can be fixed in the code block plugin itself (the way we added copy, we can add another span telling the language and just style it from CSS, it will eliminate the need for hard coding).

yxw007:
<<< @/snippets/csharp-snippets.cs{csharp}

It's work ok

yxw007:
> ### Is your feature request related to a problem? Please describe.
> Although C# is one of the available languages it does not recognize cs extension while trying to use <<< to import code.
> 
> ```shell
> [vite] Internal server error: No language registration for cs
> ```
> 
[...]

Issue #471, opened 2021-12-14
Print variable values in code snippets


deevannaveed:
### Is your feature request related to a problem? Please describe.

Basically I want to add dynamic values in code snippets see the example below, I want the links to have a dynamic version number.

`MyMarkDown.md`

```
<script setup>
  import { version } from '../../../packages/myPackage/package.json'
</script>
# Below is a code snippet that will contain our latest version.

```html
<!-- stylesheet -->
<link rel="stylesheet" href="https://mycdn.com/css/{{ version }}/style.min.css" />

<div>Some content</div>

<!-- 🚨 Script must be loaded after the component -->
<script src="https://mycdn.com/js/{{ version }}/script.min.js" ></script>
```

### Describe the solution you'd like

I want to be able to print dynamic values in code snippets as show in the example above.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/master/.github/contributing.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

kiaking:
Interesting 👀 Not sure how to achieve this, but if anyone knows a good way to do it, I'm curious to know.

brc-dd:
We are adding `v-pre` here: https://cs.github.com/vuejs/vitepress/blob/19b0758a04e9fb7863b3a961024dfe1137fbe928/src/node/markdown/plugins/highlight.ts

I have a suggestion, users can do something like this:

````
```foo-vue
Don't add v-pre if -vue suffix is there.
Also remove suffix from lang so that highlighter don't complaint.
```

```foo
Otherwise add v-pre.
```
````

**Comment by brc-dd** [comment]
Okay so this doesn't always work. Since markdown parser is run first, `{{` `}}` syntax may get highlighted so vue won't be able to substitute the values. For example in this result:

```html
<span style="color: #A6ACCD">(</span><span style="color: #89DDFF">{</span><span style="color: #A6ACCD">{ z </span><span style="color: #89DDFF">}</span><span style="color: #A6ACCD">})</span>
```

PR #875, opened 2022-06-28
feat(build): improve code blocks and snippets

brc-dd:
fixes:
- #861
- #471
- #884


emersonbottero:
Shouldn't It recognize the language by the extension?
cs => c#
py => python 

and so on?

brc-dd:
@emersonbottero Read this section https://deploy-preview-875--vitepress-docs.netlify.app/guide/markdown.html#import-code-snippets

You'll have to provide language in case of C#. For `*.py`, the language will be auto detected as the extension `py` is a valid language specifier.

emersonbottero:
> @emersonbottero Read this section https://deploy-preview-875--vitepress-docs.netlify.app/guide/markdown.html#import-code-snippets
> 
> You'll have to provide language in case of C#. For `*.py`, the language will be auto detected as the extension `py` is a valid language specifier.
> 
> I have not implemented language-extension mapping because it will be a fairly long list that we'll need to hard code. It is better to rely on users to override the detected language (the extension of file).
> 
> (If your file extension is in this list, then you don't need to specify language: https://github.com/shikijs/shiki/blob/main/docs/languages.md#all-languages)

Oh..

Ho, If I ask tô chance
 | 'csharp' | 'c#' 
to
| 'csharp' | 'c#' | 'cs' 
In the shiki lib It should work them? ... 🤔

brc-dd:
Yeah it would work then.

meteorlxy:
The syntax from [markdown-it-vuepress-code-snippet-enhanced](https://github.com/fabioaanthony/markdown-it-vuepress-code-snippet-enhanced) looks more markdown-like and extendable (I saw you have noticed that @brc-dd ), so I chose to implement it in vuepress-next (not as powerful as that plugin though).

Anyway, the `<<<` syntax has being always marked as `beta` in VuePress v1 ([link](https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets)), so we could still consider changing it.

brc-dd:
But I don't think it supports importing regions from files. Or does it?

meteorlxy:
It does: https://github.com/fabioaanthony/markdown-it-vuepress-code-snippet-enhanced#transclusion

I'd say that plugin it really powerful 😅

brc-dd:
@meteorlxy So should we go ahead with this PR or implement that syntax? This PR fixes two other issues, I'll split it. We can track snippets (#861) separately from issues with code blocks.

emersonbottero:
> @meteorlxy So should we go ahead with this PR or implement that syntax? This PR fixes two other issues, I'll split it. We can track snippets (#861) separately from issues with code blocks.

even with the plugin I think it still makes sense to keep it in vitepress since it already has the feature to import code out of the box.

brc-dd:
Let's merge this for now, we can work on changing the importing format later.

emersonbottero:
https://github.com/shikijs/shiki/pull/351

PR #881, opened 2022-06-30
fix: wrong import of shared.ts file

zonemeen:


brc-dd:
The imports are correct. You need to run build command first. It copies them to respective locations.

PR #886, opened 2022-07-01
fix: prevent line numbers from being copied. Closes #884

### LINKED ISSUE #884: Copy to clipboard feature also adding line numbers
### Describe the bug

This is an issue with the Vitepress alpha version. We have some code blocks with line numbers enabled in markdown. It would appear that using the `copy to clipboard` button is also somehow grabbing the line numbers along with it. 

### Reproduction

Attaching our docs link here: https://dev.fairdataihub.org/docs/README.html#tech-stack
You can click on the copy to clipboard button to try it out

### Expected behavior

Only the code block content should be copied. 

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (20) x64 Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz
    Memory: 16.33 GB / 31.81 GB
  Binaries:
    Node: 16.14.2 - ~\anaconda3\envs\nuxt-env\node.EXE
    Yarn: 1.22.11 - ~\anaconda3\envs\nuxt-env\yarn.CMD
    npm: 8.12.2 - ~\Desktop\fairdataihub-docs\node_modules\.bin\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (103.0.1264.37)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    vitepress: 1.0.0-alpha.4 => 1.0.0-alpha.4

Not sure if these are correct since I'm on windows 11...
```

### Additional context

_No response_

### Validations

megasanjay:
There is currently an issue where having line numbers enabled in your markdown causes the `copy to clipboard` button to also add in those line numbers at the end. 

This PR now looks at the `<code>` tag to read the `innertext` contents. This skips the line numbers from being added to the copied text. 

Tested locally with `lineNumbers` both enabled and disabled.

### DIFF
--- a/src/client/theme-default/composables/copy-code.ts
+++ b/src/client/theme-default/composables/copy-code.ts
@@ -76,7 +76,18 @@ function handleElement(el: HTMLElement) {
       parent.classList.contains('language-sh') ||
       parent.classList.contains('language-bash')
 
-    let { innerText: text = '' } = parent
+    // Select the <code> element inside the whole code block.
+    // The <code> element is currently nested inside the <pre> tag.
+    const codeBlockBody = parent.querySelector(
+      'pre > code'
+    ) as HTMLElement | null
+
+    let text: string = '' // Give it a default value
+
+    // if the code block body is not null, then we can get the text from it
+    if (codeBlockBody) {
+      text = codeBlockBody?.innerText ?? ''
+    }
 
     if (isShell) {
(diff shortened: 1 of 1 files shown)


brc-dd:
Hi! Thanks for the PR. But I had changed the code copying logic in #875. It appears to be fixed there.

PR #899, opened 2022-07-03
fix(theme): make code blocks horizontally scrollable

hir0kio:
See [my proof of concept](https://stackblitz.com/edit/node-ggaugv?file=docs%2Findex.md).

When there is a [code block](https://www.markdownguide.org/basic-syntax#code-blocks) containing a long line, the block overflows horizontally. My change sets `pre`'s `overflow-x` to `auto` to make code blocks horizontally scrollable.

### DIFF
--- a/src/client/theme-default/styles/base.css
+++ b/src/client/theme-default/styles/base.css
@@ -107,6 +107,10 @@ samp {
   font-family: var(--vp-font-family-mono);
 }
 
+pre {
+  overflow-x: auto;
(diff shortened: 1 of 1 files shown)


brc-dd:
This is not a proper fix. We are using `pre` at other places too. This will mess that up. It is better to use the fenced block syntax:

````md
```
Lorem ipsum dolor sit amet,
```
````

It won't overflow your container.

PR #900, opened 2022-07-03
feat: add head icon

Rain120:
## What had I do

- [x] Doc Tab Icon

### Before vs After

image


brc-dd:
That's Vite logo. Not VitePress'. We'll add the favicon after rebranding.

PR #901, opened 2022-07-03
docs: guidelines for adding custom dark mode styles

xfq:


brc-dd (commented):
Add this here itself: https://vitepress.vuejs.org/config/app-configs.html#appearance

brc-dd:
That's not about customizing the default theme. Whatever theme you use, setting `appearance` true will inject that script (mentioned here: https://vitepress.vuejs.org/config/app-configs.html#appearance). There is no `appearance` property in `themeConfig`. In the theme section of docs, only those properties that can be put inside the `themeConfig` object are written.

xfq:
Yeah, I understand this, but our understanding of customizing the default theme is different. I consider changing the default style to be customizing the default theme. Anyway, thanks for your reply.

PR #906, opened 2022-07-03
Added logo for Vitepress.

otabekoff:
Used Vite colors as background list with 50% opacity.
Used brigter colors that are closer to Vue colors on the front list.
Added lightning logo to the center with white color that means fast and merging that two lists(vite+vue=vitepress).


otabekoff:
Also, If you like any of theese from the below image. Then, I will share figma link or just update the logo file from github. 

logos-for-vitepress

meteorlxy:
Thanks for your design ❤️ .

However, we might not consider adding a logo yet because of #548. We may keep the logo of VuePress or choose a new one after that. You can post in that thread or open another discussion for this.

PR #908, opened 2022-07-03
Updated docs/config/app-configs.md for outlineTitle property.

otabekoff:


xfq:
`outlineTitle` seems to be a Theme Config instead of an App Config.

otabekoff:
> `outlineTitle` seems to be a Theme Config instead of an App Config.

Yes, right. How I add this to app config, I don't know. I am changing it.

otabekoff:
> `outlineTitle` seems to be a Theme Config instead of an App Config.

I did some changes.

PR #911, opened 2022-07-03
docs: add section title to all pages

userquin:
This PR also includes the `prev/next` frontmatter entries on `last/first` section pages repectively.

### DIFF
--- a/docs/config/app-configs.md
+++ b/docs/config/app-configs.md
@@ -1,3 +1,7 @@
+---
+title: App Configs | Config
+---
+
 # App Configs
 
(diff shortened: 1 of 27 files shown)


brc-dd:
I don't think we need the context on the titles. IMO the current titles looks much cleaner. Also, spelling of introduction is wrong 😅.

userquin:
it is about context on the title: just check the title with this PR and the current docs.

kiaking:
Yeah I think this is kinda too much (sorry for the late reply by the way). We already have `??? | VitePress` so adding more pipes is not worth it I think 👀 Thanks for the PR though!

PR #935, opened 2022-07-07
Added Telegram icon

otabekoff:


otabekoff:
Now you can check the changed files and if everything is OK, you can merge it.
And if there is any other thing to be changed, leave note here.
Telegram now has over 700 million users. And it should be in the list.

brc-dd:
Please go and at least read the comment I linked on your last PR.

PR #948, opened 2022-07-08
fix(build): cache key should consider file path

### LINKED ISSUE #947: can not navigate of pages with same content
### Describe the bug

When pages(.md) have same content, we can't navigate in these pages which has same content in dev mode.

### Reproduction

https://stackblitz.com/edit/github-zbjrsg-bcngzg?file=docs%2Fguide%2Faaa.md,docs%2Fguide%2Fbbb.md

1. wait for the dev server launch
2. click `a` in sidebar
3. click `b` in sidebar

The `b` link won't be active and the page is not navigated to `b.md` at all.

### Expected behavior

`b` link is active and the page is navigated to `b.md`

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.11 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.19.3 - ~/Library/Caches/fnm_multishells/46430_1657095997514/bin/node
    npm: 6.14.17 - ~/Library/Caches/fnm_multishells/46430_1657095997514/bin/npm
```

### Additional context

I have made an investigate and found it's because the cache key. I'll come up with a PR later.

### Validations

fi3ework:
fix #947

### DIFF
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -43,8 +43,9 @@ export async function createMarkdownToVueRenderFn(
   ): Promise<MarkdownCompileResult> => {
     const relativePath = slash(path.relative(srcDir, file))
     const dir = path.dirname(file)
+    const cacheKey = JSON.stringify({ src, file })
 
-    const cached = cache.get(src)
+    const cached = cache.get(cacheKey)
     if (cached) {
       debug(`[cache hit] ${relativePath}`)
       return cached
@@ -148,7 +149,7 @@ export async function createMarkdownToVueRenderFn(
       deadLinks,
       includes
     }
-    cache.set(src, result)
+    cache.set(cacheKey, result)
     return result
   }
 }


brc-dd:
IMO just changing `cached` to this would work:

```ts
    const cached = cache.get(JSON.stringify({ src, file }))
```

There is no need for complex functions as `src` and `file` are strings and their order will always be same. So stringifying `{src, file}` should yield same result every time if the content is same.

fi3ework:
I though node.js's `JSON. stringify` was not stable, but it does. So yes, we do not need the hash function anymore.

fi3ework:
> I have a doubt too. What would be the need of caching then? `file` cannot be same for two files. Will there ever be a hit?

AFAIK, the cache will be hit in `handleHotUpdate` stage

brc-dd:
> AFAIK, the cache will be hit in `handleHotUpdate` stage if you're modifying the content the same as some cached content.

Okay, I tested with debug, every markdown file is rendered twice during build, so setting cache reduces nearly half the time. It is still being triggered. And in dev server, a cache hit occurs if you change some content, save it, then undo the changes and then save it.

brc-dd:
Thanks for the PR!

PR #957, opened 2022-07-10
feat(type): make LocaleLinks text nullable

Sepush:
Signed-off-by: Sepush <sepush@outlook.com>

We don't want the text in localeLinks.
Make it nullable so we can config localeLinks in the following way.
image
image

### DIFF
--- a/src/client/theme-default/components/VPNavBarExtra.vue
+++ b/src/client/theme-default/components/VPNavBarExtra.vue
@@ -11,7 +11,7 @@ const { site, theme } = useData()
 <template>
   <VPFlyout class="VPNavBarExtra" label="extra navigation">
     <div v-if="theme.localeLinks" class="group">
-      <p class="trans-title">{{ theme.localeLinks.text }}</p>
+      <p v-if="theme.localeLinks.text" class="trans-title">{{ theme.localeLinks.text }}</p>
 
       <template v-for="locale in theme.localeLinks.items" :key="locale.link">
         <VPMenuLink :item="locale" />
--- a/src/client/theme-default/components/VPNavBarTranslations.vue
+++ b/src/client/theme-default/components/VPNavBarTranslations.vue
@@ -14,7 +14,7 @@ const { theme } = useData()
     :icon="VPIconLanguages"
   >
     <div class="items">
-      <p class="title">{{ theme.localeLinks.text }}</p>
+      <p v-if="theme.localeLinks.text" class="title">{{ theme.localeLinks.text }}</p>
 
       <template v-for="locale in theme.localeLinks.items" :key="locale.link">
         <VPMenuLink :item="locale" />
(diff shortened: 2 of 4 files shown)


brc-dd:
We need some more discussion on this. If we want it like this, then we should do something to show what's the current language. Also in `VPNavBarTranslations.vue`, you need to add `v-if` to `<p class="title">`, otherwise it'll have empty space there.

Sepush:
> We need some more discussion on this. If we want it like this, then we should do something to show what's the current language. Also in `VPNavBarTranslations.vue`, you need to add `v-if` to `<p class="title">`, otherwise it'll have empty space there. `VPNavScreenTranslations.vue` needs some patches too. It looks weird to have no text on the button:
> 
> image

Current language will be heightlight so no text here is not a problem.
If text is required then we need auto change the text when language changed.

For some project the have different repos for languages they could config `localLinks.text` in different repos config,but there are some project the choose to have a subdirectory have the same structure of others language vesion doc under the docs dir.

image

 In summary I think make LocalLinks.text be nullable is reasonable.

brc-dd:
Yeah I understand that it is reasonable. We probably will be supporting i18n out of the box before v1. But some things need to be changed when making it optional as I pointed out in my earlier comment.

> Current language will be highlighted so no text here is not a problem.

How is this being done? I don't think anything will be highlighted. We aren't using `isActive` or something like that with languages.

> If text is required then we need auto change the text when language changed.

Yeah I like this approach. This can be implemented. If text is blank then pick up current language and put it on top.

Sepush:
> 
> > Current language will be highlighted so no text here is not a problem.
> 
> How is this being done? I don't think anything will be highlighted. We aren't using `isActive` or something like that with languages.
> > I mean in current VitePress the active language will have a green color by default 
image

Sepush:
>Yeah I understand that it is reasonable. We probably will be supporting i18n out of the box before v1. But some things need to be changed when making it optional as I pointed out in my earlier comment.

So expected the out of box i18n support and the docs for it and When we could use it.

brc-dd:
> I mean in current VitePress the active language will have a green color by default

This isn't implemented that way. Suppose open some page inside `/en/` (or `/zh/`) it will no longer will be highlighted.

Sepush:
> > I mean in current VitePress the active language will have a green color by default
> 
> This isn't implemented that way. Open some other page inside `/en/` (or `/zh/`) it will no longer will be highlighted.

Oh,I got it.

brc-dd:
> So expected the out of box i18n support and the docs for it and When could we use it.

I'll try to prioritize it. But can't tell precisely when it'll land. Might take some time.

brc-dd:
So, I'm working on something. You can see the progress here: https://github.com/vuejs/vitepress/compare/chore/vite-3...feat/i18n

I've fixed this issue there. It now shows the current language. So suppose someone has config like this:

```js
  locales: {
    root: {
      label: 'English'
    },
    fr: {
      label: 'French'
    }
  }
```

It now shows English and French there. If some page inside `/fr/` is open, then it'll show `French` as the current language otherwise it'll show `English`. I am working on making other things like content inside head (title, meta, description), and stuff in theme config work with locales. Will probably finish this within the week.

Thanks for the PR but I'm gonna close this in favor of the development on feat/i18n branch.

Issue #785, opened 2022-06-13
Show dynamic headers in aside outline


ElliotThiebaut:
### Describe the bug

If you use Vue-style interpolation in a markdown title (# in markdown), the aside curtain "On this page" will show the Vue syntax and not the final text. It will also not take into account the fact that multiple secondary title exists from a loop.

### Reproduction

If you have a markdown file like this with a vfor loop and a interpolation of text for the heading:
![image](https://user-images.githubusercontent.com/60610988/173346686-c53c23d0-bec9-4b67-9663-a3d776562802.png)
```md
# My great page
## {{element.subtitle}}
```

The output will show `{{element.subtitle}}` as text in the aside curtain "On this page".
![image](https://user-images.githubusercontent.com/60610988/173346767-3c1b21fb-1712-48b2-8213-7d5dd7bce18e.png)

### Expected behavior

The output in the aside curtain "On this page" should be the string returned by Vue-style interpolation and take into account the multiple instance of the loop.
![image](https://user-images.githubusercontent.com/60610988/173347004-43c676dc-9ca5-408f-8de3-19a38554366b.png)

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
    Memory: 18.71 GB / 31.73 GB
  Binaries:
    Node: 18.2.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.9.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (102.0.1245.39)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    vitepress: ^1.0.0-alpha.1 => 1.0.0-alpha.1
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
Putting Vue-style interpolations in Markdown headings also messes up automatic element IDs (for anchors).

PR #965, opened 2022-07-11
feat(theme): support nested headings in outline

### LINKED ISSUE #954: expose a configurable outline config in frontmatter

For now, the outline config in frontmatter is not documented, however, the code is in the codebase. Just add `outline: 'deep'` to the page frontmatter and the outline can show 2 levels at most. So I think:

1. configuration of page outline of frontmatter should be documented.
2. The outline now only has one available type `'deep'`, but I think we should expose a more configurable config, such as `['h1', 'h2']` and an another `deep` type which equals to `['h1', 'h2', 'h3', 'h4', 'h5', 'h6']` for short.
3. I found that `page.headers` is resolved from https://github.com/vuejs/vitepress/blob/22006e8d6e3ed45841979d684eb6a4ef999bd707/src/node/markdown/plugins/headings.ts#L6-L23 I think we can set the default value of `headingPlugin` to `['h1', 'h2', 'h3', 'h4', 'h5', 'h6']` which means markdown-it will parse all headings in node side. And the outline configuration will only control which headings to render in the outline.

The configuration of the outline will be like

```ts
Type: ('h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6')[] | 'deep'
Default: ['h2', 'h3']
```

also, we need to fix the code which only can handle `[2, 3]` level for now.

I believe parsing all headings will not cause a performance issue because markdown-it will try to iterate over all blocks (only a guess).

### Additional context

I would like to implement this feature.

### Validations

fi3ework:
resolve #954

- [x] no breaking change
- [x] test added
- [x] new frontmatter config added

### DIFF
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -2,13 +2,14 @@ import { Ref, computed, onMounted, onUpdated, onUnmounted } from 'vue'
 import { Header, useData } from 'vitepress'
 import { useAside } from '../composables/aside'
 import { throttleAndDebounce } from '../support/utils'
+import { PageData } from '../../../../types/shared'
 
 interface HeaderWithChildren extends Header {
   children?: Header[]
   hidden?: boolean
 }
 
-interface MenuItemWithLinkAndChildren {
+export interface MenuItemWithLinkAndChildren {
   text: string
   link: string
   children?: MenuItemWithLinkAndChildren[]
@@ -30,24 +31,67 @@ export function useOutline() {
   }
 }
 
-export function resolveHeaders(headers: Header[]) {
-  return mapHeaders(groupHeaders(headers))
+export function resolveHeaders(
+  headers: Header[],
+  levelsRange: PageData['frontmatter']['outline'] = 2
+) {
+  const levels: [number, number] =
+    typeof levelsRange === 'number'
+      ? [levelsRange, levelsRange]
+      : levelsRange === 'deep'
+      ? [1, 6]
+      : levelsRange
+
+  return mapHeaders(groupMarkdownHeadersByChildren(headers, levels))
 }
 
-function groupHeaders(headers: Header[]): HeaderWithChildren[] {
-  headers = headers.map((h) => Object.assign({}, h))
+function addToParent(
+  currIndex: number,
+  headers: HeaderWithChildren[],
+  levelsRange: [number, number]
+): boolean {
+  if (currIndex === 0) {
+    return true
+  }
 
-  let lastH2: HeaderWithChildren | undefined
+  const currentHeader = headers[currIndex]
+  for (let index = currIndex - 1; index >= 0; index--) {
+    const header = headers[index]
 
-  for (const h of headers) {
-    if (h.level === 2) {
-      lastH2 = h
-    } else if (lastH2 && h.level <= 3) {
-      ;(lastH2.children || (lastH2.children = [])).push(h)
+    if (
+      header.level < currentHeader.level &&
+      header.level >= levelsRange[0] &&
+      header.level <= levelsRange[1]
+    ) {
+      if (!header.children) {
+        header.children = []
+      }
+      header.children.push(currentHeader)
+      return false
     }
   }
 
-  return headers.filter((h) => h.level === 2)
+  return true
+}
+
+function groupMarkdownHeadersByChildren(
+  headers: Header[],
+  levelsRange: [number, number]
+): HeaderWithChildren[] {
+  headers = headers.map((h) => Object.assign({}, h))
+
+  const result: HeaderWithChildren[] = []
+
+  headers.forEach((h, index) => {
+    if (h.level >= levelsRange[0] && h.level <= levelsRange[1]) {
+      const shouldAddToRoot = addToParent(index, headers, levelsRange)
+      if (shouldAddToRoot) {
+        result.push(h)
+      }
+    }
+  })
+
+  return result
 }
 
 function mapHeaders(
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -35,6 +35,10 @@ export function resolveHeaders(
   headers: Header[],
   levelsRange: DefaultTheme.Config['outline'] = 2
 ) {
+  if (levelsRange === false) {
+    return []
+  }
+
   const levels: [number, number] =
     typeof levelsRange === 'number'
       ? [levelsRange, levelsRange]
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -2,7 +2,7 @@ import { Ref, computed, onMounted, onUpdated, onUnmounted } from 'vue'
 import { Header, useData } from 'vitepress'
 import { useAside } from '../composables/aside'
 import { throttleAndDebounce } from '../support/utils'
-import { PageData } from '../../../../types/shared'
+import type { DefaultTheme } from 'vitepress/theme'
 
 interface HeaderWithChildren extends Header {
   children?: Header[]
@@ -33,7 +33,7 @@ export function useOutline() {
 
 export function resolveHeaders(
   headers: Header[],
-  levelsRange: PageData['frontmatter']['outline'] = 2
+  levelsRange: DefaultTheme.Config['outline'] = 2
 ) {
   const levels: [number, number] =
     typeof levelsRange === 'number'
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -39,7 +39,7 @@ export function resolveHeaders(
     typeof levelsRange === 'number'
       ? [levelsRange, levelsRange]
       : levelsRange === 'deep'
-      ? [1, 6]
+      ? [2, 6]
       : levelsRange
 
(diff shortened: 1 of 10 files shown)


fi3ework:
@brc-dd Friendly ping. Could you take a review or is there something else I could provide to this PR to ensure the functionality?

brc-dd:
Hi! Kia will have a look on this.

src/client/theme-default/composables/outline.ts
```diff
@@ -2,13 +2,14 @@ import { Ref, computed, onMounted, onUpdated, onUnmounted } from 'vue'
 import { Header, useData } from 'vitepress'
 import { useAside } from '../composables/aside'
 import { throttleAndDebounce } from '../support/utils'
+import { PageData } from '../../../../types/shared'
```
brc-dd, line 5: This one probably can be imported from `../../shared` itself.

src/client/theme-default/composables/outline.ts
```diff
@@ -2,13 +2,14 @@ import { Ref, computed, onMounted, onUpdated, onUnmounted } from 'vue'
 import { Header, useData } from 'vitepress'
 import { useAside } from '../composables/aside'
```
brc-dd, line 3:
```suggestion
import { useAside } from './aside'
```

types/shared.d.ts
```diff
@@ -8,7 +8,9 @@ export interface PageData {
   titleTemplate?: string | boolean
   description: string
   headers: Header[]
-  frontmatter: Record<string, any>
+  frontmatter: Record<string, any> & {
+    outline?: number | [number, number] | 'deep'
```
brc-dd, line 12: allow it to be disabled too?

src/client/theme-default/composables/outline.ts
```diff
@@ -30,24 +31,67 @@ export function useOutline() {
...
-  return mapHeaders(groupHeaders(headers))
+export function resolveHeaders(
+  headers: Header[],
+  levelsRange: PageData['frontmatter']['outline'] = 2
+) {
+  const levels: [number, number] =
+    typeof levelsRange === 'number'
+      ? [levelsRange, levelsRange]
+      : levelsRange === 'deep'
+      ? [1, 6]
```
brc-dd, line 42: Changing `deep` is actually a breaking change for some projects (like vitest) although it wasn't documented. Also if we want to include all headers in `deep`, then I think we should make it `[2, 6]`. Only one `h1` heading per page is usually recommended. It doesn't make sense adding the page title to outline.

brc-dd: Add and document both `deep` and `all`? 😅

types/shared.d.ts
brc-dd: Also, can it be configured through themeConfig too so that users don't need to put that on every page?

src/client/theme-default/composables/outline.ts
fi3ework: I don't prefer to have both `'deep'` and `'all'` together cause it's really confusing for me 😅. Yeah, let's make `deep` to `[2, 6]` and mark it in the changelog so the Vitest doc maintainer can know it.

types/shared.d.ts
fi3ework: added `false` to disable the outline

fi3ework:
Made an update with a new commit and some refactoring. And I think I made another breaking change 😅.

before:

There would be an outline title without items but only a title when the content does not have h2/h3 headings, such as `##### xx`.

now: 

Outline will only be displayed when there are really outline items to display.

fi3ework:
> Rest of the stuff seems fine. I still haven't checked the UI though.

The UI is like this, I'm also wondering is there a better way to release a preview build for custom config and docs that can reveal the different UI.

brc-dd:
> Made an update with a new commit and some refactoring. And I think I made another breaking change 😅.
> 
> before:
> 
> There would be an outline title without items but only a title when the content does not have h2/h3 headings, such as `##### xx`.

This was probably a bug 😅.

brc-dd (dismissed):

fi3ework:
I created a stackblitz project to preview the affection on UI. 

- project https://github-b4nsvg--3000.local.webcontainer.io/
- live show https://github-b4nsvg--3000.local.webcontainer.io/guide/what-is-vitepress.html

I think we can do more things to make preview more easy in two ways:

1. Provide a stackblitz like which will be created from the branch of the PR via a comment by GitHub bot for each PR automatically. So the reviewer can know the UI affection by changing some file or the PR author could provide the demo as above.
2. Add example docs that contain different use cases of Vitepress which could also be used by E2E test, and deploy a preview link by Netlify for each PR.

sishen654:
@fi3ework 
Is there a way not to add this on every file: outline: "deep", otherwise it would be too much trouble

fi3ework:
> @fi3ework Is there a way not to add this on every file: outline: "deep", otherwise it would be too much trouble

Yes, it should support. Seems like I lost some code with the app configuration. 😅 I'll bring it back later.

sishen654:
> > @fi3ework Is there a way not to add this on every file: outline: "deep", otherwise it would be too much trouble
> 
> Yes, it should support. Seems like I lost some code with the app configuration. 😅 I'll bring it back later.
Haha, trouble you,😁😁😁

fi3ework:
Ready to be reviewed again.

---

I kind of want to add preview deploy for examples before this PR. But multiple sites deploy Netlify ([doc](https://docs.netlify.com/configure-builds/monorepos/#base-directory)) require setting permission to add a new site for the example site that I don't have. If there is any vuejs member who has permission also thinks of a preview deployment of examples to facilitate observing UI changes of code, let me know and I would like to come up with the PR with your help of setting Netlify.

fi3ework:
Ready to be reviewed again.

---

I kind of want to add preview deploy for examples before this PR. But multiple sites deploy Netlify ([doc](https://docs.netlify.com/configure-builds/monorepos/#base-directory)) require setting permission to add a new site for the example site that I don't have. If there is any vuejs member who has permission also thinks of a preview deployment of examples to facilitate observing UI changes of code, let me know and I would like to come up with the PR with your help of setting Netlify.

trwnh:
would it make sense to add a max depth in this pr?

fi3ework:
> would it make sense to add a max depth in this pr?

max depth is supported in this PR. Using like this - `[2, 4]`

brc-dd:
Can we fix #785 here itself? I can fix that separately too, but then this would need rebase and might cause other issues. Here is the diff on vite-3 branch. You will need to manually apply it. That query selector h2 thing probably need to be made `h2, h3, h4, h5, h6`.

```diff
diff --git a/src/client/theme-default/components/VPDocAside.vue b/src/client/theme-default/components/VPDocAside.vue
index 4232d15..2f590ab 100644
--- a/src/client/theme-default/components/VPDocAside.vue
+++ b/src/client/theme-default/components/VPDocAside.vue
@@ -3,7 +3,7 @@ import { useData } from 'vitepress'
 import VPDocAsideOutline from './VPDocAsideOutline.vue'
 import VPDocAsideCarbonAds from './VPDocAsideCarbonAds.vue'
 
-const { page, theme } = useData()
+const { theme } = useData()
 </script>
 
 <template>
@@ -11,7 +11,7 @@ const { page, theme } = useData()
     <slot name="aside-top" />
 
     <slot name="aside-outline-before" />
-    <VPDocAsideOutline v-if="page.headers.length" />
+    <ClientOnly><VPDocAsideOutline /></ClientOnly>
     <slot name="aside-outline-after" />
 
     <div class="spacer" />
diff --git a/src/client/theme-default/components/VPDocAsideOutline.vue b/src/client/theme-default/components/VPDocAsideOutline.vue
index dad4378..42acb27 100644
--- a/src/client/theme-default/components/VPDocAsideOutline.vue
+++ b/src/client/theme-default/components/VPDocAsideOutline.vue
@@ -1,25 +1,35 @@
 <script setup lang="ts">
-import { ref, computed } from 'vue'
-import { useData } from 'vitepress'
-import {
-  resolveHeaders,
-  useOutline,
-  useActiveAnchor
-} from '../composables/outline.js'
-
-const { page, frontmatter, theme } = useData()
-
-const { hasOutline } = useOutline()
+import { useData, useRoute, type Header } from 'vitepress'
+import { computed, ref, watch } from 'vue'
+import { resolveHeaders, useActiveAnchor } from '../composables/outline.js'
+
+const { frontmatter, theme } = useData()
+const route = useRoute()
+
+const headers = ref<Header[]>([])
+
+watch(
+  () => route.path,
+  () => {
+    document.querySelectorAll('h2').forEach((el) => {
+      if (el.textContent && el.id)
+        headers.value.push({
+          level: Number(el.tagName[1]),
+          title: el.textContent,
+          slug: el.id
+        })
+    })
+  },
+  { immediate: true }
+)
+
+const hasOutline = computed(() => !!headers.value.length)
+const resolvedHeaders = computed(() => resolveHeaders(headers.value))
 
 const container = ref()
 const marker = ref()
-
 useActiveAnchor(container, marker)
 
-const resolvedHeaders = computed(() => {
-  return resolveHeaders(page.value.headers)
-})
-
 function handleClick({ target: el }: Event) {
   const id = '#' + (el as HTMLAnchorElement).href!.split('#')[1]
   const heading = document.querySelector<HTMLAnchorElement>(
```

Charles7c:
Looking forward this feature.

PR #978, opened 2022-07-13
add scroll-behavior:smooth

Jack-Zhang-1314:
* while scrolling, smooth effect, better user  experience

### DIFF
--- a/src/client/theme-default/styles/base.css
+++ b/src/client/theme-default/styles/base.css
@@ -8,6 +8,7 @@ html {
   line-height: 1.4;
   font-size: 16px;
   -webkit-text-size-adjust: 100%;
+  scroll-behavior: smooth;
 }
 
 html.dark {


brc-dd:
This is not as simple as adding a CSS property. Open this for example: https://deploy-preview-978--vitepress-docs.netlify.app/guide/getting-started.html. Click on "What's next" on the right hand side. It won't reach to the top at all, you'll have to click that again. Play around on this page with the outline: https://deploy-preview-978--vitepress-docs.netlify.app/guide/markdown.html. Most of the time you won't get the correct item highlighted if its not adjacent to the current highlighted item. URL hash will also be messed up in certain scenario because we are probably debouncing for 100ms, but the scroll animation may take much longer. Also this needs to be added **only if** the user has `no-preference` about `prefers-reduced-motion`. Fast animations can cause issues like sudden, severe vertigo in some readers.

Jack-Zhang-1314:
> This is not as simple as adding a CSS property. Open this for example: https://deploy-preview-978--vitepress-docs.netlify.app/guide/getting-started.html. Click on "What's next" on the right hand side. It won't reach to the top at all, you'll have to click that again. Play around on this page with the outline: https://deploy-preview-978--vitepress-docs.netlify.app/guide/markdown.html. Most of the time you won't get the correct item highlighted if its not adjacent to the current highlighted item. URL hash will also be messed up in certain scenario because we are probably throttling for 100ms, but the scroll animation may take much longer. Also this needs to be added **only if** the user has `no-preference` about `prefers-reduced-motion`. Fast animations can cause issues like severe vertigo in some readers. Feel free to open an issue where we can have some proper discussion on the implementation.

* Thank for your reply, it's too simple what I thought

PR #980, opened 2022-07-13
feat(theme): support hiding aside component from frontmatter

userquin:
Sometime the space in the right sidebar is required, for example adding some app but we also need the left sidebar, since the sidebar is global we cannot remove the space on the right aside.

For example, in the `vite-plugin-pwa`  we're adding an application and we want some more space (the right sidebar is there but with empty content):
https://streamable.com/3wjtca

Using the why VitePress in the docs here, the result is this (removed in this PR, just for testing purposes):
https://streamable.com/5uklw5

### DIFF
--- a/docs/config/frontmatter-configs.md
+++ b/docs/config/frontmatter-configs.md
@@ -199,3 +199,18 @@ interface Feature {
   details: string
 }
 ```
+
+## rightsidebar
+
+- Type: `boolean`
+
+This option only take effect when `layout` is set to `page`.
+
+By default, the right aside will be shown, if you want to remove it, configure the option to `false`:
+
+```yaml
+---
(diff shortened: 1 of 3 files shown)


userquin:
In fact, we can use it to remove the right sidebar if we can detect the current page has no entries to be displayed.
Of course, we also need to check if the ads are configured.

docs/config/frontmatter-configs.md
```diff
@@ -199,3 +199,18 @@ interface Feature {
   details: string
 }
 ```
+
+## rightsidebar
+
+- Type: `boolean`
+
+This option only take effect when `layout` is set to `page`.
```
brc-dd, line 207: This should be layout `doc` right?

PR #981, opened 2022-07-13
feat: smooth scrolling

### LINKED ISSUE #979: add smooth scrolling behavior

* Here is a closed <https://github.com/vuejs/vitepress/pull/978>
* But it has some problems, for example
>Open this for example: https://deploy-preview-978--vitepress-docs.netlify.app/guide/getting-started.html. Click on "What's next" on the right hand side. It won't reach to the top at all, you'll have to click that again.
>
>URL hash will also be messed up in certain scenario because we are probably throttling for 100ms, but the scroll animation may take much longer. 

* I don't understand why.
* And its scrolling is inaccurate like you said.
  * Not just "What's next" , all opportunities come up with this question

_No response_

### Additional context

_No response_

### Validations

userquin:
https://streamable.com/477fyz

closes #979

### DIFF
--- a/src/client/theme-default/styles/base.css
+++ b/src/client/theme-default/styles/base.css
@@ -7,6 +7,7 @@
 html {
   line-height: 1.4;
   font-size: 16px;
+  scroll-behavior: smooth;
   -webkit-text-size-adjust: 100%;
 }
 


brc-dd:
https://github.com/vuejs/vitepress/pull/978#issuecomment-1183549569

Even in your video, when you click on an outline item, it doesn't reach the top.

PR #983, opened 2022-07-14
refactor: upgrade docsearch to fix broken typing

meteorlxy:


brc-dd:
Can you check the vite-3 branch. I had fixed these issues there.

PR #856, opened 2022-06-25
feat(build): switch to vite 3 and support clean urls

brc-dd:
* fixes #219
* fixes #324
* fixes #412
* fixes #444
* fixes #476
* fixes #537
* fixes #715
* fixes #848
* fixes #849
* fixes #917
* fixes #1004
* fixes #1016
* fixes #1122
* fixes #1127

Huge thanks to @georges-gomes, @sapphi-red, @antfu, @patak-dev, @JounQin, @titouanmathis for helping and earlier PRs! And also to people @vitejs for building such awesome stuff!

(Do a squash-merge, the history is messed up here.)

---

In the meantime if someone wants to test this out without locally linking, you can do this (auto-publishes at 0 UTC):

```sh
pnpm add -D vitepress@npm:@brc-dd/vitepress@next
```

PR #992, opened 2022-07-14
feat: allow to use custom syntax highlighting themes

azat-io:
Adds the ability to import custom themes for syntax highlighting.

```ts
import { loadTheme } from 'shiki'
import { join } from 'path'

export default async () => {
    let theme = await loadTheme(join(__dirname, './my-theme.json'))
    return defineConfig({
      title: 'VitePress',

      markdown: {
        theme,
      },
     // ...
  })
}
```

Screenshot 2022-07-14 at 11 46 52 PM

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -1,5 +1,5 @@
 import MarkdownIt from 'markdown-it'
-import { Theme } from 'shiki'
+import { Theme, IShikiTheme } from 'shiki'
 import { parseHeader } from '../utils/parseHeader'
 import { highlight } from './plugins/highlight'
 import { slugify } from './plugins/slugify'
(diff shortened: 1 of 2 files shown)


src/node/markdown/markdown.ts
```diff
@@ -19,7 +19,9 @@ import attrs from 'markdown-it-attrs'
 import emoji from 'markdown-it-emoji'
 import toc from 'markdown-it-toc-done-right'
 
-export type ThemeOptions = Theme | { light: Theme; dark: Theme }
+export type ThemeValue = Theme | IShikiTheme
```
brc-dd, line 22: We probably can directly use `IThemeRegistration` here instead of defining `ThemeValue`.

azat-io: Thanks, fixed it!

azat-io:
@brc-dd Should I rebase this PR to this branch? 
https://github.com/vuejs/vitepress/pull/856

brc-dd:
No, keep it for `main`. I'll test this one out as soon as I get some time.

PR #993, opened 2022-07-15
docs: add set `base` in deploying with github action instructions

jesusantguerrero:
Using GitHub Action to deploy GitHub page found that styles weren't loading correctly. I was missing the `base` in my config.
the GitLab instructions have this section clear but for GitHub, it was not mentioned


brc-dd:
That's written on the very top and is highlighted too. I'll add a github example there itself.

Issue #995, opened 2022-07-15
Repeated execution of the router.go method when clicking on the jump link in the Algolia search box


zonemeen:
### Describe the bug

Repeated execution of the go method in `src/client/app/router.ts` and `src/client/theme-default/components/VPAlgoliaSearchBox.vue` when clicking on the jump link in the Algolia search box results in the need to click back key twice to return to the before page.

![click-back- twice-bug](https://user-images.githubusercontent.com/44596995/179189865-c91b93f3-7d3b-4a3f-b29b-b7a728651b49.gif)

### Reproduction

https://vitepress.vuejs.org/

### Expected behavior

Click back key once to return to the before page.

### System Info

```shell
System:
    OS: Windows 10
    CPU: Intel(R) Core(TM) i5-9400 CPU @ 2.90GHz
    Memory: 15.9 GB
  Binaries:
    Node: 16.10.0
    Yarn: 1.22.17
    npm: 7.24.0
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (103.0.5060.114)
    Internet Explorer: 11.0.22000.120
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #996, opened 2022-07-15
fix: click event triggered twice leads to repeated execution of the router.go method

zonemeen:
fix https://github.com/vuejs/vitepress/issues/995

### DIFF
--- a/src/client/theme-default/components/VPAlgoliaSearchBox.vue
+++ b/src/client/theme-default/components/VPAlgoliaSearchBox.vue
@@ -85,14 +85,6 @@ function initialize(userOptions: DefaultTheme.AlgoliaSearchOptions) {
             if (route.path === relativeHit) {
               return
             }
-
-            // if the hits goes to another page, we prevent the native link
-            // behavior to leverage the Router loading feature.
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPAlgoliaSearchBox.vue
```diff
@@ -85,14 +85,6 @@ function initialize(userOptions: DefaultTheme.AlgoliaSearchOptions) {
...
               return
             }
-
-            // if the hits goes to another page, we prevent the native link
-            // behavior to leverage the Router loading feature.
-            if (route.path !== relativeHit) {
-              event.preventDefault()
-            }
-
-            router.go(relativeHit)
```
brc-dd, line 95: If we are removing these then there is no need to provide the `onClick` prop, there is nothing being done there except for this.

If I am correct, then we can remove that safely as this https://github.com/vuejs/vitepress/blob/main/src/client/app/router.ts#L129 part of the code handles all the links.

zonemeen:
Sry, this PR was accidentally closed by me😅

PR #999, opened 2022-07-16
feat: minify html code

azat-io:
Comments not removed


brc-dd:
I don't think we need this change. At best, I suppose this will reduce ~40B (<1ms difference)? I feel that current code is more readable.

PR #1002, opened 2022-07-16
feat: smooth scroll and prefs

userquin:
POC: by default smooth scroll not enabled and we can remove it from the configuration, enabled by default (I need to change the docs).

From the comments here https://github.com/vuejs/vitepress/pull/978#issuecomment-1183549569, it seems there are some caveats with the implementation in all browsers:
- chrome: just works, but scrolling from the right aside links seems to not work; from the left aside seems to work
- firefox: ~~not working; we can add a manual focus using `scrollIntoView` in the right aside, we'll need to include `behavior: 'smooth'` (tested but removed, see safari)~~ check next comment
- safari: `smooth` not supported


userquin:
It seems Firefox is using smooth scroll when the target is close, if the distance exceeds some threshold it will apply normal behavior

kiaking:
Closing due to inactivity. The PR becoming obsolete 😓

PR #1011, opened 2022-07-18
docs: add "head" config to app-configs

postmodernistx:
image

### DIFF
--- a/docs/config/app-configs.md
+++ b/docs/config/app-configs.md
@@ -55,6 +55,35 @@ export default {
 }
 ```
 
+## head
+
+- Type: `HeadConfig[]`
+- Default: `[]`
+
+Additional elements to render in the `<head>` tag in the page HTML.
+
+```ts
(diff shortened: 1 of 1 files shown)


xfq:
Copying my comments in https://github.com/vuejs/vitepress/pull/925#issuecomment-1176951676 :

Maybe we should mention where these elements are added. If the user adds scripts/styles, there may be a big difference between adding scripts and styles before and after VP's own scripts and stylesheets:

```html
<style>
    /* VP Theme fist */
</style>
<script>
    // VP's script fist
</script>

<link rel="stylesheet" href="user.css">
<script>
    // User's script
</script>
```

vs.

```html
<!-- User's style sheet first -->
<link rel="stylesheet" href="user.css"> 
<script>
    // User's script first
</script>

<style>
    /* VP Theme */
</style>
<script>
    // VP's script
</script>
```

PR #1055, opened 2022-07-24
fix(theme): switch to navigator clipboard to copy code

azat-io:
`document.execCommand` is deprecated.


brc-dd:
`navigator.clipboard.writeText` is already inside the try block. That `document.execCommand` thing is there as fallback for non-secure contexts.

PR #1056, opened 2022-07-24
perf(a11y): change copy code span to button

azat-io:


src/client/theme-default/styles/components/vp-doc.css
```diff
@@ -384,23 +384,24 @@
   transition: opacity 0.4s;
 }
 
-.vp-doc [class*='language-']:hover > span.copy {
+.vp-doc [class*='language-']:hover > button.copy,
+.vp-doc [class*='language-'] > button.copy:focus {
```
brc-dd, line 388: This thing will not hide the button even if the contents are copied (as the focus still will be on that button). Also, it messes up the styles.

image

brc-dd: Fixed the first part. For second one needs to fix this: (set opacity 0 when button is focused)

https://github.com/vuejs/vitepress/blob/f75cafc149655694e1223142e87ffd54b68525e9/src/client/theme-default/styles/components/vp-doc.css#L432-L434

brc-dd (approved):

azat-io:
Thanks!
Waiting for a new release!

Issue #984, opened 2022-07-14
Allow links in the homepage-layout features


MajesticString:
### Is your feature request related to a problem? Please describe.

Currently, if the layout of a page is `home`, under `features`, links do not render (both markdown and html).

### Describe the solution you'd like

Allow markdown links in `features`
Example:
```md
---
layout: home
title: Ipsum
features:
    - title: [link](https://google.com/)
      details: [another link](https://google.com/)
---
```

### Describe alternatives you've considered

Alternatively, HTML links could be allowed:
```md
---
layout: home
title: Ipsum
features:
    - title: <a href="https://google.com/">link</a>
      details: <a href="https://google.com/">another link</a>
---
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
Fixing this is simple, we just need to use `v-html` instead of `{{ foo }}` syntax in `VPFeature.vue`. But might need to add styles for links to distinguish them:

![image](https://user-images.githubusercontent.com/40380293/178953027-3c14612f-e327-4f3a-9775-97a1de663ce1.png)

remirobichet:
Is this enhancement under developement ?

Another solution would be to let `{{ foo }}` syntax in `VPFeature.vue` and add extra optional property to feature section.
For example `link` as in the hero section and add a CTA at the end of each feature item.

PR #1070, opened 2022-07-26
feat(theme): link in homepage features

remirobichet:
As discussed in #984, add possibility to have action to homepage feature elements.
Links have same behavior as in hero section.

### DIFF
--- a/docs/guide/theme-home-page.md
+++ b/docs/guide/theme-home-page.md
@@ -124,7 +124,7 @@ interface Feature {
   action?: FeatureAction
 }
 
-interface HeroAction {
+interface FeatureAction {
   // Color theme of the button. Defaults to `brand`.
   theme?: 'brand' | 'alt'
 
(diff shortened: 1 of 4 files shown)


wycats:
It would be great to get this improvement. This change would really improve the experience of building pages with the default `home` layout!

docs/guide/theme-home-page.md
```diff
@@ -115,5 +119,19 @@ interface Feature {
 
   // Details of the feature.
   details: string
+
+  // Action button to display in home feature section.
+  action?: FeatureAction
+}
+
+interface HeroAction {
```
almondheng, line 127: It should be `FeatureAction` instead.

kiaking (approved):

kiaking:
Yeah this is good one and I do have an idea. Please hold on a bit. I think we should only have 1 action which is text link instead of a button being displayed. I'll try to show what I mean by this soon.

remirobichet:
Any news @kiaking ?

PR #1082, opened 2022-07-28
feat(build): support code highlight in uppercase

xinconan:
Support code highlight in uppercase, like Bash、HTML、JavaScript etc can normally build.

```
Error: No language registration JavaScript
```


src/node/markdown/plugins/highlight.ts
```diff
@@ -15,6 +15,7 @@ export async function highlight(theme: ThemeOptions = 'material-palenight') {
   return (str: string, lang: string) => {
     const vPre = vueRE.test(lang) ? '' : 'v-pre'
     lang = lang.replace(vueRE, '')
+    lang = lang.toLowerCase()
```
brc-dd, line 18:
```suggestion
    lang = lang.replace(vueRE, '').toLowerCase()
```

Issue #998, opened 2022-07-16
[discussion] Changing style of code blocks


jd-solanki:
### Is your feature request related to a problem? Please describe.

Hi, When I was checking my libs docs in mobile I found that in mobile there's no `border-radius` and code block hits the border of mobile _(due to negative margin on left & right)_.

![image](https://user-images.githubusercontent.com/47495003/179343899-8a3b1ffe-a972-4b1d-b17e-c3a90d36af81.png)

Later I found that it is intentional and VuePress is also doing the same.

However, I want to point out where it fails.

If we have indented code block then it looks like this:

![image](https://user-images.githubusercontent.com/47495003/179343994-45b01053-cab4-471f-87b7-4ac91557e8f1.png)

```
1. Remove anu plugin use in `main.js` file if you haven't

    ```diff
    -  import { anu } from 'anu-vue'

      createApp(App)
    -   .use(anu)
        .mount('#app')
    ```
```

### Describe the solution you'd like

1. Remove the negative margins
2. Remove breakpoint for adding border-radisu (apply border radius on all screens)

### Describe alternatives you've considered

Revering styles:

```css
.vp-doc div[class*='language-']:not(:where(.vp-raw *)) {
    border-radius: 8px !important;
    margin-right: 0 !important;
    margin-left: 0 !important;
}
```

### Additional context

This is how it looks with the new UI:

![image](https://user-images.githubusercontent.com/47495003/179344046-4b018cea-1374-424d-ba4c-7329367a560a.png)

![image](https://user-images.githubusercontent.com/47495003/179344062-45a07b44-f127-4995-83b0-0d38fbb9b5b4.png)

It aligns perfectly with the content ❤️ 

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
Ah, this issue is also on https://vitepress.vuejs.org/guide/deploying.html

yyx990803:
I don't necessarily consider this a bug - the style looks fine to me, and it separates indented vs. non-indented code blocks. It could be more of a personal preference.

PR #1103, opened 2022-08-02
fix: inconsistency of code block style in mobile

zonemeen:
fix: https://github.com/vuejs/vitepress/issues/998


brc-dd (changes requested):
This needs to be done for elements inside lists only. If we do that for all elements, then there is extra space in right if the block is not indented:

image

kiaking:
Closing due to inactivity.

PR #1132, opened 2022-08-06
fix(build): recreate server on config change

brc-dd:
It's not just limited to markdown, any change in vite/vue options too wasn't affecting unless manually restarted. These changes should restart the server every time the config or its dependencies are changed.

### DIFF
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -21,6 +21,10 @@ export interface MarkdownCompileResult {
   includes: string[]
 }
 
+export function clearCache() {
+  cache.reset()
+}
+
 export async function createMarkdownToVueRenderFn(
   srcDir: string,
   options: MarkdownOptions = {},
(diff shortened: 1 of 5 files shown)


src/node/plugin.ts
```diff
@@ -244,17 +245,23 @@ export async function createVitePressPlugin(
...
+              process.cwd(),
+              file
+            )} changed, restarting server...`
           )
+        )
+        try {
+          clearCache()
+          await recreateServer!()
+        } catch (err) {
+          console.error(err)
```
brc-dd, line 262:
```suggestion
          console.error(c.red(`failed to restart server. error:\n`), err)
```

src/node/markdownToVue.ts
```diff
@@ -21,6 +21,10 @@ export interface MarkdownCompileResult {
   includes: string[]
 }
 
+export function clearCache() {
+  cache.reset()
```
brc-dd, line 25: **TODO:** change this to `cache.clear()` in vite-3 branch. `.reset()` is deprecated in newer versions of `lru-cache`.

PR #1136, opened 2022-08-07
fix: better support for native ESM

### LINKED ISSUE #917: Better support for `type: "module"` from user

The outputs ESM entries are not fully compatible with native ESM actually, take the following as example

https://unpkg.com/browse/vitepress@1.0.0-alpha.4/dist/client/index.js

All relative file paths should be ended with correct `.js` extension, and `/index.js` can not be omitted anymore.

That's being said, we should use `moduleResolution: "Node16"` instead.

And when the user enables `type: "module"` in their `package.json` or `.vitepress/package.json`, all temporary `.js` files in `.vitepress/.temp` would be considered as ESM only, and now it would emit error:

```log
vitepress v1.0.0-alpha.4
- building client + server bundles...
@alauda/chart doesn't appear to be written in CJS, but also doesn't appear to be a valid ES module (i.e. it doesn't have "type": "module" or an .mjs extension for the entry point). Please contact the package author to fix.
quill-image-uploader doesn't appear to be written in CJS, but also doesn't appear to be a valid ES module (i.e. it doesn't have "type": "module" or an .mjs extension for the entry point). Please contact the package author to fix.
✓ building client + server bundles...
- rendering pages...
✖ rendering pages...
build error:
 ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/home/jenkins/agent/workspace/frontend/frontend-vision-docs/docs/.vitepress/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///home/jenkins/agent/workspace/frontend/frontend-vision-docs/docs/.vitepress/.temp/app.js:2:25
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
    at async renderPage (file:///home/jenkins/agent/workspace/frontend/frontend-vision-docs/node_modules/vitepress/dist/node/serve-05342c3d.js:39534:25)
    at async build (file:///home/jenkins/agent/workspace/frontend/frontend-vision-docs/node_modules/vitepress/dist/node/serve-05342c3d.js:39853:9)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
script returned exit code 1
```

Enable `moduleResolution: "Node16"` and add a `package.json` into `.vitepress/.temp` with `type: "commonjs"` for now, and migrate all outputs files as pure ESM later.

We can use `.cjs` files extensions for files outputs in `.vitepress/.temp`

### Additional context

Reference https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing

### Validations

JounQin:
close #917

### DIFF
--- a/client.d.ts
+++ b/client.d.ts
@@ -2,4 +2,4 @@
 // be able to reference vite/client in project root.
 /// <reference types="vite/client" />
 
-export * from './dist/client'
+export * from './dist/client-types/index.js'
--- a/client.d.ts
+++ b/client.d.ts
@@ -2,4 +2,4 @@
 // be able to reference vite/client in project root.
 /// <reference types="vite/client" />
 
-export * from './dist/client-types/index.js'
+export * from './dist/client/index.js'
--- a/src/client/package.json
+++ b/src/client/package.json
@@ -1,3 +1,0 @@
-{
-  "type": "module"
-}
--- a/src/client/package.json
+++ b/src/client/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "module"
+}
--- a/src/client/theme-default/components/VPAlgoliaSearchBox.vue
+++ b/src/client/theme-default/components/VPAlgoliaSearchBox.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import type { DefaultTheme } from 'vitepress/theme'
-import docsearch from '@docsearch/js'
+// FIXME: it's workaround for TypeScript bug
+import { default as docsearch } from '@docsearch/js'
 import { onMounted } from 'vue'
 import { useRouter, useRoute, useData } from 'vitepress'
 
--- a/src/client/app/components/Content.ts
+++ b/src/client/app/components/Content.ts
@@ -1,5 +1,5 @@
 import { defineComponent, h } from 'vue'
-import { useRoute } from '../router'
+import { useRoute } from '../router.js'
 
 export const Content = defineComponent({
   name: 'VitePressContent',
--- a/src/client/app/composables/head.ts
+++ b/src/client/app/composables/head.ts
@@ -1,6 +1,6 @@
 import { watchEffect, Ref } from 'vue'
-import { HeadConfig, SiteData, createTitle, mergeHead } from '../../shared'
-import { Route } from '../router'
+import { HeadConfig, SiteData, createTitle, mergeHead } from '../../shared.js'
+import { Route } from '../router.js'
 
 export function useUpdateHead(route: Route, siteDataByRouteRef: Ref<SiteData>) {
   let managedHeadTags: HTMLElement[] = []
--- a/src/client/app/composables/preFetch.ts
+++ b/src/client/app/composables/preFetch.ts
@@ -1,9 +1,9 @@
 // Customized pre-fetch for page chunks based on
 // https://github.com/GoogleChromeLabs/quicklink
 
-import { useRoute } from '../router'
+import { useRoute } from '../router.js'
 import { onMounted, onUnmounted, watch } from 'vue'
-import { inBrowser, pathToFile } from '../utils'
+import { inBrowser, pathToFile } from '../utils.js'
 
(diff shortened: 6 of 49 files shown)


src/client/theme-default/components/VPAlgoliaSearchBox.vue
```diff
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import type { DefaultTheme } from 'vitepress/theme'
-import docsearch from '@docsearch/js'
+// FIXME: it's workaround for TypeScript bug
+import { default as docsearch } from '@docsearch/js'
```
JounQin, line 4: It should be a TypeScript bug, I'll raise an issue on its repository.

src/client/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
JounQin, line 2: This is required to make sure it will always work to make `.js` considered as ESM, whatever the user's package type.

client.d.ts
```diff
@@ -2,4 +2,4 @@
 // be able to reference vite/client in project root.
 /// <reference types="vite/client" />
 
-export * from './dist/client'
+export * from './dist/client-types/index.js'
```
brc-dd, line 5: dist/client-types is deleted after bundling

JounQin: @brc-dd Done!

src/client/package.json
brc-dd: We actually don't need this here, this is auto generated by our vite plugin while building.

JounQin: image

I didn't see it when I remove this and run `pnpm run docs`, is there anything I missed?

brc-dd: I meant when you build the site, it will be generated in the .temp directory. Why do we need this in dist/client?

JounQin: > This is required to make sure it will always work to make .js considered as ESM in temporary client directory, no matter whatever the user's package type.

brc-dd: Yeah I read that 😅, but its still not clear. Any particular example where not having this will break stuff?

JounQin: When the user is not using `type: module` or using `type: commonjs`, all `.js` will be treated as commonjs by default, unless there is a closer `package.json` declaring `type: module` all `.js` should be considered as ESM.

```plan
- docs
  - .vitepress
    - .temp
      pacage.json # `type: "module"` is required
- package.json # `commonjs` will also affects `docs/.vitepress/.temp` (I forgot the real path here, the depth doesn't matter)
```

brc-dd: Yeah inside that .temp we are emitting this file already: https://github.com/vuejs/vitepress/blob/chore/vite-3/src/node/plugin.ts#L227

JounQin: > Yeah inside that .temp we are emitting this file already: [https://github.com/vuejs/vitepress/blob/chore/vite-3/src/node/plugin.ts#L227](https://github.com/vuejs/vitepress/blob/chore/vite-3/src/node/plugin.ts?rgh-link-date=2022-08-07T06%3A44%3A20Z#L227)

Thanks for pointing it out, I'm not so familiar with the code base. 🤣

brc-dd: It will emit it by default, unless user has set ssr format in vite.config to cjs for some reasons. In that case we'll respect that.

JounQin: @brc-dd 

OK, let's see if it will cause some issues later. As I imagine, all `.js` in `.temp` are ESM now, so setting is as cjs is meaningless and won't work.

brc-dd: All .js in temp are actually generated by vite, so if user has specified cjs there, then they would be cjs.

JounQin: Thanks for pointing!

PR #1163, opened 2022-08-14
fix(theme): fix custom Not found page rendering

Roman3349:
Signed-off-by: Roman Ondráček <ondracek.roman@centrum.cz>


brc-dd:
What was the issue?

Roman3349:
@brc-dd If you are using the default theme with custom Not found page, without this fix the page is rendered in `doc` layout.

**Without fix**
without-fix

**With fix**
with-fix

PR #1166, opened 2022-08-14
test: fix failing test due to timeout

brc-dd:


Roman3349:
@brc-dd Test are failing due ` Uncaught ReferenceError: can't access lexical declaration 'Theme' before initialization`. The error is caused by your change in commit https://github.com/vuejs/vitepress/pull/1163/commits/f808723bb13ce86b1b5d70400ac9e7295ec02072.

brc-dd:
Thanks, I thought directly exporting would have worked. 🤔

PR #1168, opened 2022-08-15
fix(theme): don't let navbar obstruct clicks to top part of scrollbar

curtgrimes:
On desktop screen sizes, if the sidebar has a scrollbar, the very top
part of the scrollbar can't be clicked on because the navbar obstructs
it due to having a higher z-index. (This higher z-index is necessary so
that the navbar is visible.) With this fix, the navbar container has
`pointer-events: none;` so that clicks to the scrollbar underneath will
work. Child elements of the navbar that need to receive clicks have
`pointer-events: all;` turned on for them.

fix vueuse/vueuse#2086

## Before

The red highlighted part of the scrollbar could not be clicked and dragged:

image

## After

The part of the scrollbar above can now be clicked and dragged, and all of the other parts of the sidebar and navbar remain clickable as before.

### DIFF
--- a/src/client/theme-default/components/VPNav.vue
+++ b/src/client/theme-default/components/VPNav.vue
@@ -33,6 +33,7 @@ provide('close-screen', closeScreen)
   left: 0;
   z-index: var(--vp-z-index-nav);
   width: 100%;
+  pointer-events: none;
 }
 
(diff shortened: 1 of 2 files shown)


brc-dd:
Thanks!

Issue #1169, opened 2022-08-15
UI inconsistencies with code blocks


brc-dd:
_Originally posted by @brc-dd in https://github.com/vuejs/vitepress/issues/1152#issuecomment-1214753422_

PR #1152, opened 2022-08-09
Replace the line highlighting logic with Shiki lineOptions

Jinjiang:
Re-implemented line highlighting logic according to @octref's approach https://github.com/vuejs/theme/issues/41

Changes:

- `src/node/markdown/plugins/highlightLines.ts`: only normalize the line info and append it back to `token.info`
- `src/node/markdown/plugins/highlight.ts`: parse the normalized line info into `lineOptions` and pass it into shiki.
- `src/client/theme-default/styles/components/vp-doc.css`: style update accordingly to code blocks

Thanks.

PR #1170, opened 2022-08-15
fix(theme): restore styles for code blocks

### LINKED ISSUE #1175: New lines are ignored in markdown code snippets in 1.0.0-alpha.5
### Describe the bug

In Makrdown code snippets new lines are now being ignored. You need to manually add `<space>` with a newline to render it in the docs

### Reproduction

Create a code snippet which has new lines

### Expected behavior

Should render new lines without additional steps

### System Info

```shell
  System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz 
    Memory: 6.91 GB / 15.96 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE       
    Yarn: 1.22.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.81
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.54)
  npmPackages:
    vitepress: ^1.0.0-alpha.5 => 1.0.0-alpha.5
```

### Additional context

_No response_

### Validations

brc-dd:
fixes: #1169, fixes #1175

x-ref: #1152

### DIFF
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -310,23 +310,23 @@
 }
 
 .vp-doc [class*='language-'] code {
-  display: flex;
-  width: max-content;
+  display: block;
+  padding: 0 24px;
+  width: fit-content;
   min-width: 100%;
-  flex-direction: column;
   line-height: var(--vp-code-line-height);
   font-size: var(--vp-code-font-size);
   color: var(--vp-code-block-color);
   transition: color 0.5s;
 }
 
-.vp-doc [class*='language-'] code .line {
-  padding: 0 24px;
-}
-
 .vp-doc [class*='language-'] code .highlighted {
   background-color: var(--vp-code-line-highlight-color);
   transition: background-color 0.5s;
+  margin: 0 -24px;
+  padding: 0 24px;
+  width: calc(100% + 2 * 24px);
+  display: inline-block;
 }
 
 .vp-doc div[class*='language-'].line-numbers-mode {


brc-dd:
@Jinjiang Hi! Can you check if this doesn't causes any other UI issues? I have restored the styles to what we had before #1152 and modified them for `.highlighted`.

Jinjiang:
@brc-dd my bad 😅 thanks for pointing this out.

After trying a little bit, I found another simple way to solve this is having a outer padding + inner negative-margin:

```diff
diff --git a/src/client/theme-default/styles/components/vp-doc.css b/src/client/theme-default/styles/components/vp-doc.css
index 84e9d08..1038045 100644
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -312,6 +312,7 @@
 .vp-doc [class*='language-'] code {
   display: flex;
   width: max-content;
+  padding: 0 24px;
   min-width: 100%;
   flex-direction: column;
   line-height: var(--vp-code-line-height);
@@ -321,6 +322,7 @@
 }
 
 .vp-doc [class*='language-'] code .line {
+  margin: 0 -24px;
   padding: 0 24px;
 }
```

Just FYI. I'm still checking other use cases. 🙏

brc-dd:
> After trying a little bit, I found another simple way to solve this is having a outer padding + inner negative margin:

But with keeping display flex on code, we can't have empty spans in it, or can we? (newlines were getting ignored earlier too)

src/client/theme-default/styles/components/vp-doc.css
```diff
@@ -310,23 +310,23 @@
...
 }
 
-.vp-doc [class*='language-'] code .line {
-  padding: 0 24px;
-}
-
 .vp-doc [class*='language-'] code .highlighted {
   background-color: var(--vp-code-line-highlight-color);
   transition: background-color 0.5s;
+  margin: 0 -24px;
```
Jinjiang, line 326: Here, I prefer applying these layout properties on `.line` instead of `.highlighted` to keep the layout of each line consistent. And the `display` can be `block` since it actually fills the whole line, not an inline element anymore. WDYT?

brc-dd: Putting it on .line directly was causing some issue like extra blank line at end IG. Regarding that block thing, I'll check it.

Jinjiang: I see. Then I think so far `inline-block` is the best choice, whatever it's on `.line` or `.highighted`. No problem from my side. 👍

Jinjiang: weird...

Take the `[[toc]]` for example. The generated code has 2 lines actually, which causes the extra blank line as you mentioned.

image

Anyway, let's take this.

Jinjiang (approved):

src/client/theme-default/styles/components/vp-doc.css
brc-dd: Yeah we had discussed this stuff internally too before alpha. Then it was happening only on Firefox. Ideally, that last line should not be there IMO. I'll check if we can fix that from the plugin itself.

brc-dd: Okay, so removing that extra line has not done simplified anything. Because it's actually not guaranteed to have a `.line` or even a span inside `code`. Gonna go ahead with these changes themselves.

PR #1177, opened 2022-08-16
add base url for favicon

botue:
There is no patched base path when adding favicon


brc-dd:
Users should themselves be doing this. Also, this won't work if someone has absolute, scheme relative or page relative href.

botue:
ok, I see

PR #1180, opened 2022-08-16
add indent for sidebar link

botue:


kiaking:
Hi, thanks for the PR! I understand you prefer it to be intended, but sorry, no indention is intentional 👀

PR #1184, opened 2022-08-17
add a plugin for wrapped table

botue:
support scrolling table in small screen


kiaking:
Thanks for the PR but not sure how badly we want this. Because it's already scrollable right now.

Using table with responsive design is quite tough one. Like whether the content value should be word-wrapped or not, how about freezing some rows or columns and such.

If user needs such powerful table, I think they're better off creating custom Vue Component for that.

And for those who just need a simple table, I think current table is enough.

So I'll be closing this for now, but thank you for the PR!

botue:
ok

PR #1216, opened 2022-08-20
feat(client): export VPSocialLinks

Sepush:
Signed-off-by: Sepush <sepush@outlook.com>

need this component to display sociallinks of author in blog page some workaround in https://github.com/vuejs/vitepress/discussions/1123 but seems broken in alpha8
export it and we can reuse it will be convenient
eg:
image


brc-dd:
Please refer the discussion. We aren't gonna expose it as it likely won't work as you want in future releases.

PR #1218, opened 2022-08-20
feat: support markdown frontmatter options

meteorlxy:
- use @mdit-vue/plugin-frontmatter

### DIFF
--- a/src/node/markdown/plugins/link.ts
+++ b/src/node/markdown/plugins/link.ts
@@ -3,7 +3,8 @@
 // 2. normalize internal links to end with `.html`
 
 import MarkdownIt from 'markdown-it'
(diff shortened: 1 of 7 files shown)


src/node/markdown/plugins/link.ts
```diff
@@ -37,7 +44,7 @@ export const linkPlugin = (
         // links to files (other than html/md)
         !/\.(?!html|md)\w+($|\?)/i.test(url)
       ) {
-        normalizeHref(hrefAttr, env.cleanUrl)
+        normalizeHref(hrefAttr, env.cleanUrls)
```
meteorlxy, line 47: It looked like a typo here, so I corrected it by the way @brc-dd

PR #1226, opened 2022-08-23
chore: fix example

meedfine:
Compared to https://vitepress.vuejs.org/, have extra comma

https://vitepress.vuejs.org/
image

example
image


kiaking (approved):
Thanks!

PR #1230, opened 2022-08-23
refactor: externalize markdown headers plugin

meteorlxy:


src/client/theme-default/components/VPDocAsideOutline.vue
```diff
@@ -45,16 +40,15 @@ function handleClick({ target: el }: Event) {
 
         <ul class="root">
           <li
-            v-for="{ text, link, children, hidden } in resolvedHeaders"
-            v-show="!hidden"
```
meteorlxy, line 49: Currently the `hidden` property is totally redundant because it would always be `undefined`, so I removed it for simplicity

brc-dd:
Regarding this header stuff, do we actually need it? Because we're just using it for the aside outline, but using headers extracted from markdown is preventing users from using dynamic headers (Vue-style interpolations or added using some component). In future, we'll probably need to make that component client-only so that it can handle such headings.

src/client/theme-default/components/VPDocAsideOutline.vue
```diff
@@ -45,16 +40,15 @@ function handleClick({ target: el }: Event) {
...
+            v-for="{ title, link, children } in page.headers"
```
brc-dd, line 43: Just a heads up, @vuejs/theme would also need to update this (`text` -> `title`).

meteorlxy: Yeah, but not necessary. Here we changed it to `title` because we don't need an extra `resolvedHeaders` transform. The type of `Header` is still compatible.

types/shared.d.ts
```diff
@@ -12,12 +12,7 @@ export interface PageData {
   lastUpdated?: number
 }
 
-export interface Header {
-  level: number
-  title: string
```
meteorlxy, line 17: @brd-cc Check here, the original property is `title`. It is transformed to `text` in the `mapHeaders` method. :eyes:

brc-dd (approved):

PR #1235, opened 2022-08-24
refactor: use markdown title plugin and remove parseHeader utils

meteorlxy:
In this PR, the `parseHeader` / `deeplyParseHeader` utils are totally removed. We don't need to maintain the complex regexp-based parsing anymore.

Instead, we use markdown-it to resolve the title, or parse the `frontmatter.title`. It is safer and more accurate, and has better compatibility with other markdown-it plugins. For example, we don't need extra process for `markdown-it-attrs` as before.


brc-dd:
Just to be sure, can you add tests for `resolveTitleFromToken` in the mdit-vue repo?

src/node/markdownToVue.ts
```diff
@@ -243,18 +244,21 @@ function genPageDataCode(tags: string[], data: PageData, replaceRegex: RegExp) {
...
-  const match = content.match(/^\s*#+\s+(.*)/m)
-
-  if (match) {
-    return deeplyParseHeader(match[1].trim())
+const inferTitle = (
+  md: MarkdownRenderer,
+  frontmatter: Record<string, any>,
+  title: string
+) => {
+  if (typeof frontmatter.title === 'string') {
```
brc-dd, line 252: should it be !== here?

src/node/markdownToVue.ts
```diff
@@ -243,18 +244,21 @@ function genPageDataCode(tags: string[], data: PageData, replaceRegex: RegExp) {
...
+    const titleToken = md.parseInline(frontmatter.title, {})[0]
```
brc-dd, line 253: Why are we parsing frontmatter.title? Do people actually put random markdown/html syntax there?

meteorlxy: It should be migrated from vuepress v1, or comes from some feature request?

As you can see, we were using `deeplyParseHeader(frontmatter.title)` before.

meteorlxy: Yes it's from vuepress v1: https://github.com/vuejs/vuepress/blob/d692cc852a26ed343f26a7bdd9d336ad3b0da191/packages/%40vuepress/shared-utils/src/inferTitle.ts#L11-L22

In fact, I didn't migrate it to vuepress v2, because I don't think people may expect markdown syntax to be worked in frontmatter (I think you might have similar thoughts lol). The only thing might be the emoji syntax `:foo:`

__tests__/node/utils/deeplyParseHeader.spec.ts
```diff
@@ -1,34 +0,0 @@
-import { test, expect } from 'vitest'
-import { deeplyParseHeader } from 'node/utils/parseHeader'
-
-test('deeplyParseHeader', () => {
-  const asserts: Record<string, string> = {
```
brc-dd, line 5: You probably can copy these test cases (without leading # and spaces in outcome) to https://github.com/mdit-vue/mdit-vue/blob/main/packages/plugin-title/tests/title-plugin.spec.ts

src/node/markdownToVue.ts
brc-dd: Ah fine, let's keep it this way. 😅

brc-dd (approved):

__tests__/node/utils/deeplyParseHeader.spec.ts
meteorlxy: Yeah we could. But IMO we don't need to worry about them because all the markdown syntax has already been parsed correctly by markdown-it 😄 We can add some cases once when found any unexpected behaviors.

PR #1240, opened 2022-08-26
refactor: remove non-standard markdown-it extensions

meteorlxy:


brc-dd (approved):

yyx990803:
This broke current vuejs.org builds!

Please do not make risky changes like this in the future.

meteorlxy:
@yyx990803 Sorry for that. Using `env` of markdown-it should be a more recommended way for passing md render context.

@brc-dd We should also review the usage in vue/theme, too.

meteorlxy:
Should we create a PR for https://github.com/vuejs/docs repo? I saw the legacy usage in that repo. @yyx990803

Issue #728, opened 2022-06-06
Support code groups and allow adding title to code blocks


jd-solanki:
### Is your feature request related to a problem? Please describe.

VueJS community is moving from JS to TS now. Hence, we have to provide both TS & JS code.

I already started using VitePress alpha in my OSS project but for another project, I prefer providing TS & JS code via code groups like VuePress.

### Describe the solution you'd like

Provide code blocks via markdown (using component is bit ugly for markdown only content). We can inspire from Nuxt content 2 for new syntax

<pre>
:::code group

:::group{ts}

```ts
const a: number = 1
```

:::

:::group{js}

```js
const a = 1
```

:::

:::
</pre>

### Describe alternatives you've considered

None

### Additional context

https://content.nuxtjs.org/guide/writing/mdc

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

kecrily:
Maybe it could be simpler?

~~~markdown
:::code group
```ts
const a: number = 1
```

```js
const a = 1
```
:::
~~~

brc-dd:
@kecrily I guess the OP meant to have title of each group too. `:::group{foo}` --> title is `foo`.

kecrily:
We can custom group title in code block 

~~~markdown
```ts [TypeScript]
const a: number = 1
```
~~~

Zhengqbbb:
Looking forward this feature, which will have a different experience with such code experience:
![image](https://user-images.githubusercontent.com/40693636/172436242-f645ee81-083a-4c68-a5eb-90b94e4fc2c8.png)

marshallswain:
Can we expose the markdown-it rendered to be used inside of Tab components?  It would be really nice to be able to use Vue components to render tab groups:

```
<Tabs>
  <Tab name="TypeScript">
    ``ts
    nested fenced code block here.
    ``
  </Tab>
  <Tab name="JavaScript">
  ``js
    nested fenced code block here.
  ``
  </Tab>
</Tabs>
```

brc-dd:
[...]

PR #1242, opened 2022-08-26
feat: Support code group in Markdown

VoVAllen:
Signed-off-by: Jinjing.Zhou <allenzhou@tensorchord.ai>

Ref https://github.com/vuejs/vitepress/issues/728

I made a new page to test it and found I still need to manually import the SFC. How can I enable this SFC by default? @brc-dd 

Change: Originally the upper-right side will show the code language. Now it will show title specified by user, if not will be the language itself.

### DIFF
--- a/docs/guide/test.md
+++ b/docs/guide/test.md
@@ -1,0 +1,51 @@
+<script setup>
+  import { VPCodeGroup } from "vitepress/theme";
+</script>
+
+# Code Groups
+
+test
+
+:::code-group
+
+```js [app.vue]
+<template>
+  <NuxtLayout>
+    <NuxtPage />
+  </NuxtLayout>
+</template>
+```
+
+```js{3,4}  [layouts/custom.vue]
+<template>
+  <div>
+    Some *custom* layout
+    <slot />
+  </div>
+</template>
+```
+
+```js{1-3,5} [layouts/default.vue]
+export default {
+  name: 'MyComponent'
+  // ...
+}
+<template>
+  <div>
+    Some *custom* layout
+    <slot />
+  </div>
+</template>
+```
+
+```js
+<template>
+  <NuxtLayout>
+    <NuxtPage />
+  </NuxtLayout>
+</template>
+```
+
+:::
+
+
--- a/src/client/theme-default/components/VPCodeGroup.vue
+++ b/src/client/theme-default/components/VPCodeGroup.vue
@@ -1,0 +1,89 @@
+<script setup lang="ts">
+import { onMounted, ref, watch } from 'vue';
+
+interface Tab {
+    label: string;
+    elm: Element;
+}
+
+const codeblocks = ref<HTMLElement | null>(null)
+let tabs = ref<Array<Tab>>([])
+const activeTabIndex = ref(0)
+
+onMounted(() => {
+    if (codeblocks !== undefined) {
+        codeblocks.value?.querySelectorAll(".code-block").forEach((element, i) => {
+            const title = element.querySelector("span.code-title")?.textContent!
+            tabs.value.push({
+                label: title,
+                elm: element
+            })
+        })
+        switchTab(0)
+    }
+})
+
+function switchTab(i: number) {
+    tabs.value.forEach((tab) => {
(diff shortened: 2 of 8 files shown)


src/client/theme-default/components/VPCodeGroup.vue
```diff
@@ -0,0 +1,89 @@
...
+    outline: none;
+}
+
+.tabs-header button:hover {
+    color: rgba(255, 255, 255, 0.82);
+}
+
+
+.tabs-header button.active {
+    background-color: rgb(63 63 70);
```
Sepush, line 73: `background-color: var(--vp-c-gray);`
maybe better

Sepush:
this feature pretty cool

Sepush:
> Signed-off-by: Jinjing.Zhou [allenzhou@tensorchord.ai](mailto:allenzhou@tensorchord.ai)
> 
> Ref #728
> 
> I made a new page to test it and found I still need to manually import the SFC. How can I enable this SFC by default? @brc-dd
> 
> Change: Originally the upper-right side will show the code language. Now it will show title specified by user, if not will be the language itself.
> 
> demo: https://deploy-preview-1242--vitepress-docs.netlify.app/guide/test

you can register as a global component

docs/guide/test.md
```diff
@@ -0,0 +1,51 @@
+<script setup>
```
Sepush, line 1: can export in `src/client/theme-default/index.ts`
and register in `src/client/app/index.ts` as a global component
so that we don't need import in every page

rainbowatcher:
it's looks like doesn't work well on mobile

VoVAllen:
Current color pattern doesn't seems good. Any suggestion is welcomed

meteorlxy:
As code-group is a theme-specific feature, I wonder if we should put it into the core, including:

- The markdown container plugin
- The markdown preWrapper plugin

VoVAllen:
@meteorlxy The prewrapper change should work for all themes. Is it fine to keep it for all themes?

For the container part, does it break anything if that is not implemented by other themes? (transform `::code-group` in to `<CodeGroup></CodeGroup>`)

VoVAllen:
Ping @meteorlxy @brc-dd

brc-dd:
> For the container part, does it break anything if that is not implemented by other themes? (transform `::code-group` in to `<CodeGroup></CodeGroup>`)

Not unless they are using `::code-group` in their markdown files, in that case they would need to register a global component `CodeGroup`.

Just a thought - this likely can be done in pure HTML/CSS without need of any Vue component. (make md plugin render something like radio input + pre, then use :checked). I'll give that a try, otherwise we'd go ahead with this PR.

VoVAllen:
@brc-dd I see what you mean (like https://codepen.io/xboxyan/pen/oMxmxm). It seems viable in this scenario.

brc-dd:
> Seems dynamic CSS generation is needed here.

Not quite. In your example too the CSS is static (its independent of those tab IDs).

> Is this a good option?

Perhaps. I'm not sure yet. This would prevent the need of defining a custom component, so this will be helpful to developers writing a custom theme. It would sort of become a markdown feature instead of theme one. Also, there are already certain markdown-it plugins that do that, so it wouldn't be hard to implement. My only concern is with a11y. That would need investigation.

VoVAllen:
Actually I think we can avoid using SFC, but doesn't have to be pure CSS. We can still include js usage for simplicity. WDYT? @brc-dd

brc-dd:
Yeah that would work too. We are injecting JS for copy code stuff too so that it works across custom themes.

VoVAllen:
I've updated the code to implement this without SFC. Please take a look @brc-dd . Thanks!

VoVAllen:
@brc-dd Does current implementation work? If so I can fix the docs so that we can merge it

VoVAllen:
Also suggestion on color pattern is welcomed

brc-dd:
Can you guys review #1560 and see if there are any issues (especially with UI)? Here is the deploy preview: https://deploy-preview-1560--vitepress-docs.netlify.app/test

PR #930, opened 2022-07-07
feat: support mermaid

7nohe:
It would be nice to support mermaid in markdown files.
The idea is based on [this issue](https://github.com/vuejs/vitepress/issues/423#issuecomment-1134585626).

You can add diagrams to your markdown file as follows.

```
::: mermaid
sequenceDiagram
User->>Server: Request page
::: 
```

PR #1249, opened 2022-08-29
feat: support mermaid (for alpha)

7nohe:
Replaces #930 for alpha

You can draw diagrams by using [Mermaid](https://mermaid-js.github.io/).

**Input**

```md
::: mermaid
sequenceDiagram
User->>Server: Request page
::: 
```

**Output**

```mermaid
sequenceDiagram
  User->>Server: Request page
```


brc-dd:
There are a some document is not defined errors on the console while building (probably need to initialize mermaid onMounted). Also, I have another concern about this PR - bundling mermaid will drastically increase the bundle size. (Their own site is not using Mermaid AFAICT, they have used cli-generated pngs everywhere IG.)

7nohe:
@brc-dd
I replaced watchEffect with onMounted in Mermaid component. It worked, thanks!
For the issue of increasing bundle size, can this be solved by dynamic import of mermaid?

brc-dd:
> For the issue of increasing bundle size, can this be solved by importing mermaid dynamically?

Most likely yes. Currently while building you'll be seeing some large chunk warnings on terminal. They shouldn't be there.

7nohe:
Yeah, I saw the following warnings.

> (!) Some chunks are larger than 500 KiB after minification. Consider:
> - Using dynamic import() to code-split the application
> - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/guide/en/#outputmanualchunks
> - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.

I will try the above solutions.

7nohe:
@brc-dd Now, loading mermaid is disabled by default. When disabled, the warnings also disappears.
It can be enabled by setting the `markdown.mermaid` option to true in .vitepress/config.ts.

src/client/app/components/Mermaid.vue
```diff
@@ -0,0 +1,32 @@
...
+  src: string
+}>()
+
+const svg = ref<string | null>(null)
+const route = useRoute()
+
+onMounted(() => {
+  mermaid.initialize({
+    startOnLoad: false,
+    theme: 'dark',
```
brc-dd, line 16: This needs some change as currently the theme is dark on light mode too:

image

Emerson is using a mutation observer for this: https://github.com/emersonbottero/vitepress-plugin-mermaid/blob/main/src/Mermaid.vue, but IG this can be much simplified if we somehow expose something like isDark.

Also, slidev has done some cool optimizations that we can include too: https://github.com/slidevjs/slidev/blob/main/packages/client/builtin/Mermaid.vue

I'll let you know once I manage to expose isDark (current logic resides in https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/components/VPSwitchAppearance.vue).

7nohe: Thanks. I will check the above code and wait for the implementation to expose isDark.

emersonbottero:
Oh.. If I knew it I would make an PR not a plugin!😁

something you probably is missing.. 
In my plugin there is 3 features that I really like.

you use fences ``` for default charts
you can use ```mmd to show the code instead of the chart.
you can use images in the charts!
the isDark should be updated when changed so it is not only onMounted.

I can help improve this one.. just let me know.
But I really thought it should be an plugin reading the orientations in the docs.

7nohe:
[vitepress-plugin-mermaid](https://github.com/emersonbottero/vitepress-plugin-mermaid) looks good, and I agree with @kiaking's opinion.
I just created this PR because I thought mermaid is common, adopted by major services like Github, etc., so I'm closing the PR.

PR #1257, opened 2022-08-30
feat: bundle type definitions of dev-deps

### LINKED ISSUE #1254: Bundle types of dev-deps properly
Context: RN people need to manually add dev deps to get their types. This includes (but is not limited to) types of all markdown-related packages.

Expected behavior: If we are exporting some type, it should ideally be standalone. **So users don't need to install other packages for proper type checking.**

Solution: `respectExternal` in `rollup-plugin-dts`?

~~Downfalls: It will likely increase package size of VitePress. But we need to bundle types of only those of our dev-deps which are referenced from our exported APIs and types.~~ Edit: There is only ~0.1MB increase in unpacked size, so that's bearable IG.

x-ref: https://github.com/vuejs/theme/pull/70#issuecomment-1231159544

/cc @meteorlxy

brc-dd:
fixes #1254


src/node/markdown/env.ts
```diff
@@ -1,7 +1,31 @@
-import type { MarkdownItEnv } from '@mdit-vue/types'
-import { CleanUrlsMode } from '../shared'
+import type { MarkdownItHeader } from '@mdit-vue/plugin-headers'
+import type { MarkdownSfcBlocks } from '@mdit-vue/plugin-sfc'
+import type { CleanUrlsMode } from '../shared'
 
-export interface MarkdownEnv extends MarkdownItEnv {
```
meteorlxy, line 4: Let's add comments for why we re-declared those properties

meteorlxy (approved):

brc-dd:
There are some more changes. We need to manually define `Header` because shared.d.ts is not processed by rollup (hence, types from plugin-header won't be bundled). Also, shared.d.ts was getting bundled (and duplicated) in client and node types, so externalized that. Also, stuff like this can now again be done without any issues:

```ts
declare module 'vitepress' {
  interface Header {
    compositionOnly?: boolean
    optionsOnly?: boolean
  }
}
```

meteorlxy (approved):
LGTM

Issue #1271, opened 2022-09-01
Nav items overflowing navbar on medium viewports


fvsch:
### Describe the bug

When using the default theme, the navbar items are hidden away in a menu on screen widths lower than 768px.

At 768px and slightly larger sizes, the navbar may have to render on one line:

- The logo and site title
- A search box (when using the `algolia` options)
- Several nav items
- And the "extra navigation" button

When the nav items don't fit, their text may wrap, and their extreme `line-height` makes that result particularly bad:

<img width="862" alt="Screenshot 2022-09-01 at 23 57 42" src="https://user-images.githubusercontent.com/243601/188020460-7355ef38-ea1e-4d18-b68d-cf7052cb4f9b.png">

### Reproduction

1. Visit reproduction on StackBlitz: https://stackblitz.com/edit/vitepress-theme-issue-navbarmenulink?file=index.md
2. Resize the preview area to be roughtly 800px wide

### Expected behavior

There are two levels of fixes:

1. Mitigate the breakage when nav items do wrap.
    - Nav links should not use the header height as their `line-height` (that's a bad way to do vertical centering of text, for the reason demonstrated in this issue).
    - Instead use something like `min-height: var(--vp-nav-height-mobile); line-height: normal;` (plus probably a tad of `padding-block`).

2. If the nav items still overflow their container, what should happen? Should they be clipped (`overflow: hidden` on the container), or should the container height grow (i.e. the container should be using `min-height` and not `height`)? Clipping might be better to conserve the design, but note that it's a WCAG violation (loss of content).

3. Prevent this situation altogether:
    - One option is to use the mobile layout earlier, e.g. for widths lower than 1000px instead of widths lower than 769px. (This might not be enough for some accessibility use cases, like text-only zoom.)
    - Another one is to implement a type of dynamic overflow menu, where all items that don't fit in the navbar are moved to the overflow menu. This requires runtime JS, probably using IntersectionObserver.

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1
    Memory: 98.88 MB / 8.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.17.0 - /opt/homebrew/opt/node@16/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.12.1 - ~/.npm-packages/bin/npm
  Browsers:
    Chrome: 105.0.5195.52
[...]

PR #1273, opened 2022-09-02
fix(theme): limit damage of nav items overflowing navbar

fvsch:
This is a partial fix for #1271, using CSS only:

- Prefer `min-height` over `line-height` to set the height of navbar items. This means that if an item's text wraps on 2 lines, it will stay within the overall height of the navbar instead of becoming twice that height.

- Prefer `min-height` over `height` to set the height of the navbar. This means that if nav items do become higher than the specified height, the navbar height will expand instead of making the items overflow. Note: this is done on small and medium layouts only (<960px), because on larger screens (>=960px) the navbar is fixed and only a specific height is reserved in the sidebar and in the main content.

The navbar height change should work for most sites, but if a site has injected custom content in the navbar and that content happens to have a height higher than the `--vp-nav-height-mobile` value (e.g. that content is 80px tall, with some of that tallness being transparent content that was not noticed so far), this fix would make the navbar increase its height. That is, it would reveal an existing problem.

As far as I know, letting the navbar height expand instead of letting the content overflow (and end up on top of page content, obscuring that content) is necessary to pass WCAG 2.2's [Success Criterion 1.4.4 Resize text (AA)](https://www.w3.org/TR/WCAG22/#resize-text).

### Screenshots

I made a stress case with a long `siteTitle` and long labels for `nav` items in the `themeConfig`. On large window widths (with no global or text zoom) everything is alright:

Screenshot of a Vuepress site's navbar at 1360px width, with all nav items fitting in a single line. There is a little room left between the search button and the first nav item.

Lowering the viewport width to 1150px, nav item labels can't fit one a single line and start wrapping. Whether that works well or not depends on the labels you're using and whether they contain line break opportunities (whitespace, hyphens in all browsers, and some other characters like `/` in some browsers). The result is not very beautiful, but much better than without this PR.

Screenshot of a Vuepress site's navbar at 1150px, with all nav item labels wrapping to two lines. The line-height for nav items is low enough that nav items still fit in the same vertical space as when there is no text wrapping.
 
We can go down to 960px, and in this particular example it still works, but barely. If we had a longer word in some label, or an extra nav item, then the navbar's menu could not be compressed further and it would start pushing the rest of the content out of the viewport on the right side.

Now we go down to 950px, triggering the "medium" layout (`768px >= width < 960px`). In this layout, if you have more than a couple nav items, plus a search button, plus a site name that is not super short… it's almost guaranted that content won't fit. Which is why this PR is not a full solution, and it would be better to also look at a solution for this specific layout. I have a few ideas in mind, but it's beyond the scope of this PR; let's focus on damage control. ^^

So, damage control: here we have one nav item that wraps on 4 lines! And the result is not beautiful, and looks broken-ish, but all the content is still accessible so that's a win.

Screenshot of a Vuepress site's navbar at 950px wide, where the long site title's on the left takes a lot of room and navigation items are compressed, their text wrapping to two, three or even 4 lines. The navbar's height expanded to contain all the navigation items.

At lower sizes like 768px, things get a bit harder. I'm not sure the default theme can effectively show a site title, a search button, some custom button and 2-3 navigation items all in one line at this size, even without pushing the limits like in this demo. Still, damage is partially controlled (not completely, content still overflows on the right).

Screenshot of a Vuepress site's navbar at 768px wide, where there is just not enough room to fit all content and a couple buttons have disappeared to the right, while a third is partially cropped.

Finally, at lower phone-like widths, things should be unchanged:

Screenshot of a Vuepress site's navbar at 600px wide, showing the site title, an icon-only search button and an icon-only menu button. All content fits in the available space.


src/client/theme-default/components/VPFlyout.vue
```diff
@@ -90,7 +90,7 @@ function onBlur() {
 .button {
   display: flex;
   align-items: center;
-  padding: 0 12px;
+  padding: 4px 12px;
```
fvsch, line 93: When text wraps, it's useful to have a bit of vertical padding acting as a security to ensure that the text will not end up touching the top or bottom edge of the navbar.

With the specified line-height of 20px:

- Height for two lines `4 + 2x20 + 4 = 48`
- Height for three lines: `4 + 3x20 + 4 = 68`

We could theoretically use 8px of vertical padding and exactly match the default `--vp-nav-height-mobile` value (56px) for 2 lines of text. But that could get in the way of users making a custom theme and trying to use a lower `--vp-nav-height-mobile` value.

src/client/theme-default/components/VPFlyout.vue
```diff
@@ -105,19 +105,13 @@ function onBlur() {
 .text {
   display: flex;
   align-items: center;
-  line-height: var(--vp-nav-height-mobile);
```
fvsch, line 108: Button styles already center the button's contents, so we don't need to use the big line-height trick here anyway.

src/client/theme-default/components/VPFlyout.vue
```diff
@@ -126,6 +120,7 @@ function onBlur() {
 }
 
 .text-icon {
+  flex: none;
```
fvsch, line 123: When space become scarce, Flexbox on the parent compresses this SVG icon. In one of my tests it ended up being 9px wide instead of the specified 14px wide.

Rule of thumb for SVG icons that are flex children: always use `flex: none` or `flex-shrink: 0`.

src/client/theme-default/components/VPNavBar.vue
```diff
@@ -52,8 +52,8 @@ const { hasSidebar } = useSidebar()
   position: relative;
   border-bottom: 1px solid var(--vp-c-divider-light);
   padding: 0 8px 0 24px;
-  height: var(--vp-nav-height-mobile);
-  transition: border-color 0.5s, background-color 0.5s;
+  min-height: calc(var(--vp-nav-height-mobile) + 1px);
+  transition: background-color 0.5s;
```
fvsch, line 56: Transitioning the border-color made the border flash from white to its final dark color when using the dark theme and resizing the window (between large and medium layouts). Seen in Chrome and Firefox.

I think the goal was to have a smoother transition when changing the theme (in the small and medium layouts). The result for that without the border-color transition was not shocking to my eyes, unlike the white-to-dark flash on resize. But I can restore this transition setting if need be.

src/client/theme-default/components/VPNavBar.vue
```diff
@@ -52,8 +52,8 @@ const { hasSidebar } = useSidebar()
   position: relative;
   border-bottom: 1px solid var(--vp-c-divider-light);
   padding: 0 8px 0 24px;
-  height: var(--vp-nav-height-mobile);
-  transition: border-color 0.5s, background-color 0.5s;
+  min-height: calc(var(--vp-nav-height-mobile) + 1px);
```
fvsch, line 55: Two things here:

- We use a min-height rather than a fixed height for the NavBar. (This will be changed for the large layout a bit lower.)
- We need to account for the NavBar's border. The NavBar contains many elements which have a `height: var(--vp-nav-height-mobile)`. That means that if we use a fixed height, elements inside the NavBar will be painted on top of the border, and may look slightly misaligned. They will also obscure the border if they ever get an opaque or semi-opaque background. I think it's better to make the total height of the NavBar be the specified height plus one pixel for the border (or to make the height of NavBar content use `calc(var(--vp-nav-height-mobile) - 1px)` instead, but that's not ideal for optical alignment).

src/client/theme-default/components/VPNavBarMenu.vue
```diff
@@ -24,6 +24,8 @@ const { theme } = useData()
 @media (min-width: 768px) {
   .VPNavBarMenu {
     display: flex;
+    align-items: center;
+    padding-inline-start: 12px;
```
fvsch, line 28: The NavBarMenu didn't vertically center its children. Instead, all its children had a `height: var(--vp-nav-height-mobile)`. When changing those children heights to `min-height` instead, some children can end up being taller than others, so we do need `align-items` as well.

The `padding-inline-start` is to prevent menu items from ending up _very_ close to the Search button (or to the sidebar when there is no Search button).

src/client/theme-default/components/VPNavBarTitle.vue
```diff
@@ -22,14 +22,13 @@ const { hasSidebar } = useSidebar()
 <style scoped>
 .VPNavBarTitle {
   flex-shrink: 0;
-  border-bottom: 1px solid transparent;
```
fvsch, line 25: This was adding to the height of the title on small screens, making its height 57px instead of 56px, which made the navbar height 58px instead of 57px, etc.

Since we only mean to have a border on large screens, let's define the border's space only on large screens.

src/client/theme-default/components/VPNavScreen.vue
```diff
@@ -43,7 +43,7 @@ function unlockBodyScroll() {
 <style scoped>
 .VPNavScreen {
   position: fixed;
-  top: var(--vp-nav-height-mobile);
+  top: calc(var(--vp-nav-height-mobile) + 1px);
```
fvsch, line 46: We need to account for the new calculation of the navbar height, where the `--vp-nav-height-mobile` value is the height of the navbar's _excluding its border_.

fvsch (commented):
A few pointers to explain some CSS changes.

kiaking:
Closing due to https://github.com/vuejs/vitepress/issues/1271#issuecomment-1296940615

PR #1274, opened 2022-09-02
fix: update language regex to avoid user unexpect input

CHOYSEN:
reproduction:

`markdown.lineNumbers` with the following incorrect language `ts c`

```md
\```ts c
const a = 1
\```
```

image


brc-dd:
How is this compiling in the first place? 👀

brc-dd:
IG this is already fixed on main.

PR #1280, opened 2022-09-03
feat(theme): support dynamic headers and nesting in outline

### LINKED ISSUE #785: Show dynamic headers in aside outline

### LINKED ISSUE #954: expose a configurable outline config in frontmatter

brc-dd:
most of the stuff is based on #965

fixes #785, fixes #954, closes #965

TODO:
- [ ] clean up stuff
- [ ] update tests
- [ ] add docs

### DIFF
--- a/src/client/theme-default/components/VPDocAsideOutline.vue
+++ b/src/client/theme-default/components/VPDocAsideOutline.vue
@@ -1,39 +1,12 @@
 <script setup lang="ts">
-import type { DefaultTheme } from 'vitepress/theme'
 import { useData } from 'vitepress'
 import { computed, onMounted, onUnmounted, ref } from 'vue'
 import VPDocAsideOutlineItem from './VPDocAsideOutlineItem.vue'
-import {
-  useActiveAnchor,
-  resolveHeaders,
-  type MenuItem
-} from '../composables/outline.js'
+import { useActiveAnchor, getHeaders } from '../composables/outline.js'
 
-const { frontmatter, theme } = useData()
+const { theme } = useData()
 
-const pageOutline = computed<DefaultTheme.Config['outline']>(
-  () => frontmatter.value.outline ?? theme.value.outline
-)
-
-const getHeaders = () => {
-  if (pageOutline.value === false) return []
-  let updatedHeaders: MenuItem[] = []
-  document
-    .querySelectorAll<HTMLHeadingElement>('h2, h3, h4, h5, h6')
-    .forEach((el) => {
-      if (el.textContent && el.id) {
-        updatedHeaders.push({
-          level: Number(el.tagName[1]),
-          title: el.innerText.split('\n')[0],
-          link: `#${el.id}`,
-          children: []
-        })
-      }
-    })
-  return resolveHeaders(updatedHeaders, pageOutline.value)
-}
-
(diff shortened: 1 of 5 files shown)


src/client/theme-default/components/VPDocAsideOutline.vue
```diff
@@ -1,21 +1,56 @@
...
+const { frontmatter, theme } = useData()
+const route = useRoute()
+
+const pageOutline = computed<DefaultTheme.Config['outline']>(
+  () => frontmatter.value.outline ?? theme.value.outline
+)
+
+const headers = ref<MenuItem[]>([])
+
+watch(
```
brc-dd, line 21: This watcher thing is currently running on route change. It won't work if some content is manipulated later or a node is added/removed. Need to use MutationObserver here, but things are getting complex in that.

PR #1283, opened 2022-09-05
fix: nav items overflow

dboooo:
solve:https://github.com/vuejs/vitepress/issues/1271 and https://github.com/ElMassimo/vitepress-theme/issues/1
seems like this:https://github.com/vuejs/theme/pull/59


dboooo:
https://github.com/vuejs/vitepress/pull/1273 is better

PR #1320, opened 2022-09-11
perf: render pages asynchronously

sanjaiyan-dev:
Improving the performance by running the independent async functions in parallel ✈️.

_Sorry if I made any mistakes :(_

### DIFF
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -65,8 +65,8 @@ export async function build(
 
       const pages = ['404.md', ...siteConfig.pages]
 
-      for (const page of pages) {
-        await renderPage(
+      await Promise.allSettled(
+        pages.map((page) => renderPage(
           render,
           siteConfig,
           page,
@@ -75,8 +75,8 @@ export async function build(
           cssChunk,
           pageToHashMap,
           hashMapString
-        )
-      }
+        ))
+      )
     } catch (e) {
       spinner.stopAndPersist({
         symbol: failMark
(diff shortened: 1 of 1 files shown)


src/node/build/build.ts
```diff
@@ -65,8 +65,8 @@ export async function build(
 
       const pages = ['404.md', ...siteConfig.pages]
 
-      for (const page of pages) {
-        await renderPage(
+      await Promise.allSettled(
```
brc-dd, line 68: Any specific reason to use `allSettled` instead of `all`? `allSettled` will always get resolved (even if there are errors while rendering).

sanjaiyan-dev: Sorry, I thought no need to throw an error even if some page failed to generate.

So do I need to change it to `Promise.all` ?

brc-dd: Yeah change it to `Promise.all` and fix linting errors (run `pnpm run lint` before committing).

sanjaiyan-dev:
@brc-dd Done 🤝

src/node/build/build.ts
sanjaiyan-dev: > Yeah change it to `Promise.all` and fix linting errors (run `pnpm run lint` before committing).

Extremely sorry, I used github web based editor to commit the changes :( so, I dont have choice to run cli commands (my local machine having some issues) so can you please help me with it :) (sorry for incontinent )

src/node/build/build.ts
```diff
@@ -114,5 +116,5 @@ function linkVue(root: string) {
       fs.unlinkSync(dest)
     }
   }
-  return () => {}
+  return () => { }
```
brc-dd, line 119:
```suggestion
  return () => {}
```

sanjaiyan-dev:
Tks a lot @brc-dd

PR #1323, opened 2022-09-11
feat: add `transformHead` hook

### LINKED ISSUE #1322: allow modify head before SSG rendering

Sometimes there is no way to add headers to VitePress config, for example, when using `vite-plugin-pwa` we need to add the webmanifest link to the head, but, since the plugin will not generate it in dev mode, we'll have errors if we register it in the config file.

Another asset is the one for registering the service worker: `vite-plugin-pwa` will allow the user registering using an inlined script or generating a new asset (`registerSW.js`). This script can be also omited, if the user import one of the virtual modules provided also by the plugin:

```html
<script>
if('serviceWorker' in navigator) {window.addEventListener('load', () => {navigator.serviceWorker.register('/sw.js', { scope: '/' })})}
</script>
```
or

```html
<script src="/registerSW.js"></script>
```

In both cases, if the pwa plugin needs to generate it is known in the  `load` pwa main plugin hook (importing the virtual: https://github.com/antfu/vite-plugin-pwa/blob/main/src/plugins/main.ts#L36).

Add a new build hook similar to `transformHtml`.

Write custom scripts to modify all pages: you can check it in the following repos:
- Vite Plugin PWA: https://github.com/antfu/vite-plugin-pwa/blob/main/docs/scripts/assets.ts#L46
- Vitest: https://github.com/vitest-dev/vitest/blob/main/docs/.vitepress/scripts/assets.ts#L56

The main problem with this approach is that VitePress runtime will have an hydration error, since we're modifying the html files.

### Additional context

I'll provide a PR for it, adding the logic in the `render.ts` module (like `transformHtml` build hook).

### Validations

userquin:
closes #1322

### DIFF
--- a/docs/config/app-configs.md
+++ b/docs/config/app-configs.md
@@ -255,6 +255,37 @@ VitePress build hooks allow you to add new functionality and behaviors to your w
 - Search Indexing
 - PWA
 
+### transformHead
+
+- Type: `( ctx: Omit<TransformContext, 'content'> ) => Awaitable<void>`
+
+`transformHead` is a build hook to transform the head before rendering each page: it will allow you to add head entries that cannot be statically added to your VitePress config module.
+
+``` warning
+The header provided in `TransformContext` is a clone, any headers you want to include on all pages will need to be added on each call.
+```
+
+```ts
+import { defineConfig } from 'vitepress'
+
+export default defineConfig({
+  /* other vitepress options */
+  async transformHead(ctx: TransformContext) {
+  }
+})
+```
+
+```ts
+interface TransformContext {
+  siteConfig: SiteConfig
+  siteData: SiteData
+  pageData: PageData
+  title: string
+  description: string
+  head: HeadConfig[]
+}
+```
+
 ### transformHtml
 
 - Type: `( code: string, id: string, ctx: TransformContext ) => Awaitable<string | void>`
@@ -276,13 +307,7 @@ export default defineConfig({
 ```
 
 ```ts
-interface TransformContext {
-  siteConfig: SiteConfig
-  siteData: SiteData
-  pageData: PageData
-  title: string
-  description: string
-  head: HeadConfig[]
+interface HtmlTransformContext extends TransformContext {
   content: string
 }
 ```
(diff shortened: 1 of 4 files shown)


userquin:
you can check it working in this repo: https://github.com/userquin/vite-pwa-vitepress-integration
- from root folder run `pnpm install && pnpm run build` (node with corepack enabled required: read the contributing guide)
- change to `exmaples/pwa-simple` and run `pnpm run preview`

brc-dd:
I've made some changes, check once if they work well for you.

userquin:
@brc-dd I will try to check it later

docs/config/app-configs.md
```diff
@@ -255,24 +255,21 @@ VitePress build hooks allow you to add new functionality and behaviors to your w
...
 
-- Type: `( code: string, id: string, ctx: TransformContext ) => Awaitable<string | void>`
+- Type: `(ctx: TransformContext) => Awaitable<HeadConfig[]>`
 
-`transformHtml` is a build hook to transform the content of each page before saving to disk (SSG).
+`transformHead` is a build hook to transform the head before generating each page. It will allow you to add head entries that cannot be statically added to your VitePress config. You only need to return extra entries, they will be merged automatically with the existing ones.
 
 ::: warning
-Modifying the html content may cause hydration problems in runtime.
+Don't mutate anything inside the `ctx`.
```
userquin, line 265: shouldn't be also added to  `transformHtml`?

userquin:
integration repo updated: the pwa-prompt also configured

docs/config/app-configs.md
```diff
@@ -287,18 +284,32 @@ interface TransformContext {
...
 ```
 
+### transformHtml
+
+- Type: `(code: string, id: string, ctx: TransformContext) => Awaitable<string | void>`
+
+`transformHtml` is a build hook to transform the content of each page before saving to disk.
+
+::: warning
+Modifying the html content may cause hydration problems in runtime.
```
brc-dd, line 294:
```suggestion
Don't mutate anything inside the `ctx`. Also, modifying the html content may cause hydration problems in runtime.
```

userquin: I'll check if we can also integrate the head in the client build to solve the hydration missmatch, changes LGTM (I think it can be merged: in the meantime, we can add also the hydration problems to `transformHead`)

userquin:
I'll check if we can also integrate the head in the client build to solve the hydration missmatch, changes LGTM (I think it can be merged: in the meantime, we can add also the hydration problems to transformHead)

PR #1325, opened 2022-09-12
fix(socialIcons): Add `aria-label` attribute

cawa-93:
For better accessibility, I propose add an explicit label to the icons


brc-dd:
Refer https://github.com/simple-icons/simple-icons/issues/1135#issuecomment-454674687, https://github.com/simple-icons/simple-icons/issues/1191

Basically these changes won't do anything. SVG 2.0 is not standardized yet. (And there is no browser compatibility data associated with that.) SVG 1.1 has no accessibility API mappings. Also, we already have `title` present in the SVGs, so assistive technologies should be fine with that.

PR #1350, opened 2022-09-17
Optimize image by adding decoding attribute 🖼

sanjaiyan-dev:
Optimize ``

> 1. Adding `decoding` attribute with `async` value which decode the image asynchronously to reduce delay in presenting other content.

_Sorry if I made any mistakes :(_


kiaking:
Checking if this is a good idea or not. I'll come back soon.

kiaking:
Closing the PR due to couldn'T get confirmation on if this is a good idea to add as a default 🤔 For generic images, I think users could add this manually, and if we're missing the feature to add this attr to default theme images, maybe we should add that feature instead.

Thank you for the PR though!

PR #1353, opened 2022-09-17
fix(build): line numbers mode when language specifier has symbol

### LINKED ISSUE #1352: language-c# missing 'line-numbers-mode' class
### Describe the bug

I have enabled `lineNumbers` option in `config`
The `ts`, `html`, `js` works fine

But the `c#` is missing `line-numbers-mode` class so that losing `padding-left` style
Screen Shot 2022-09-17 at 20 39 53

Is this a bug? maybe the `#` problem?

By vitepress-1.0.0-alpha.15

### Validations

sun0day:
fix #1352 

Code block's `language-` class should  compatible with special char, such as '-', '+', '#'...

After PR:

```
"language-" => "language- line-numbers-mode"
"language-js" => "language-js line-numbers-mode"
"language-c#" => "language-c# line-numbers-mode"
"language-c++" => "language-c++ line-numbers-mode"
"language-Objective-C" => "language-Objective-C line-numbers-mode"
```

### DIFF
--- a/src/node/markdown/plugins/lineNumbers.ts
+++ b/src/node/markdown/plugins/lineNumbers.ts
@@ -21,7 +21,7 @@ export const lineNumberPlugin = (md: MarkdownIt) => {
 
     const finalCode = rawCode
       .replace(/<\/div>$/, `${lineNumbersWrapperCode}</div>`)
-      .replace(/"(language-.*)"/, '"$1 line-numbers-mode"')
+      .replace(/"(language-\S*?)"/, '"$1 line-numbers-mode"')
 
     return finalCode
   }
--- a/src/node/markdown/plugins/lineNumbers.ts
+++ b/src/node/markdown/plugins/lineNumbers.ts
@@ -21,7 +21,7 @@ export const lineNumberPlugin = (md: MarkdownIt) => {
 
     const finalCode = rawCode
       .replace(/<\/div>$/, `${lineNumbersWrapperCode}</div>`)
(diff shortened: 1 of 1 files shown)


brc-dd (changes requested):
This won't work with string like this:

```html
<div class="language-">..."...</div>
```

Probably something like `/"(language-[^\s"]*?)"/` should work

sun0day:
use `/"(language-.*?)"/` ?

`/"(language-[^\s"]*?)"/` dont work with `<div class="language-    xxxx">..."...</div>`

brc-dd:
There won't be any `<div class="language- xxxx">..."...</div>` case if that div is generated by our code highlighting plugin.

sun0day:
> There won't be any `<div class="language- xxxx">..."...</div>` case if that div is generated by our code highlighting plugin.

From the current code, yes, this case won't happen.

But what I concern is that from the `lineNumberPlugin` API aspect, it is better to be compatible with more cases

brc-dd:
Change it to either. I'd still recommend `\S` one, we should keep regexes stricter because they tend to mess back 😅. However, in this case both should work fine IG.

sun0day:
> Change it to either. I'd still recommend `\S` one, we should keep regexes stricter because they tend to mess back 😅. However, in this case both should work fine IG.

Alright. Let's change it to `/"(language-\S*?)"/` for now.

Done. Merge it at your convenience. @brc-dd

brc-dd (approved):

PR #1355, opened 2022-09-17
feat(build): support `cacheDir`

shigma:
I recommend keeping default vite cache directory under `.vitepress` for the following reasons:

- When the codebase its depends on vite, cache stored in `node_modules/.vite` would overwrite each other, causing unexpected and undetectable problems
- [vuepress next](https://vuepress2.netlify.app/reference/config.html#cache) already has similar configuration


kiaking:
I think the idea is nice. @shigma do you think you can resolve the failing tests?

shigma:
@kiaking Thanks for your attention.

The failing tests can be solved by https://github.com/vitejs/vite/pull/10242. If it can not be accepted, we can still manully add `ignore` pattern here in vitepress. It is just better to have it in vite from my perspective.

kiaking:
Ah, I see. Thanks! OK, let's wait for Vite side on this 👍

PR #1356, opened 2022-09-17
feat: add vue to `resolve.dedupe`

shigma:
Developers who wants to extend the theme may use libraries depending on different versions of vue, causing client errors. Adding vue to `resolve.dedupe` can avoid such conditions.

### DIFF
--- a/src/node/build/buildMPAClient.ts
+++ b/src/node/build/buildMPAClient.ts
@@ -41,6 +41,9 @@ export async function buildMPAClient(
           }
         }
       }
-    ]
+    ],
+    resolve: {
+      dedupe: ['vue']
+    }
   }) as Promise<RollupOutput>
 }
--- a/src/node/build/bundle.ts
+++ b/src/node/build/bundle.ts
@@ -48,6 +48,9 @@ export async function bundle(
       pageToHashMap,
       clientJSMap
     ),
+    resolve: {
+      dedupe: ['vue']
+    },
     ssr: {
       noExternal: ['vitepress', '@docsearch/css']
     },
--- a/src/node/server.ts
+++ b/src/node/server.ts
@@ -22,6 +22,9 @@ export async function createServer(
     base: config.site.base,
     // logLevel: 'warn',
     plugins: await createVitePressPlugin(config, false, {}, {}, recreateServer),
-    server: serverOptions
+    server: serverOptions,
+    resolve: {
+      dedupe: ['vue']
+    }
   })
 }


brc-dd:
I don't think `resolve.dedupe` will work for us as we are using SSR + ESM builds. Also, I suppose one can create a vite.config.ts themselves and add this option there (along with SSR + CJS settings).

shigma:
> I don't think resolve.dedupe will work for us as we are using SSR + ESM builds.

Multiple instances of vue may cause failure in development and result in multiple chunks in the bundle.

> I suppose one can create a vite.config.ts themselves and add this option there (along with SSR + CJS settings).

So does vitepress support vite config out of the box? (One may would like to use vitepress together with other component libraries.)

brc-dd:
> Multiple instances of vue may cause failure in development and result in multiple chunks in the bundle.

Yeah, I understood that part. But according to [Vite's docs](https://vitejs.dev/config/shared-options.html#resolve-dedupe), that doesn't work on SSR + ESM builds (<-- that's what we are using here). One needs to use CJS mode for SSR for it to work. And regarding Vite config, yes it is supported.

shigma:
> But according to [Vite's docs](https://vitejs.dev/config/shared-options.html#resolve-dedupe), that doesn't work on SSR + ESM builds.

Well I see. So should I just remove `resolve.dedupe` from SSR builds and keep the other two?

kiaking:
> Developers who wants to extend the theme may use libraries depending on different versions of vue

I think we should simply discourage users from using different versions of Vue...? What is the reason why you can't update the Vue version?

shigma:
@kiaking It is not about updating the vue version. Adding dependencies in a composite project may result in multiple instances of vue, something like `root > vue` and `root > @vueuse/core > vue`. This problem can be fixed but is sometimes hard to find out.

kiaking:
So, seems like Vite 4 still has same behavior at the moment. Maybe we should simple document about this issue...? It might be simpler.

@brc-dd Do you happen to have any news on this topic? 👀

kiaking:
@shigma Evan is currently looking into this problem. It's also an issue in Nuxt as well. While adding this option to Vite config is one way of solving it, but we're trying to fix the issues by looking into the problem deeper.

We're trying gather as much as reproduction here to see the actual cases. If you have reproduction code, please share. It would be super helpful 🙇

yyx990803:
Maybe only dedupe for client-side builds?

brc-dd:
vue plugin already dedupes this for client builds now - https://github.com/vitejs/vite-plugin-vue/blob/69fb8c811d8fdbc7bdc968033834ddede59ee885/packages/plugin-vue/src/index.ts#L163

PR #1361, opened 2022-09-18
feat: add support for collapsible multi-level sidebar

### LINKED ISSUE #1360: Add support for collapsible multi-level sidebar

I would like to have to collapsible navigation sidebar since I organize my documents  by using multiple layers of folders, which is frustrating to browse or lookup the content/page I want.

VitePress only supports the top-level folders to be collapsible in the navigation sidebar currently.

The default theme that VitePress provided could support multi-level sidebar to be collapsible since many of the people may also face such kind of the problem while using VitePress.

I have tried to override the `VPSidebarNav` component from default theme that VitePress provided, however, the theme configuration overridding is not supported currently (related issue is still open, ref: #235 ), or I would have to duplicate the default theme as a single repository and publish it as a npm package after the development is done by myself. Anyways, I may think this kind of the feature is general for most of the users, it should be provided by VitePress, nor by users.

### Additional context

_No response_

### Validations

LittleSound:
fixes #1360

I would like to have to collapsible navigation sidebar since I organize my documents by using multiple layers of folders, which is frustrating to browse or lookup the content/page I want.

VitePress only supports the top-level folders to be collapsible in the navigation sidebar currently.

### demo

I'm using this feature in my knowledge base site, if you want you can go see the results: [Nólëbase](https://nolebase.ayaka.io/)

collapsed
image
expand
image

### description

- When this node is both a page and a parent node, a split line will be added between the collapse button and the link to distinguish the functional area.
- If there are no links to jump to, switch to expand when clicking on the text
- Auto-expand active routes


types/default-theme.d.ts
```diff
@@ -148,7 +148,25 @@ export namespace DefaultTheme {
...
+         * @default false
+         */
+        collapsible?: boolean
+
+        /**
+         * If `true`, collapsible group is collapsed by default.
+         *
+         * @default false
+         */
+        collapsed?: boolean
```
emersonbottero, line 168: can't we only have this property? and if collapsed != undefined we assume the toggle should show?, this way we don't need the collapsible property. 
I know we already have collapsible but collapsed? should cover all options with that logic. 
Does that makes sense @kiaking?

LittleSound: What if the user would want to accomplish collapsible and expanded initially?

brc-dd: @LittleSound 

```
collapsed == null -> not collapsible
collapsed === true -> collapsible but collapsed
collapsed === false -> collapsible but expanded
```

LittleSound: I see. However, I ended up here like this by following the configuration design pattern that already existed. Modifying the configuration property is not the purpose of this PR, maybe it should be made as a separate issue?

brc-dd: > Modifying the configuration property is not the purpose of this PR

Ah don't worry, it wasn't specific to your PR. It needs to be done at top-level.

Zhengqbbb:
I note a feature. I dont it is need to support. But i think it will be nice~ 🧐

About **enter page auto open `collapsed: true` item** by URL. like navbar `active` option 

The demo: I have `collapsed: true` item. when i refresh or enter page by URL. The sidebar which `collapsed: true` item can be expanded.

[screen-capture.webm](https://user-images.githubusercontent.com/40693636/197682774-60bb1f18-9261-4387-b5cc-2c9456935135.webm)

brc-dd:
@Zhengqbbb It should open by default, we added that in #1094 . Can you confirm it's not auto expanding on the latest version? If so, please create a new issue. We'll fix that.

LittleSound:
> @Zhengqbbb It should open by default, we added that in #1094 . Can you confirm it's not auto expanding on the latest version? If so, please create a new issue. We'll fix that.

I actually found this problem in the code when I wrote this PR, and I wanted to submit a PR to rewrite that part of the problematic code with multi-layered collapsed sidebar supported after I finished this multi-layered collapsed navigation bar feature, but this PR was put on hold.

LittleSound:
@Zhengqbbb @brc-dd I implemented the auto-expand feature for the multi-layer navigation bar in the last commit, and it's now faster since it doesn't need to Iterating over the entire route 🚀 .

https://user-images.githubusercontent.com/19204772/198825072-5362ad48-c5aa-492f-8284-93b997dd2764.mov

kiaking:
I've opened new PR due to quite large design concept difference.
https://github.com/vuejs/vitepress/compare/1360-2nd-level-sidebar-collapse

PR #1363, opened 2022-09-18
feat: Line Numbers only for single code block

### LINKED ISSUE #1256: Line Numbers Only for some Code Blocks
### Discussed in https://github.com/vuejs/vitepress/discussions/1255

<div type='discussions-op-text'>

<sup>Originally posted by **akil-rails** August 30, 2022</sup>
Currently there is a global configuration to display Line numbers in all code blocks:  https://vitepress.vuejs.org/guide/markdown#line-numbers

It will be great to have line numbers only for some code blocks, and not for all. 

In certain cases (like for [example](https://docs.apigee.com/api-platform/troubleshoot/runtime/500-badformdata)) the code block is showing a sample message; which doesn't need line numbers. </div>

g4rry420:
Signed-off-by: GurkiranSingh <gurkiransinghk@gmail.com>

feat: Line Numbers only for single code block
- If a user specifies prop name `{showLineNumbers}`  then it will show line numbers for that particular code block
- Resolves #1256 

Please, let me know if you see any issues :)


src/node/markdown/plugins/lineNumbers.ts
```diff
@@ -3,10 +3,22 @@
...
+export const lineNumberPlugin = (
+  md: MarkdownIt,
+  isLineNumberOption: Boolean
+) => {
   const fence = md.renderer.rules.fence!
   md.renderer.rules.fence = (...args) => {
     const rawCode = fence(...args)
+    const [tokens, idx] = args
+    const token = tokens[idx]
+    const attr = token.attrs && token.attrs[0]
```
brc-dd, line 15: This logic won't work if markdown-it-attrs is disabled. As a fallback, you need to manually check token.info too for that option (refer highlightLines plugin for example, we had done similar stuff there). Also, have you checked if this works with line highlighting and snippets import? If yes, then consider adding some example to the docs.

g4rry420: @brc-dd On disabling `markdown-it-attrs`, I noticed that on the top right corner of code block where we are displaying the language for the code, the `{1,4,6-8}` - line highlightning syntax is also coming up. Is this intended or a bug ?
Screenshot 2022-09-24 132412

brc-dd: Yeah it's a bug.

g4rry420: I have resolved the above bug and also line numbers are now working even if `markdown-it-attrs` is disabled. Please, let me know if you see any issues :)

src/node/markdown/plugins/highlightLines.ts
```diff
@@ -4,7 +4,7 @@
 
 import MarkdownIt from 'markdown-it'
 
-const RE = /{([\d,-]+)}/
+const RE = /{([\w,-]+)}/
```
g4rry420, line 8: @brc-dd Only thing I am concerned about here is, I am allowing any word or character in it. I have tried this pattern `/{([(\d|\bshowLineNumbers\b),-]+)}/` but it didn't work. Please, let me know about a better regex pattern

g4rry420: I ended up using this - `/{(([\d,-]|\bshowLineNumbers\b)+)}/`

Issue #1376, opened 2022-09-21
Customize default page layout


JingMatrix:
### Is your feature request related to a problem? Please describe.

The default page layout is 'doc', which is hard coded. It will easier the user's life to be able to provide an aletrnative layout so that he doesn't need to write a whole theme.

### Describe the solution you'd like

Use dynamic component in VPContent, and add a config option in the themeConfig.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
Can you explain a bit on what's your use case?

JingMatrix:
My main motivation is that I want to set frontmatters programtically, but it is readonly.
 So I have to use different default layouts for different sets of files using locales to specify them.

kiaking:
Closing due to inactivity. I think we need more clear explanation on what this issue is trying to achive. When creating custom page layout, it's better to use `page` instead of `doc`.

PR #1377, opened 2022-09-21
feat(theme): customize default page layout

JingMatrix:
The default value of layout in the frontmatter is VPDoc, which can be changed now in themeConfig.

Solution for #1376

### DIFF
--- a/src/client/theme-default/components/VPContent.vue
+++ b/src/client/theme-default/components/VPContent.vue
@@ -7,7 +7,7 @@ import VPDoc from './VPDoc.vue'
 import { inject } from 'vue'
 
 const route = useRoute()
-const { frontmatter } = useData()
+const { frontmatter, theme } = useData()
 const { hasSidebar } = useSidebar()
 
 const NotFound = inject('NotFound')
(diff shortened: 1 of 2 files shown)


JingMatrix:
I should do some further work to support any custom layout option in the frontmatter.

JingMatrix:
Here is the code:
```js
let contentLayout = theme.pageLayout || "VPDoc"     ■ 'contentLayout' is declared but its value is never read.               
  if (frontmatter.layout != null) {
     if (frontmatter.layout == "doc") {
        contentLayout = "VPDoc"
      } else {    
        contentLayout = frontmatter.layout
      }                                                                                                                        
  }  
```

What do you think, guys? Should I keep the convention that `doc` refers to `VPDoc`? I am afraid that this name convention might be confusing some day

JingMatrix:
Sorry to pollute the comments you might receive, an actually working code shoul like
```vue
<component v-else :is="frontmatter.layout == 'doc' ? VPDoc : frontmatter.layout || theme.pageLayout || VPDoc" />
```
the js code in last comment won't work as `layout` is not a known key for frontmatter

kiaking:
I'm not sure how much we need this. Because if people want to have `page` as default, I think they're better off creating their own custom theme instead of using the default theme 🤔

PR #1385, opened 2022-09-22
feat(theme): allow frontmatter config for layout

JingMatrix:
One can set false to disable components such as
skiplink, backdrop, navbar, sidebar, footer

### DIFF
--- a/src/client/theme-default/Layout.vue
+++ b/src/client/theme-default/Layout.vue
@@ -29,9 +29,9 @@ const { frontmatter } = useData()
 <template>
   <div v-if="frontmatter.layout !== false" class="Layout">
     <slot name="layout-top" />
-    <VPSkipLink />
-    <VPBackdrop class="backdrop" :show="isSidebarOpen" @click="closeSidebar" />
-    <VPNav>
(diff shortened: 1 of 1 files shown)


CodyBontecou:
I'm a fan of this level of layout customization per page. Could you also add the documentation that would discuss these changes?

JingMatrix:
You meant add them as a commit? Let's wait to see how the author think about it.
I would be happy to do that.
If you want to see an example for now, my personal blog is one:

https://github.com/JingMatrix/viteblog/blob/8de48f7f2183de29006fad3200d7d0e7b37cbd68/content/en/index.md?plain=1#L2-L4

It renders a page with nav bar but no sidebar: https://jingmatrix.github.io/en/

kiaking:
I think it's good idea to have this kind of config but let's do it separately on component bases. For example I don't think it makes sense to have `skipLink: false` and `backdrop: false`. For example, backdrop wouldn't do nothing when `sidebar` is disabled so it wouldn't have any effect.

But it's going to be cluttered so let's tackle each issues in separate issues and PRs 👍

PR #1396, opened 2022-09-25
fix(theme): a gap appear when DocSearch is clicked in macOS

zonemeen:
A gap appear when DocSearch is clicked in macOS.

image


brc-dd:
That's weird. It's not happening for me on both Safari and Chrome. Are you sure that this is an issue with our theme and not because of one of your extensions?

zonemeen:
> That's weird. It's not happening for me on both Chrome and Safari (15.3, Monterey). Are you sure that this is an issue with our theme and not because of one of your extensions?

The problem is with chrome(102.0.5005.61), not with Safari. Let me check if the problem is caused by one of the extensions.

zonemeen:
I uninstalled all the extensions and still have this problem.

PR #1404, opened 2022-09-27
feat(theme): support links in homepage features (#984)

Jimmylxue:
Now, the features configuration on the home page is very beautiful, but sometimes we need to be able to jump to the specified articles for some features。

just like actions config on the home page, now just need to add "link" config that we can jump to other page。

## The configuration is modified as follows

```
---
layout: home

features:
  - icon: ⚡️
    title: vitepress
    details: hello world
    link: https://github.com/vuejs/vitepress
  - icon: 🖖
    title: vue
    details: Thanks for checking
  - icon: 🛠️
    title: TypeScript
    details:Thanks for checking
    link: https://github.com/vuejs/vitepress
---
```

> Link is an optional configuration. It remains unchanged by default when it is not configured


brc-dd:
I suppose we need either this or #1070. Let Kia decide what's best. The difference between these is this PR makes whole feature a link, while the other PR adds a button in it.

Jimmylxue:
fine~，I didn't know that pr existed。
I just think it is necessary. I added links and corresponding effect animations to the whole body, not through buttons.
Thank you for your reply。

yuetaibai:
Link for the homepage is very useful feature, when could it be merged into dev version, please update the progress, wait for use, thanks

kiaking:
Yes, I'll take this over. Sorry for the delay!

kiaking:
I've updated the design, and also added `linkText` feature too 🚀 

Screen Shot 2022-10-31 at 18 31 24

Jimmylxue:
> I've updated the design, and also added `linkText` feature too 🚀
> 
> Screen Shot 2022-10-31 at 18 31 24

that's awesome!

yuetaibai:
Great, thanks for the update !

Issue #280, opened 2021-04-03
feat: changeable entry file


privatenumber:
**Is your feature request related to a problem? Please describe.**

Vitepress defaults to `index.md` as the entry file. GitHub defaults to `README.md`/`readme.md` as the entry file for the directory, and I would like my repo and site to be consistent.

**Describe the solution you'd like**
An `entry` config that defaults to `index.md` but can be changed to `readme.md`

**Describe alternatives you've considered**

**Additional context**

kiaking:
Interesting. So you wanna make the docs readable in both on GitHub and on deployed site right...? That makes sense.

privatenumber:
Yeah, they would both be readable anyway, but I want the entry points to be consistent.

Also, the default entry point in Vuepress is `README.md` so supporting this also offers a smoother migration path.

rickalex21:
@privatenumber I think both are supported in vuepress but not in vitepress, I may be wrong here. I know I have both working in vuepress.

sigtm:
Would love this feature. In our case, it's not actually about `/docs` being readable in the actual site repo – large sections of our docs are automatically pulled from other repos when they have new releases. And those repos use `README.md` as the entry points, since they need to be readable on GitHub, independently of the main docs site. Which works great with Vuepress, but now hoping to migrate to Vitepress.

Bukunmikuti:
I was about filing a bug report for this. 
I believe this should be the expected behaviour as it allows for consistency.

PR #1429, opened 2022-10-02
feat: Specify the different entry file rather than index.md

g4rry420:
Signed-off-by: GurkiranSingh <gurkiransinghk@gmail.com>

User can specify different entry rather `index.md` by using the `entry` config option. Please, let me know if you see any issues :)
Resolves: #280


src/node/serve/serve.ts
```diff
@@ -31,9 +31,17 @@ export async function serve(options: ServeOptions = {}) {
   const notAnAsset = (pathname: string) => !pathname.includes('/assets/')
   const notFound = fs.readFileSync(path.resolve(site.outDir, './404.html'))
   const onNoMatch: IOptions['onNoMatch'] = (req, res) => {
-    res.statusCode = 404
-    if (notAnAsset(req.path)) res.write(notFound.toString())
-    res.end()
+    if (site.site.entry !== 'index' && req.path === '/') {
```
brc-dd, line 34: Yeah but this logic won't work when someone is deploying their site. It only works with our custom server which is only suitable for previewing the website locally after build. The servers specifically need `index.html` to be present. That being said, we need to build `entry` as if it was `index.md` (and emit `index.html`).

VuePress has probably done this using their permalink feature, which is still missing in VitePress.

g4rry420: I agree, I was worried by this too. I will look into VuePress implementation for this.

src/node/plugin.ts
```diff
@@ -153,6 +154,8 @@ export async function createVitePressPlugin(
           id,
           config.publicDir
         )
+        // console.log('id -- ', id)
+        // console.log('deadLinks -- ', deadLinks)
         if (deadLinks.length) {
```
g4rry420, line 159:
```suggestion
          config.publicDir
        )
        if (deadLinks.length) {
```

g4rry420:
@brc-dd If user specifies the different `entry` file other than `index`, then I am emitting `index.html` file, however, the entry file js chunk remained the same name, which shouldn't be a problem as we are only importing it in the index file.
Also, if user wants to add link to the `entry` page, then they will have to point to `index` file only, otherwise it will result in deadlinks.
Please, let me know if you see any issues :)

PR #1448, opened 2022-10-08
fix(theme): Make the menu traversable only when the menu is visible

### LINKED ISSUE #1332: menu (navigation drawer) should not be traversable with keyboard when not shown
### Describe the bug

When the menu is not shown (narrow screens), it shouldn't be traversable via keyboard tab.

Once the menu is shown, should have a focus trap: the tab key will traverse again the content (after a few tabs, the menu entries are traversed). The menu should trap the focus, and select the current entry once activated.

### Reproduction

Open VitePress docs, resize the browser window and use the keyboard tab key.

### Expected behavior

Menu shouldn't be traversable.

### System Info

```shell
NA
```

### Additional context

_No response_

### Validations

7nohe:
fixes #1332 

Set `tabindex="-1"`  and disable tab key navigation.

reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex


userquin:
@7nohe with tabindex we onky fix focus but we need to trap focus when menu is shown. We also need to add escape key listener to hide the menu once it is shown, selecting/focusing current entry (and maybe we need to add a close button).

7nohe:
Sure, we need to trap focus. I'll work on it later.

7nohe:
accidentally pushed wrong commits. I will create new one.

PR #1449, opened 2022-10-08
feat: Smooth Scrolling behavior

g4rry420:
Resolves: #979


Zhengqbbb:
Can provide preview website link?
I have tried to add, but there will be a jump header-anchor link position problem.

And this is not a feature that everyone wants it.  I feel that the jump link is very fast, I enjoy it without excessive animation and it is intuitive after I cancel it.

kiaking:
Let's discuss this a bit more. As [Zhengqbbb](https://github.com/Zhengqbbb) says, I think it needs to be configurable 👀 And, probably, there'll be some edge cases.

PR #1339, opened 2022-09-14
feat: support i18n

brc-dd:
## Breaking Changes

- Legacy i18n stuff is removed/replaced.

  - `localePath` is removed from `useData()`. One can have similar functionality using `localeIndex` or `lang` property.

- For people using the default theme, `themeConfig.localeLinks` is removed in favor of `locales`.

  Change:

  ```ts
  themeConfig: {
    localeLinks: {
        text: 'English',
        items: [
          { text: '简体中文', link: 'https://cn.vitejs.dev' },
          { text: '日本語', link: 'https://ja.vitejs.dev' },
          { text: 'Español', link: 'https://es.vitejs.dev' }
        ]
    }
  }
  ```

  to:

  ```ts
  locales: {
    root: { label: 'English' },
    zh: { label: '简体中文', link: 'https://cn.vitejs.dev' },
    ja: { label: '日本語', link: 'https://ja.vitejs.dev' },
    es: { label: 'Español', link: 'https://es.vitejs.dev' }
  }
  ```

- There are internal changes in Algolia Search and Carbon Ads components.

## Features

To use in-built i18n features, one need to create a file structure as following:

```
docs/
├─ es/
│  ├─ baz.md
├─ fr/
│  ├─ bar.md
├─ foo.md
```

Then in `docs/.vitepress/config`:

```ts
defineConfig({
  // shared properties and other top-level stuff...
  
  locales: {
    root: {
      label: 'English',
      lang: 'en'
    },
    fr: {
      label: 'French',
      lang: 'fr', // optional, will be added  as `lang` attribute on `html` tag
      link: '/fr/guide' // default /fr/ -- shows on navbar translations menu, can be external

      // other locale specific properties...
    }
  }
})
```

The following properties can be overridden for each locale (including root):

```ts
interface LocaleSpecificConfig<ThemeConfig = any> {
  lang?: string
  dir?: string
  title?: string
  titleTemplate?: string | boolean
  description?: string
  head?: HeadConfig[] // will be merged with existing head entries, duplicate meta tags are automatically removed
  themeConfig?: ThemeConfig // will be shallow merged, common stuff can be put in top-level themeConfig entry
}
```

- All of the labels (placeholder texts) of default theme can now be customized. Refer [`DefaultTheme.Config`](https://github.com/vuejs/vitepress/blob/feat/i18n-next/types/default-theme.d.ts) type for details.

<details>
  <summary><h2>Linked Issues</h2></summary>

- fixes #631
- fixes #628
- fixes #955
[...]

PR #1466, opened 2022-10-11
fix(theme): fix locale links ignoring base option

hir0kio:
Fix locale links under the "Languages" section on navigation bar not working properly on mobile devices when both `base` and `themeConfig.localeLinks` options are set by passing them to `withBase()` function first.


brc-dd:
Thanks, but the current implementation of locale links will soon be changed (#1339). Also, currently non-external locale links don't work properly. So I've added fix for this in that PR itself.

PR #1471, opened 2022-10-12
revert: "fix(build): remove leading underscore from chunks"

brc-dd:
Reverts vuejs/vitepress#1394

### DIFF
--- a/src/node/build/bundle.ts
+++ b/src/node/build/bundle.ts
@@ -68,7 +68,6 @@ export async function bundle(
         // other
         preserveEntrySignatures: 'allow-extension',
         output: {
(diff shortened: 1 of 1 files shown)


Mister-Hope:
Hi, could you make a little explanation why this is reverted? Is some issues caused by this?

brc-dd:
It was breaking client builds if documents have leading underscore. I'm not sure if people have that, but it was sort of an edge case. The regex needs some modifications to remove leading _ from file name not the path. I'll try to add it in #1475 itself. For now consider adding a `.nojekyll` file in your public directory if the issue is with GH Pages.

PR #1488, opened 2022-10-14
fix(build): better align server and client side filename sanitization

brc-dd:
same as #1475, just gonna remove lookbehind for safari compat


src/shared/shared.ts
```diff
@@ -161,3 +161,21 @@ function hasTag(head: HeadConfig[], tag: HeadConfig) {
...
+export function sanitizeFileName(name: string): string {
+  const match = DRIVE_LETTER_REGEX.exec(name)
+  const driveLetter = match ? match[0] : ''
+
+  return (
+    driveLetter +
+    name
+      .slice(driveLetter.length)
+      .replace(INVALID_CHAR_REGEX, '_')
+      .replace(/(?<=^|\/)_+(?=[^/]*$)/, '')
```
brc-dd, line 179:
```suggestion
      .replace(/(^|\/)_+(?=[^/]*$)/, '$1')
```

Issue #902, opened 2022-07-03
How to set the text direction?


xfq:
### Is your feature request related to a problem? Please describe.

There is currently no easy way to set the text direction of an entire website. While it is possible to add global CSS, that is [not the recommended approach](https://www.w3.org/International/tutorials/bidi-xhtml/index#browser).

### Describe the solution you'd like

It would be useful to add a config to set `dir="ltr/rtl"` of the `<html>` element, so that docs written in [RTL scripts](https://www.w3.org/International/questions/qa-scripts) using VP can be better supported.

### Describe alternatives you've considered

Use the CSS [direction](https://www.w3.org/TR/css-writing-modes-4/#direction) property to customize the theme.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

xfq:
The UI also needs to adapt automatically, for example, the sidebar on the left should change to the right in `dir="rtl"`.

If VitePress uses [CSS Logical Properties and Values](https://css-tricks.com/css-logical-properties-and-values/) (I haven't checked yet), there shouldn't be many changes that need to be made.

Issue #631, opened 2022-05-19
Add internationalization feature


kiaking:
related https://github.com/vuejs/vitepress/issues/628

VitePress Next don't have this feature yet. Let's implement it. While doing so, I would also like to suggest new config structure for `locales` option.

Currently, we must define two `locales` when defining multi locale site. One in app config level, and one in theme config level.

I think it's much easier to just be able to define locale on top level only.

```js
export default {
  // Top level config for default and fallbacks.
  lang: 'en',
  title: 'VitePress',
  themeConfig: { ... }
  
  // All other locale settings.
  locales: {
    fr: {
      lang: 'fr',
      title: 'VitePress',
      themeConfig: { ... }
    },

    ja: {
      lang: 'ja',
      title: 'VitePress',
      themeConfig: { ... }
    }
  }
}
```

This should make typings much simpler too (maybe). It should eliminate weird situation like in #628 🤔 

Also, maybe not required for 1.0.0 release, but consider #291 as well.

xfq:
FWIW, W3C has a [Developing Localizable Manifests document](https://www.w3.org/TR/localizable-manifests/) (since `.vitepress/config.js` is also a manifest file).

brc-dd:
Let's also track #902 here.

fvsch:
Is the intent for RTL layouts to use CSS logical properties?
I could work on that for the default theme, if that helps.

trincadev:
I think this issue is related to https://github.com/vuejs/vitepress/issues/41

Zhengqbbb:
@brc-dd I'm ready use the test package `vitepress@npm:@brc-dd/vitepress@next` in my website to support i18n. Thx u great work
And I noticed

> Don't override themeConfig.algolia and themeConfig.carbonAds at locale-level.

That I have a question, **Have some way to configure algolia search to support i18n**
my config: [en](https://github.com/Zhengqbbb/cz-git/blob/dd976879fb7787f74d7719f35690369a1b8d13a8/docs/.vitepress/config/en.ts#L16-L23) | [zh](https://github.com/Zhengqbbb/cz-git/blob/dd976879fb7787f74d7719f35690369a1b8d13a8/docs/.vitepress/config/zh.ts#L16-L25)
preview link: https://deploy-preview-80--cz-git.netlify.app/zh/
You can try search, it disappeared after I clicked it, sure my console should be warn it if it not support.

brc-dd:
[...]

PR #1512, opened 2022-10-19
feat(theme): add rtlcss directive

sadeghbarati:
https://vitepress-rtlcss.vercel.app/

This is just simple [**rtlcss**](https://www.npmjs.com/package/rtlcss) directive

Still requires to User to add `rtlcss` as their `devDependencies` and use it in `vitepress` config,
Because it has not been implemented in `node` folder

```ts
// .vitepress/config

import rtlcss from 'rtlcss'

export default defineConfig({
  lang: 'en-US',
  title: 'VitePress',
  description: 'Vite & Vue powered static site generator.',
  
  vite: {
    postcss: {
      plugins: [
        rtlcss()
      ]
    }
  }
})
```

<hr />

I saw [**i18n**](https://github.com/vuejs/vitepress/tree/feat/i18n) and [**i18n-next**](https://github.com/vuejs/vitepress/tree/feat/i18n-next) branch for handling `dir` and `lang` options and other feats,

If you add `dir` frontmatter would be so good, documentation sites support both RTL and LTR at the same time

`page-rtl.md`
```md
---
dir: 'rtl'
---

RTL Content
```

`page-ltr.md`
```md
---
dir: 'ltr'
---

LTR Content
```


brc-dd:
> If you add dir frontmatter would be so good, documentation sites support both RTL and LTR at the same time.

I can add that, but I don't think it will be necessary as the current implementation (i18n-next) already allows overriding `dir` based on locale. So, if you have some structure like this:

```
- foo.md
- fa/
--- foo.md
```

in your config you can override `dir` and `lang` for whole `fa/` directory. This way `/foo` can serve pages with ltr direction and `fa/foo` can serve pages in rtl direction at the same time.

---

Regarding your PR, I don't have much experience, so if someone can confirm this works fine I'll merge it.

brc-dd:
x-ref: #631, #902

---

@xfq, @fvsch Hi guys! Can you take a look if this sorts out issues with the RTL layout, and point out if something seems missing?

xfq:
"K⌘" in the search bar should be "⌘K" (on macOS). The same for other operating systems.

I didn't see any other obvious problems, but I think it would be useful to have an Arabic/Hebrew-speaking expert double-check the page and/or the code.

brc-dd:
I'm merging this with i18n(-next) branch for now. We can tackle other issues directly on #1339.

Regarding documenting both ltr and rtl in the same app, there will be some issues as the styles will be shared, so if you add the rtlcss postcss plugin. the styles for ltr languages will also change.

brc-dd:
> there will be some issues as the styles will be shared

Regarding this, I wanted to ask are you using override mode in `postcss-rtlcss` (at https://vitepress-rtlcss.vercel.app/)? Because according to its docs, the default is the combined mode, where ltr and rtl styles are separate (prefixed). I'm guessing that it somehow messes up with the specificity of selectors?

sadeghbarati:
Hi, 
as I mentioned before I'm using [`rtlcss`](https://github.com/MohammadYounes/rtlcss) not [`postcss-rtl`](https://github.com/vkalinichev/postcss-rtl), rtlcss package will transforms all styles from `RTL` to `LTR` or vice versa

`rtlcss`:
```css
/* before */
body {
  direction: ltr;
}
/* after */
body {
  direction: rtl;
}
```

---

but `postcss-rtl` will generate another style besides of ltr or rtl style, It can increase the size of CSS and also change selectors specificity

`postcss-rtl`:
```css
/* before */
body {
  direction: ltr;
}
/* after */
body {
  direction: ltr;
}
[dir="rtl"] body {
   direction: rtl;
}
```

sadeghbarati:
I didn't check the vitepress node folder completely https://github.com/vuejs/vitepress/blob/12ca9dd3c736bb8b3938452c36b316ec0396ef6a/src/node/build/render.ts#L103-L105
When you are using `rtlcss` you have to generate CSS two times `(.vitepress/dist/assets)`, one for `ltr` without any rtlcss proccess `(style.********.css)` and another with rtlcss `(style-rtl.********.css)`

since `vitepress` is `SSR` I thought it can handle two stylesheets based on locale directories

```
- foo.md => ltr stylesheet
- fa/
--- foo.md => rtl stylesheet
```

sadeghbarati:
> I can make it work, but it will need a full page reload for styles to apply 

Ow you're right I thought it would work

I think CSS Logical Properties is the propper way to achieve i18n goal then, however [support table](https://caniuse.com/?search=CSS%20Logical%20Properties) is colorful :dizzy_face:

brc-dd:
On second thoughts, we can use `:where([dir="rtl"])` as the prefix, it will keep the specificity same, and apply appropriate styles when the page is rtl.

Issue #903, opened 2022-07-03
Change URL hash when user scrolls the page


brc-dd:
Currently this code changes the URL hash when users click some link to a fragment identifier causing unwanted issues like #890 and #788.

https://github.com/vuejs/vitepress/blob/baf8083e668ce50a9d26d9398609e196237ebc95/src/client/theme-default/composables/outline.ts#L64-L152

Related SO thread: https://stackoverflow.com/q/7035896/11613622

I am thinking lets remove below line. The outline will still highlight current header. Just the hash won't change on URL.

https://github.com/vuejs/vitepress/blob/baf8083e668ce50a9d26d9398609e196237ebc95/src/client/theme-default/composables/outline.ts#L123

Side note, we also need to support `outline: false`.

brc-dd:
Removed that line for now, let's see if we can reintroduce that functionality properly.

PR #1527, opened 2022-10-22
feat: add hash to URL according to outline activate link

ylw5:
ezgif com-gif-maker


brc-dd:
This actually was intentionally removed. Refer #903 and linked issues.

PR #1530, opened 2022-10-24
fix(build): explicitly exit process after build to prevent hangup

NelsonYong:
When I used automated building, I found that vitepress did not quit the task at the end of the construction, causing the subsequent process


NelsonYong:
Blocking the subsequent process

PR #1544, opened 2022-10-27
Added smooth scrolling

jessekelly881:
Makes the ux feel a lot nicer. Specifically, adding this causes the toc to scroll to selected header on click.


jessekelly881:
And I just realized there is another pr for the same change. jaja. Closing.

PR #1569, opened 2022-11-02
feat: if image is configured to generate `background-image` value

btea:
When image is configured in `index.md`, the css variables `--vp-home-hero-image-background-image` and `--vp-home-hero-image-filter` will be applied to elements with class `image-bg`. However, their value is `none`, so I created this pr to generate it.

### DIFF
--- a/src/client/theme-default/components/VPHero.vue
+++ b/src/client/theme-default/components/VPHero.vue
@@ -2,6 +2,7 @@
 import type { DefaultTheme } from 'vitepress/theme'
 import VPButton from './VPButton.vue'
 import VPImage from './VPImage.vue'
+import { getColors } from '../../app/utils.js'
 
 export interface HeroAction {
   theme?: 'brand' | 'alt'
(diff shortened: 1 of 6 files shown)


src/client/theme-default/components/VPHero.vue
```diff
@@ -16,6 +17,22 @@ defineProps<{
...
+    const root = document.documentElement
+    let v: string[] = []
+    const useColors = colors.slice(0, 5)
+    const gap = Math.floor(100 / (useColors.length - 1))
+    useColors.forEach((c: number[], i: number) => {
+      v.push(`rgb(${c[0]}, ${c[1]}, ${c[2]}) ${i * gap}%`)
+    })
+    const bg = `linear-gradient(-45deg, ${v.join(', ')})`
+    root.style.setProperty('--vp-home-hero-image-background-image', bg)
+    root.style.setProperty('--vp-home-hero-image-filter', 'blur(72px)')
```
btea, line 33: I refer to the vite official website for the value here, and set it directly to `blur(72px)`. Maybe a configuration should be provided for users to set by themselves?

btea, line 33: I refer to the vite official website for the value here, and set it directly to `blur(72px)`. Maybe a configuration should be provided for users to set by themselves?

btea:
Yeah, I agree with what you. But, in my opinion, this is very convenient for people who have this need, and they don't need to get the color matching in the picture by themselves. 
Maybe provide a configurable option to let the user decide whether to apply this feature, WDYT?

brc-dd:
Not sure what would be the optimal approach here. But these operations seem heavy considering they will be run every time one opens the home page. Maybe we could have this in docs sections? Users can upload their image there and it will generate the required CSS code?

btea:
Yeah, that sounds good, could you add the relevant section or provide details like the layout?

brc-dd:
Yeah sure, I'll update this in 1-2 days.

btea:
Hi, it seems like it's been a while, did you have time to update the relevant section?

brc-dd:
Hi! Sorry for making you wait. Can you publish a web app where one can upload an image and generate gradients/code? Once you do that, I think we can add a section near this: https://vitepress.vuejs.org/guide/theme-home-page#customizing-the-name-color, mentioning the app URL and required CSS variables. Thanks!

PR #1575, opened 2022-11-03
fix: handle cleanUrls with subfolders when using a trailing slash

### LINKED ISSUE #1442: CleanUrls with with-subfolders/without-subfolders fails lookup from `__VP_HASH_MAP__` upon trailing slash
### Describe the bug

When entering a page from outside the main index `/`, the hydration fails when trying to access the chunk in [pathToFile](https://github.com/vuejs/vitepress/blob/main/src/client/app/utils.ts#L42) requesting `/assets/docs_quick-reference_index.md.undefined.lean.js` leading to a  `Hydration completed but contains mismatches.`

The underlying page loads successfully, but any further navigation is broken.

**NOTE**: This only occurs when there's a trailing slash. 

The key being used is `docs_quick-reference_index.md`, but the key should be `docs_quick-reference.md`. 

### Reproduction

Set cleanUrls with `with-subfolders` or `without-subfolders`, add a page further down the content tree, build the project, and access it directly from the browser with a trailing slash.

### Expected behavior

The lookup from `__VP_HASH_MAP__` should be successful and correctly load the lean chunk making the page interactive.

### System Info

```shell
System:
    OS: Linux 5.10 Ubuntu 18.04.6 LTS (Bionic Beaver)
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 16.96 GB / 23.44 GB
    Container: Yes
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.17 - ~/.npm-global/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.0/bin/npm
  npmPackages:
    vitepress: ^1.0.0-alpha.19 => 1.0.0-alpha.19
```

### Additional context

Possible solution could be to check for both an `index.md` key as well as a stripped key to cover both use cases.

### Validations

rigor789:
The main change is that when using `with-subfolders` the page should always be generated as `/page_name.md` -> `/page_name_index.md` (since it's generated as `/page_name/index.html`

On the client side, whether or not there's a trailing slash, it resolves to the `_index` file when using `cleanUrls.with-subfolders`. ~~Still need to check if this works/breaks `without-subfolders`.~~

~~Also need to clean up, there's duplicated logic now in various places.~~

There is one thing to note about relative links, since they are resolved by the browser, they look for the last `/` in the url, and resolve them relative to that. So depending on how the host redirects and handles trailing slashes, relative URLs might break.

For example, /pageA.md would produce the following files:
| mode | file path |
|-|-|
| disabled | /pageA.html |
| without-subfolders | /pageA.html |
| with-subfolders | /pageA/index.html |

And these depending on the hosting can end up with the following urls in the browser (excluding what vue-router would normally produce for internal links):
| /pageA.html | /pageA/index.html |
|-|-|
| /pageA | /pageA |
| /pageA/ | /pageA/ |
| /pageA.html | /pageA/index.html |

And depending on the url, a `./pageB` link would resolve like this
| URL | Resolved |
|-|-|
| /pageA | ✅ /pageB   |
| /pageA/ | ❌ /pageA/pageB |
| /pageA.html | ✅ /pageB |
| /pageA/index.html | ❌ /pageA/pageB |

So depending on the hosting, relative URLs might break, even with `cleanUrls` disabled.

/cc @brc-dd 

fixes #1442.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -7,7 +7,7 @@ export default defineConfig({
   description: 'Vite & Vue powered static site generator.',
 
   lastUpdated: true,
-  cleanUrls: 'with-subfolders',
+  cleanUrls: 'without-subfolders',
 
   head: [['meta', { name: 'theme-color', content: '#3c8772' }]],
 
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -7,7 +7,7 @@ export default defineConfig({
   description: 'Vite & Vue powered static site generator.',
 
   lastUpdated: true,
-  cleanUrls: 'without-subfolders',
+  cleanUrls: 'with-subfolders',
 
   head: [['meta', { name: 'theme-color', content: '#3c8772' }]],
 
--- a/docs/config/app-configs.md
+++ b/docs/config/app-configs.md
@@ -69,7 +69,10 @@ Additional elements to render in the `<head>` tag in the page HTML. The user-add
 ```ts
 export default {
   head: [
-    ['link', { rel: 'preconnect', href: 'https://fonts.gstatic.com', crossorigin: '' }]
+    [
+      'link',
+      { rel: 'preconnect', href: 'https://fonts.gstatic.com', crossorigin: '' }
+    ]
     // would render: <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
   ]
 }
@@ -163,6 +166,9 @@ interface MarkdownOptions extends MarkdownIt.Options {
     allowedAttributes?: string[]
     disable?: boolean
   }
+
+  // specify default language for syntax highlighter
+  defaultHighlightLang?: string
 
   // @mdit-vue/plugin-frontmatter plugin options.
   // See: https://github.com/mdit-vue/mdit-vue/tree/main/packages/plugin-frontmatter#options
@@ -298,6 +304,7 @@ Don't mutate anything inside the `ctx`.
 ```ts
 export default {
   async transformHead(ctx) {
+    // ...
   }
 }
 ```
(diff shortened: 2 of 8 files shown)


brc-dd:
Dev appears to be broken after this if we have some path like `/foo/index.md` (without `cleanUrls` set).

rigor789:
> Dev appears to be broken after this if we have some path like `/foo/index.md` (without `cleanUrls` set).

This seems to fix it:
image

What is the expected behavior when you have `cleanUrls` enabled, and have a `/guide/markdown.md` - then open the following urls:
 - `/guide/markdown`
 - `/guide/markdown/`

(note, I'll push up the above change, just wanted to clarify if it's working as intended, or we need additional logic somewhere)

brc-dd:
/guide/markdown --> /guide/markdown.md

/guide/markdown/ --> /guide/markdown.md or /guide/markdown/index.md whichever exists if (cleanUrls === 'with-subfolders') else /guide/markdown/index.md

rigor789:
> /guide/markdown --> /guide/markdown.md
> 
> /guide/markdown/ --> /guide/markdown.md or /guide/markdown/index.md whichever exists if (cleanUrls === 'with-subfolders') else /guide/markdown/index.md

Ok, then we need some additional logic, with the above it works as before for the most part in dev mode:
- /guide/markdown -> /guide/markdown.md
- /guide/markdown/ -> /guide/markdown/index.md

brc-dd:
Yeah, push it anyway though. If everything is working for the other two options and we can fix build properly for with-subfolders+trailing-slash, then that's fair too.

PR #1580, opened 2022-11-04
feat(client): use v-html in VPDocFooter

userquin:
I need to check a few more links.

From discord https://discord.com/channels/325477692906536972/790509637598314527/1037735666789912596


userquin:
@brc-dd any data comming from the theme config should be escaped and use it with `v-html` directive. We also need to review the frontmatter to also allow configure it.

brc-dd (approved):

PR #1582, opened 2022-11-05
feat(theme) sidebar nav slots

marshallswain:
For the new FeathersJS docs, we've been using [this hack](https://github.com/vuejs/core/issues/2097#issuecomment-707975628) to show select boxes above the sidebar nav.  It works ok but occasionally detaches from the DOM. (Geez, lazy much, Marshall? lol 🤷)

This PR adds `sidebar-nav-before` and `sidebar-nav-after` slots to the default theme layout.  I've also listed the new slots in the documentation.

Below is a screenshot of what we're achieving.  You can see it in action on the Feathers v5 Dove site: https://dove.feathersjs.com/guides/basics/starting.html

Screen Shot 2022-11-05 at 11 27 55 AM

### DIFF
--- a/docs/guide/theme-introduction.md
+++ b/docs/guide/theme-introduction.md
@@ -201,6 +201,8 @@ Full list of slots available in the default theme layout:
   - `doc-footer-before`
   - `doc-before`
   - `doc-after`
+  - `sidebar-nav-before`
+  - `sidebar-nav-after`
   - `aside-top`
   - `aside-bottom`
   - `aside-outline-before`
--- a/src/client/theme-default/Layout.vue
+++ b/src/client/theme-default/Layout.vue
@@ -40,7 +40,11 @@ const { frontmatter } = useData()
       <template #nav-screen-content-after><slot name="nav-screen-content-after" /></template>
     </VPNav>
     <VPLocalNav :open="isSidebarOpen" @open-menu="openSidebar" />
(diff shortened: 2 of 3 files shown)


marshallswain:
Thanks for merging and thanks for the style tip.  Could you tell me if there is a scheduled date for pre.28?

marshallswain:
I've been impressed with the default theme.  We've been very happy with how nice it is to customize.

brc-dd:
I'll release it today itself.

erondpowell:
@marshallswain we just built a "toggler dropdown" like you have in your sidebar for our docs site, except we opted to make it "in the page". Is it possible to view your dropdown component? I'd like to see how you did yours just for fun 😃

marshallswain:
@erondpowell sure, it can be seen here: https://github.com/feathersjs/feathers/blob/dove/docs/.vitepress/components/Select.vue

It's just a basic HTML select with some experimental code to get the Select component into the sidebar.  The only props it should really have is the value, label, and options.  Everything else is a tightly coupled workaround for not previously having slots.  The routing code shouldn't really be inside this component, which I'll fix now that we have slots in the sidebar.  The hack I used to get the component into the sidebar is here: https://github.com/feathersjs/feathers/blob/dove/docs/.vitepress/theme/FeathersLayout.vue#L32. May god bless anybody who tries to use it, especially now that we have proper slots  :)

PR #1587, opened 2022-11-07
refactor tests

brc-dd:


userquin:
@brc-dd it seems to be ok, I run `test`, `test-unit`, `test-serve` and `test-build` multiple times with no errors on Windows

Issue #1515, opened 2022-10-20
More detailed description about `useData()` in docs


Jinjiang:
### Is your feature request related to a problem? Please describe.

In the docs, the returned value of `useData()` is described as:

```ts
interface VitePressData {
  site: Ref<SiteData>
  page: Ref<PageData>
  theme: Ref<any> // themeConfig from .vitepress/config.js
  frontmatter: Ref<PageData['frontmatter']>
  lang: Ref<string>
  title: Ref<string>
  description: Ref<string>
  localePath: Ref<string>
}
```

However, for `theme`, some of the fields are `Proxy` objects, and some of them are raw.

Take `@vue/theme` as an example:

![image](https://user-images.githubusercontent.com/206848/196982335-123b8b9a-fead-4744-9791-8e56a3421e64.png)

These details are not reflected in our docs. This potentially leads users confused when they consume the data.

### Describe the solution you'd like

Provide a more detailed description of the return value, and even better, an example.

### Additional context

A relevant issue: https://github.com/vuejs/theme/pull/77

Thanks.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

PR #1601, opened 2022-11-12
docs: add useData theme type

Soya-xy:
#1515


brc-dd:
Not everyone uses the default theme.

Issue #1608, opened 2022-11-15
mono+dev+enhanceApp+fs.server.strict: sometimes run error


meedfine:
### Describe the bug

![image](https://user-images.githubusercontent.com/35157761/201805961-e14485fd-2dd3-4aaa-bace-35d8324ead45.png)

### Reproduction

https://github.com/meedfine/server-fs-allow.git
test win10,win11, it happened

1. pnpm run docs
2. visit http://localhost:5173/
3.close
4. pnpm run docs
then will happen,or retry 3,4

https://user-images.githubusercontent.com/35157761/201807046-c6e534ae-722a-4149-86f0-aab839296381.mp4

### Expected behavior

work well

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (4) x64 Intel(R) Core(TM) i3-3240 CPU @ 3.40GHz
    Memory: 9.72 GB / 15.87 GB
  Binaries:
    Node: 16.14.2 - C:\Program Files\nodejs\node.EXE
    npm: 8.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 78.0.3904.108
    Edge: Spartan (44.19041.1266.0), Chromium (107.0.1418.42)
    Internet Explorer: 11.0.19041.1566
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #1609, opened 2022-11-15
fix mono+dev+enhanceApp+fs.server.strict: sometimes run error ?

meedfine:
fix https://github.com/vuejs/vitepress/issues/1608

maybe
while not use theme/index.js, vitepress will use theme-default/index.js, and will prebundling VPNavBarSearch.vue and '@docsearch/css'   and will hit  transformMiddleware, then be well

while use theme/index.js, it likes not prebundled,  and will  hit  vite/serveRawFsMiddleware, then be 403

### DIFF
--- a/src/node/plugin.ts
+++ b/src/node/plugin.ts
@@ -112,7 +112,7 @@ export async function createVitePressPlugin(
         },
         optimizeDeps: {
           // force include vue to avoid duplicated copies when linked + optimized
-          include: ['vue'],
+          include: ['vue', 'vitepress/theme'],
           exclude: ['@docsearch/js', 'vitepress']
         },
         server: {


brc-dd:
Can you try if instead of this fix, changing this line:

https://github.com/vuejs/vitepress/blob/428eec3750d6b5648a77ac52d88128df0554d4d1/src/node/plugin.ts#L120

to:

```ts
// import { searchForWorkspaceRoot } from 'vite'

 allow: [DIST_CLIENT_PATH, srcDir, searchForWorkspaceRoot(process.cwd())] 
```

works?

brc-dd:
Can you try changing this line instead of your fix:

https://github.com/vuejs/vitepress/blob/428eec3750d6b5648a77ac52d88128df0554d4d1/src/node/plugin.ts#L120

to:

```ts
// import { searchForWorkspaceRoot } from 'vite'

allow: [DIST_CLIENT_PATH, srcDir, searchForWorkspaceRoot(process.cwd())]
```

I've tried running dev after this several times, and the issue seem to be resolved.

PR #1631, opened 2022-11-21
feat: add Set custom menu/return to top labels

tomiaa12:
The text of nav can be modified when moving


brc-dd:
This is already fixed in feat/i18n-next branch (#1339). Thanks for the PR though.

PR #1644, opened 2022-11-28
docs: GitHub Actions deploy Workflow

FelixLuciano:
Previously I needed 2 workflows to publish to GitHub Pages. The documentation workflow and another through the repository's Pages settings.

This PR changes the current setup for deploying using GitHub Actions. It allows to build and deploy in a single job run using GitHub's officials actions instead of building and deploying from a branch.

Here is a brief of it working on my project. https://github.com/FelixLuciano/celulas-solares/actions/runs/3567337170

[Learn more about deploying to GitHub Pages using custom workflows](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow)

### DIFF
--- a/docs/guide/deploying.md
+++ b/docs/guide/deploying.md
@@ -73,15 +73,17 @@ Don't enable options like _Auto Minify_ for HTML code. It will remove comments f
 
    ```yaml
    name: Deploy
-
    on:
+     workflow_dispatch: {}
      push:
        branches:
          - main
-
    jobs:
      deploy:
        runs-on: ubuntu-latest
+       environment:
+         name: github-pages
+         url: ${{ steps.deployment.outputs.page_url }}
        steps:
          - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


brc-dd:
Yeah I'm aware of the new actions based publishing workflows. But they are still not available to everyone. So I'm keeping this PR on hold for now. After general availability and their stable release, we'll update our recommended way.

FelixLuciano:
> Yeah I'm aware of the new actions based publishing workflows. But they are still not available to everyone. So I'm keeping this PR on hold for now. After general availability and their stable release, we'll update our recommended way.

Thanks!

> Also, earlier too, I don't think 2 workflows were necessary. In the repo pages settings, one just needs to select the branch not create another workflow.

Deploy from branch is an Action under the hood, I mean.

PR #1655, opened 2022-12-01
docs: add interface of the "image"

lmq352489099:
docs: add interface of the "image"


lmq352489099:
@brc-dd   Hello, let's see if it's missing

Issue #472, opened 2021-12-15
Issues with peer dependencies


lampewebdev:
### Describe the bug

vitepress has missing peer dependencies

```
➜ pnpm up
Packages: +4 -4
++++----
Progress: resolved 777, reused 759, downloaded 2, added 4, done

devDependencies:
- @types/node 16.11.12
+ @types/node 16.11.13
- vitepress 0.20.6
+ vitepress 0.20.9

 WARN  Issues with peer dependencies found
.
└─┬ vitepress
  └─┬ @docsearch/js
    └─┬ @docsearch/react
      ├── ✕ missing peer @types/react@">= 16.8.0 < 18.0.0"
      ├── ✕ missing peer react@">= 16.8.0 < 18.0.0"
      ├── ✕ missing peer react-dom@">= 16.8.0 < 18.0.0"
      └─┬ @algolia/autocomplete-preset-algolia
        └── ✕ missing peer @algolia/client-search@^4.9.1
Peer dependencies that should be installed:
  @algolia/client-search@^4.9.1      react-dom@">= 16.8.0 < 18.0.0"     
  @types/react@">= 16.8.0 < 18.0.0"  react@">= 16.8.0 < 18.0.0" 
```

### Reproduction

`pnpm up` or `pnpm i`

### Expected behavior

Have no missing peer dependencies

### System Info

```shell
System:
    OS: macOS 11.6.2
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 2.74 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/Library/pnpm/node
    Yarn: 1.22.17 - ~/Library/pnpm/yarn
    npm: 8.1.2 - ~/Library/pnpm/npm
    Watchman: 2021.12.06.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 96.0.4664.55
    Chrome Canary: 99.0.4766.0
    Firefox: 95.0
    Firefox Developer Edition: 96.0
    Firefox Nightly: 97.0a1
    Safari: 15.1
  npmPackages:
    vitepress: ^0.20.9 => 0.20.9
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/coc)
- [X] Read the [docs](https://vitepress.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

[...]

PR #1678, opened 2022-12-08
docs: use pnpm

kirklin:


brc-dd:
Refer the discussion here: https://github.com/vuejs/vitepress/discussions/1149

Using pnpm still gives a bunch of errors during first installation (due to Algolia stuff) and requires some additional config (https://github.com/vuejs/vitepress/issues/472). So, it might not be a good idea to default it yet. We will surely update it once DocSearch has fixed its issues.

PR #1687, opened 2022-12-12
fix(snippet): dedent of a single-line region, close #1686

### LINKED ISSUE #1686: `dedent` in Snippet Markdown Plugin doesn't work for a single-line region
### Describe the bug

I have the following snippet:

```rs
fn main() {
    // #region ipsum
    println!("Lorem ipsum");
    // #endregion ipsum
}
```

I am importing it with the syntax:

```md
<<<@/snippets/lorem.rs#ipsum
```

The rendered result:

```rs
    println!("Lorem ipsum");
```

### Reproduction

https://stackblitz.com/edit/vite-skqr7n

### Expected behavior

Expected to see:

```rs
println!("Lorem ipsum");
```

and not

```rs
    println!("Lorem ipsum");
```

### System Info

```shell
N/A
```

### Additional context

I've almost created a PR with fix & tests.

### Validations

0x009922:
I had to `export dedent` function in order to cover it with unit tests. However, it would be much better to leverage [in-source testing](https://vitest.dev/guide/in-source.html) in this case.

Closes #1686


src/node/markdown/plugins/snippet.ts
```diff
@@ -3,29 +3,18 @@ import path from 'path'
...
-          ? Math.min(minIndentLength, indentLength)
-          : indentLength
-    } else break
-  }
+export function dedent(text: string): string {
+  const lines = text.split('\n')
+
+  const minIndentLength = lines.reduce((acc, line) => {
+    for (let i = 0; i < line.length; i++) {
+      if (line[i] !== ' ') return Math.min(i, acc)
```
brc-dd, line 11:
```suggestion
      if (line[i] !== ' ' && line[i] !== '\t') return Math.min(i, acc)
```

PR #1729, opened 2022-12-22
fix: [guide/markdown] found a bug

cklwblove:
I found that the `Input` in the document shows `[!code xx]` with an extra space, which would result in the use of copied code, which would not work in markdown. I fixed it.


brc-dd:
image

PR #1734, opened 2022-12-24
fix: refresh ads per page navigation

posva:
This is the same behavior we had before: https://unpkg.com/browse/vitepress@0.20.10/dist/client/theme-default/Layout.vue#L135

### DIFF
--- a/src/client/theme-default/components/VPCarbonAds.vue
+++ b/src/client/theme-default/components/VPCarbonAds.vue
@@ -3,16 +3,16 @@ import { ref, watch, onMounted } from 'vue'
 import { useData } from 'vitepress'
 import { useAside } from '../composables/aside.js'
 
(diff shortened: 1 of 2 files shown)


brc-dd:
IG we should better refresh using Carbon's API, because currently it is creating a bunch of script tags with the same ID:

image

Something like this in `VPCarbonAds.vue` should do the job:

```ts
watch([page, isAsideEnabled], () => {
  hasInitalized && isAsideEnabled.value && (window as any)._carbonads?.refresh()
})
```

WDYT?

posva:
You're right! I need to do some testing to see if the refresh works the same way and some extra checks

posva:
Done. I also renamed the variable because it had a typo

brc-dd (approved):

PR #1743, opened 2022-12-29
fix(build): handle `-` in title of code blocks with line highlighting

brc-dd:
closes #1740

Co-authored-by: Ron0115 <jayguess@qq.com>


ron0115:
greate job, thanks~

PR #1750, opened 2023-01-02
feat(build): don't hard fail on unknown languages in fences

clark-cui:
## bug
When a language that isn't registered by Shiki is used, page will be crashed and redict to 404.
截屏2023-01-02 16 54 36

## fix
use defaultLang when language isn't registered, and give a warning log.


clark-cui:
Is there any update? Do I need to provide more information about this?

brc-dd:
No, sorry, I was busy last few days. Will get this merged in 1-2 days. Thanks for the PR!

PR #1787, opened 2023-01-12
docs(gitlab pages): add base url configuration in the deploying section

Okamidev:
It appears that setting the `base` url is also needed for Gtilab Pages deployment.

It's basically a simple copy/paste of the Github Pages section, but it will probably help some people avoid looking for why assets are not loaded when deploying to Gitlab Pages.


kiaking (approved):

kiaking:
Thanks a lot!

kiaking (approved):

PR #1813, opened 2023-01-18
docs: add node alpine alternative to GitLab CI deploy

aendre:
I was using GitLab CI to build and deploy a site generated by Vitepress, but it failed because I have used the lighter image of node (`node:16-alpine`). Installing git on the alpine version solved the problem. I have added an alternative `.gitlab-ci.yml` config for those who want to use the alpine version. I hope it helps. Cheers!

### DIFF
--- a/docs/guide/deploying.md
+++ b/docs/guide/deploying.md
@@ -139,6 +139,25 @@ Don't enable options like _Auto Minify_ for HTML code. It will remove comments f
        - main
    ```
 
+4. Alternatively, if you want to use an _alpine_ version of node, you have to install `git` manually. In that case, the code above modifies to this:
+   ```yaml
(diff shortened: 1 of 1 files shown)


kiaking:
I'm not sure if we wanna keep adding this kind of stuff. Because it's obvious that you need git to check out the repo. If we keep doing this, we'll get more issues like for this image, you need to add `node`, and `yarn`, for this build you need to add this and that.

And the build is not related to VitePress really, it's about how node works. Users may use what ever image they want, but I'll say for VitePress doc, stick to most flexible image (node) and be done with it 👀

brc-dd:
Let's just merge this for now, I'll rewrite deploying docs later.

Issue #1049, opened 2022-07-23
Blur backdrop filter on navbar noticeably impacts scroll performance


vberlier:
### Describe the bug

The blurred background on the navbar looks nice but I noticed that it causes some lag when scrolling. I used the "Rendering" tab in the chrome devtools to measure the actual impact on performance and it looks like disabling the `backdrop-filter` property in the inspector improves the framerate by almost 30%.

This is a common problem that plagues the "frosted background" trend. If the content behind an element with a blurred backdrop updates rapidly (such as when scrolling), the page will start to stutter. This is even more noticeable on 4k displays where the blur filter has to process a lot more pixels.

![image](https://user-images.githubusercontent.com/24430071/180603883-9b99daf5-93e2-4658-9b06-88d61437101b.png)

| `backdrop-filter` enabled | `backdrop-filter` disabled |
| - | - |
| ![image](https://user-images.githubusercontent.com/24430071/180603945-4f370a68-3340-423d-93e9-6df938cd4dfe.png) | ![image](https://user-images.githubusercontent.com/24430071/180603925-de05287c-ce3b-466d-b24e-a1aab03d35fd.png) |

### Reproduction

To gather the above metrics I opened the official [vitepress website](https://vitepress.vuejs.org/guide/theme-introduction.html) and held the down arrow key to scroll continuously for a few seconds. For testing without the blur effect I disabled the `backdrop-filter` property on the `VPNavBar` component in the inspector.

### Expected behavior

This could be solved by removing the blur effect or only enabling it at low dpi using a media query. Another option would be to make it easy to disable the blur effect when extending the default theme.

### System Info

```shell
Chrome 103, Windows
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

gmqiyue:
has same problem on macos12.5/chrome 106

kiaking:
I've removed blur background due to design finalization at #1790.

PR #1816, opened 2023-01-20
fix(theme): adjust the position of the curtain to avoid block sidebar

CHOYSEN:
refs #1049 [#1709 ](https://github.com/vuejs/vitepress/pull/1790)

## reproduction

https://vitepress.vuejs.org/guide/what-is-vitepress

try collapse the first item of sidebar

### DIFF
--- a/src/client/theme-default/components/VPNavBar.vue
+++ b/src/client/theme-default/components/VPNavBar.vue
@@ -166,8 +166,7 @@ const { hasSidebar } = useSidebar()
     position: absolute;
     right: 0;
     bottom: -32px;
-    padding-left: var(--vp-sidebar-width);
-    width: 100%;
+    width: calc(100% - var(--vp-sidebar-width));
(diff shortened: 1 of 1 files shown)


brc-dd (approved):

CHOYSEN:
After this change, maybe the shadow(`background: linear-gradient(var(--vp-c-bg), transparent 70%)`) can be apply in `.curtain` directly instead of the `before` element. Did i miss anything :)

https://github.com/vuejs/vitepress/blob/48f0b015694c17767180eb2f75e9eb12c3f2a358/src/client/theme-default/components/VPNavBar.vue#L173-L179

brc-dd:
/cc @kiaking 

PS: if this 100% thing was intentional, then probably we can revert this and just add pointer-events none.

kiaking:
No, it wasn't! This was planned so all good 👍

Issue #1814, opened 2023-01-19
URL hash is not removed when navigating to direct URL


Andrew-web-coder:
### Describe the bug

URL hash is not removed when navigating to direct URL, see reproduction steps. 
Note: Vuepress works a bit differently, but, for example, Docusaurus works as expected

### Reproduction

1) Go to https://vitepress.vuejs.org/guide/getting-started
2) Scroll down a bit and click `#` anchor next to "Step. 1: Create a new project", observe as URL changes to `https://vitepress.vuejs.org/guide/getting-started#step-1-create-a-new-project`
3) Now, in the sidebar, click "Getting Started" - nothing happens

### Expected behavior

I would expect that page would reload or scroll to the top and URL hash would be removed.

### System Info

```shell
-
```

### Additional context

I could fix in the source code here - https://github.com/vuejs/vitepress/blob/main/src/client/app/router.ts#L172 - by adding another `} else { go(href) }`. But I am not sure that this is a proper fix, therefore I did not create a pull request, but I can provide, if asked.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

Jimmylxue:
#step-1-create-a-new-project is part of getting-started, so i think is not problem or bug then click getting-started nothing happens

Andrew-web-coder:
@Jimmylxue And what if I would give you link like https://vitepress.vuejs.org/guide/markdown#advanced-configuration and you would want to read the page from the top, you would then have to scroll to top like a caveman.

Jimmylxue:
You're right that's true, there is a better experience if can realize it.

PR #1829, opened 2023-01-21
feat: sliderBar link support clear hash

Jimmylxue:
# About PR
> thanks for checking
about [issue 1814](https://github.com/vuejs/vitepress/issues/1814) ,really other documents website support slider bar link to remove hash, so, i realize it. 

it's really improve user experience if can realize it, so, i hope this pr can be merged!😋


kiaking:
I think it's not good idea to remove `#` like this because users may want to have `#` in sidebar and that should be fine.

We should handle this in router level maybe?

brc-dd:
Anyway, the implementation is not right. Changing URL like this will make the whole site load again if someone clicks another sidebar item.

PR #1845, opened 2023-01-26
fix(theme): don't show border on navbar when sidebar is there

brc-dd:


brc-dd:
nope doesn't work 😂

brc-dd:
ok, now it does 😅

PR #1847, opened 2023-01-27
docs: update getting-started.md

jofftiquez:
Remove unnecessary dot (.) after every "Step" word. I don't know if this is intentional, but it doesn't look right to me. I could be wrong, so yeah. 

```diff
- # Step. 1
+ # Step 1
```


kiaking (approved):
yeah why not 👍  Thanks for the PR!

PR #1865, opened 2023-01-30
fix(theme): infer collapsible from collapsed

### LINKED ISSUE #1418: [sidebar] infer collapsible from collapsed and drop the former
can't we only have this property? and if collapsed != undefined we assume the toggle should show?, this way we don't need the collapsible property. 
I know we already have collapsible but collapsed? should cover all options with that logic. 

_Originally posted by @emersonbottero in https://github.com/vuejs/vitepress/pull/1361#discussion_r984184979_
      

---

Just to clarify:

```
collapsed == null -> not collapsible
collapsed === true -> collapsible and collapsed
collapsed === false -> collapsible but expanded
```

This will be a breaking change, but IG it'll a minor one and can be done before more people adopt collapsible sidebar group.

brc-dd:
⚠️  BREAKING CHANGES AHEAD

* change `collapsible: true, collapsed: true` to `collapsed: true`
* change rest of `collapsible: true` to `collapsed: false`

closes #1418

### DIFF
--- a/__tests__/e2e/.vitepress/config.ts
+++ b/__tests__/e2e/.vitepress/config.ts
@@ -4,7 +4,7 @@ const sidebar: DefaultTheme.Config['sidebar'] = {
   '/': [
     {
       text: 'Frontmatter',
-      collapsible: true,
+      collapsed: false,
       items: [
         {
           text: 'Multiple Levels Outline',
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -79,7 +79,7 @@ function sidebarGuide() {
   return [
     {
       text: 'Introduction',
-      collapsible: true,
+      collapsed: false,
       items: [
         { text: 'What is VitePress?', link: '/guide/what-is-vitepress' },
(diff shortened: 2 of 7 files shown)


kiaking (approved):

kiaking:
@brc-dd 👍 👍 👍  I guess when merging, adding BREAKING CHANGE comment will generate Changelog with braking change notice 👀  Please go ahead and merge this when you're ready 🙌

brc-dd:
Hi, where to put that breaking change? 😅 I tried these locally, none is working:

```
fix!(theme): ...
fix(theme)!: ...
BREAKING CHANGE: ...
BREAKING CHANGE: fix(theme): ...
```

kiaking:
Whaaat... those syntax should work... 🤔 Maybe we need to directly modify Changelog on release process?

brc-dd:
Okay, we need to add it in footer 😅 https://github.com/conventional-changelog-archived-repos/conventional-changelog-angular/blob/master/convention.md

Issue #1872, opened 2023-02-01
The value of the class of the pre tag is truncated


jerrywu001:
### Describe the bug

online demo:
https://stackblitz.com/edit/vitejs-vite-79ocfq?file=docs%2Findex.md

1. Register a public component. called `SandBox`.

```js
// docs/components/SandBox.vue
ctx.app.component('SandBox', SandBox);
```
2. Use markdown-it-direcitve-webcomponents plugin.

```js
import { createRequire } from 'module';
import { defineConfig } from 'vitepress';

const require = createRequire(import.meta.url);

export default defineConfig({
  markdown: {
    config(md) {
      md.use(require('markdown-it-directive')).use(
        require('markdown-it-directive-webcomponents'),
        {
          components: [
            {
              present: 'both',
              name: 'sandbox',
              tag: 'sand-box',
              allowedAttrs: ['readonly', 'closabletabs', 'template'],
              parseInner: true,
            },
          ],
        }
      );
    },
  },
});
```

3. So I can use it like this in .md.

<img width="605" alt="image" src="https://user-images.githubusercontent.com/12525415/215923598-a750300a-41b5-4649-97ac-e2e9952b498d.png">

4. The expect effect is as follows

https://vue3-npm-template.netlify.app/get-started/introduction.html#demo

<img width="700" alt="image" src="https://user-images.githubusercontent.com/12525415/215923868-05067303-ee50-453a-a6f2-f8220db4d778.png">

### But actually:

<img width="700" alt="image" src="https://user-images.githubusercontent.com/12525415/215924089-901a2c77-7005-4a3a-8821-0813704e2fb8.png">

Version `1.0.0-alpha.31` does not have this problem.

**This issue starts from `1.0.0-alpha.32`**.

See: https://github.com/vuejs/vitepress/commit/a684b67ec084fdc3b3a300ffbdd21e19fdcf7b1e#diff-caaa5859b69d40532a36e00cd217b000c7653293a148088a1dbc0275e8c56570

**Because The value of the class of the pre-tag is truncated**.

### Reproduction

See: https://stackblitz.com/edit/vitejs-vite-79ocfq?file=docs%2Findex.md

### Expected behavior

Expected to happen:

https://vue3-npm-template.netlify.app/get-started/introduction.html#demo

<img width="700" alt="image" src="https://user-images.githubusercontent.com/12525415/215923868-05067303-ee50-453a-a6f2-f8220db4d778.png">

[...]

PR #1873, opened 2023-02-01
fix: The value of the pre-tag class should be retained

jerrywu001:
#1872 

Have Tested.


brc-dd:
That `extractLang` was there for a reason. If you remove that (directly put `info` as class), the resultant class may look something like `language-ts [config.ts]` for code blocks, `language-ts:line-numbers` for line numbers. All these are undesirable and might break other functionalities (especially regexes) that depend on a stricter structure of output.

jerrywu001:
So.How can I solve this？ @brc-dd

PR #1879, opened 2023-02-01
docs: fix carbonAds

manchan4869:
- Remove incorrect hash symbols and make the right sidebar navigable to the correct text fragment
- Synchronize CarbonAds option type declarations from ·`default-theme.d.ts`
- Fix syntax issue
- Empty line at the end, forgot about last commit


kiaking (approved):
Thanks!

PR #1881, opened 2023-02-02
fix: VPNavBarMenuLink no external icon

meedfine:
i found externalIcon only show in VPMenuLink,   not show in VPNavBarMenuLink
why?   so i remove the noIcon bind
image

image

image

image


meedfine:
@brc-dd

brc-dd (approved):

PR #1893, opened 2023-02-04
docs(api): use code snippet for type definition

CHOYSEN:
The advantage is that the source code will be automatically synchronized to the document when it is modified, but some noise is added to the code.


kiaking:
Smart idea! But, not now. It wouldn't always be the same in terms of structure. For example, we might wanna write it like this.

```ts
import { Component } from 'vue'
import { PageData } from 'vitepress'

interface Route {
  path: string
  data: PageData
  component: Component | null
}
```

Issue #1692, opened 2022-12-14
Import Code Snippets Can you provide a relative path


tomiaa12:
### Is your feature request related to a problem? Please describe.

在文件目录过深时用@会不方便

### Describe the solution you'd like

<<< ./demo/index.ts

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

sky124380729:
that what i need also

yzqdev:
Good idea, maybe we could define our own alias like`@src`, `@theme`, `@guide`

chgeo:
Would like to have this feature as well.

Handpear:
any plans for release in future?

PR #1894, opened 2023-02-04
feat: support relative path for code snippet

CHOYSEN:
closes https://github.com/vuejs/vitepress/issues/1692


CHOYSEN:
Maybe we should resolve '/' to srcDir, so we may not need '@/'🤔

chgeo:
@CHOYSEN any plans to move this forward?  Would love to get this.  The current way of using absolute paths is a bit of an alien.

PR #1903, opened 2023-02-07
fix(theme-default): remove duplicate judgments in the preconnect function

manchan4869:
* The `preconnect()` has been judged before calling, so there is no need to repeat the judgment
 ```js
onMounted(() => {
  if (!theme.value.algolia) {
    return
  }
  preconnect()
...
}
```

### DIFF
--- a/src/client/theme-default/components/VPNavBarSearch.vue
+++ b/src/client/theme-default/components/VPNavBarSearch.vue
@@ -34,22 +34,20 @@ const preconnect = () => {
 
   const rIC = window.requestIdleCallback || setTimeout
   rIC(() => {
-    if (!theme.value.algolia || document.head.querySelector(`#${id}`)) return
-
     const preconnect = document.createElement('link')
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/client/theme-default/components/VPNavBarSearch.vue [hunk 1]
```diff
@@ -34,22 +34,20 @@ const preconnect = () => {
 
   const rIC = window.requestIdleCallback || setTimeout
   rIC(() => {
-    if (!theme.value.algolia || document.head.querySelector(`#${id}`)) return
```
brc-dd, line 37: The second one is necessary to prevent adding the preconnect element multiple times in head.

brc-dd: But IG it won't by triggered more than once, will it?

manchan4869: what I tested was that it only triggered once

PR #1906, opened 2023-02-08
docs: add linkage for `code-groups` in `getting-started`

manchan4869:
Set the selection of package manager in `getting-started` to be linked

It seems that only this page can be optimized in this way, so it is directly written in the`<script>`


brc-dd:
Just FYI, if someone is coming to this later, one can also directly use https://vitepress-plugins.sapphi.red/tabs/

brc-dd:
Gonna revert this, it actually breaks code groups on all the pages once you open index. Probably should have used onmounted and onunmounted here.

PR #1913, opened 2023-02-09
fix(theme): make features support line wrapping

you-hengh:
When I add Feature to my blog, the titles of some articles are too long. I wish I could use identifiers to wrap freely.

In fact, I redefined the style of the Feature and then needed to position the title, but I found that the title could not be wrapped when the title was too long, so I submitted this Pr

未命名 3


Jimmylxue:
I think it's maybe useful

PR #1916, opened 2023-02-10
fix 1915: aside always rendered even when outline is false

### LINKED ISSUE #1915: support disabling aside in themeConfig
### Describe the bug

Even when you set theme.outline = false vitepress still adds class has-aside and therefore content is unnecessary constrained to narrow space

### Reproduction

Set theme.outline to false to hide aside. Even thou it's not displayed it still ocupies space and therefore width of .VPDoc container is the same as with theme.outline

### Expected behavior

When aside is hidden it should free up space

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 5.57 GB / 15.80 GB
  Binaries:
    Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.11 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.2.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (109.0.1518.78)
```

### Additional context

_No response_

### Validations

dominikcz:
fixes #1915

### DIFF
--- a/src/client/theme-default/composables/sidebar.ts
+++ b/src/client/theme-default/composables/sidebar.ts
@@ -51,7 +51,7 @@ export function useSidebar() {
 
   const hasAside = computed(() => {
     return (
-      frontmatter.value.layout !== 'home' && frontmatter.value.aside !== false
+      frontmatter.value.layout !== 'home' && frontmatter.value.aside !== false && theme.value.outline !== false
     )
   })
 


kiaking:
This is intended. Aside contains more than outline.

PR #1925, opened 2023-02-11
feat(theme): support disabling aside globally

### LINKED ISSUE #1915: support disabling aside in themeConfig

dominikcz:
fix #1915

### DIFF
--- a/docs/config/theme-configs.md
+++ b/docs/config/theme-configs.md
@@ -158,10 +158,11 @@ export type SidebarItem = {
 ```
 
 ## aside
+
 - Type: `boolean`
 - Default: `true`
 
-By default aside container is always rendered (even if empty). Setting this value to `false` prevents rendering of aside container.
+Setting this value to `false` prevents rendering of aside container.
 
 ## outline
 
@@ -306,7 +307,7 @@ An option to support searching your docs site using [Algolia DocSearch](https://
 
 ```ts
 export interface AlgoliaSearchOptions extends DocSearchProps {
-   locales?: Record<string, Partial<DocSearchProps>>
+  locales?: Record<string, Partial<DocSearchProps>>
 }
 ```
 
--- a/docs/config/theme-configs.md
+++ b/docs/config/theme-configs.md
@@ -157,6 +157,12 @@ export type SidebarItem = {
 }
 ```
 
+## aside
+- Type: `boolean`
+- Default: `true`
+
+By default aside container is always rendered (even if empty). Setting this value to `false` prevents rendering of aside container.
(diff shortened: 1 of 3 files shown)


dominikcz:
I'm not sure if that last commit with logic makes it better. For me original logic, which I just extended was more legible as it clearly stated that aside is turned off in case of:
- home page
- explicitly setting `frontmatter.value.aside` to false (not true, null or undefined)
- explicitly setting `theme.value.aside`  to false (not true, null or undefined). 

Now for some reason you treat `frontmatter.value.aside` in a different way than `theme.value.aside` while they do behave in similar way: can be null, undefined, true or false. Don't get me wrong, both approaches will work correctly, for me original writing was just clearer as far as developer's intentions are considered.

There is also one issue that bothers me a bit: in documentation I stated that default value is `true`. It's not really the case s default value is `undefined`, but figured that it's easier for end-user to think about it as `true` because it behaves like `true`. Hope it's OK for you.

Anyway thank you for merging :)

brc-dd:
The logic for the change was that frontmatter should always override themeConfig.

> There is also one issue that bothers me a bit: in documentation I stated that default value is true. It's not really the case s default value is undefined, but figured that it's easier for end-user to think about it as true because it behaves like true. Hope it's OK for you.

Yeah that's fine. For some of the other things in the config, we are doing the same.

dominikcz:
> The logic for the change was that frontmatter should always override themeConfig. Your logic would not have worked if someone wants to "enable" aside selectively for a page via frontmatter (when they have disabled aside globally).

I see, wasn't ware of that. Thank you!

PR #1834, opened 2023-01-22
fix: custom titles of code snippets inside code groups

Raiondesu:
This PR fixes #1785, by allowing the following syntax when a snippet is placed in a code group:

```md
::: code-group

<<< @/snippets/snippet.js [snippet.js]

<!-- with custom tab titles -->
<<< @/snippets/snippet-with-region.js [snippet with region custom title]

<!-- filename is now grabbed by-default -->
<<< @/snippets/snippet.cs{1,2,4-6 c#}

:::
```

See updated docs for an example.

PR #1969, opened 2023-02-21
feat: add code-group title

lavecoral:
Add same custom title as custom container to code group

### DIFF
--- a/docs/guide/markdown.md
+++ b/docs/guide/markdown.md
@@ -679,6 +679,34 @@ You can also [import snippets](#import-code-snippets) in code groups:
 
 :::
 
+You can also custom title in code groups:
+
+**Input**
+
+```md
+::: code-group from `snippets/snippet-with-region.js`
+
+<!-- filename is used as title by default -->
+
+<<< @/snippets/snippet.js
+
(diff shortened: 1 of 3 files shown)


MarkusKeck:
Isn't this the same as #1834 

It got added in the latest version 1.0.0-alpha.47 (2023-02-20)
https://github.com/vuejs/vitepress/blob/main/CHANGELOG.md

brc-dd:
@MarkusKeck No, this one adds something like a caption to a code-block:

image

I am still not sure if we'll be adding this though. For images this might make sense, but for code groups/blocks I don't think this has any use case.

MarkusKeck:
@brc-dd Thanks for clarification, I also dont really see a usecase, if you want the whole filename you can just add it to the title.

MarkusKeck:
@lavecoral Some of the tests have failed. You can check this manually before pushing:

`pnpm check`

If something is not formatted correctly run: `prettier --check .`
After that `prettier --write <filename>`

lavecoral:
@brc-dd link that: https://docs.spring.io/spring-security/reference/migration/servlet/session-management.html

PR #1972, opened 2023-02-22
feat(theme): make prev/next links changeable

### LINKED ISSUE #1914: Make prev / next links changeable

Currently it is only possible to customize the text of the links, but not the link itself, this is taken from the sidebar. However, if a page does not exist in the sidebar, no prev / next link can be set or modified.

The current behavior causes the vitepress application to crash if the page is not in the sidebar and you set a prev link via frontmatter, because there is no prev link whose text you can change. As next link always the first page of the sidebar is used.

I think it would be good if you could customize not only the text but also the link.

image

Change the frontmatter options from 

```yml
---
prev: 'Get Started | Markdown'
---
```

to

```yml
---
prev:
    text: 'Get Started | Markdown'
    link: '/gettingstarted/markdown'
---
```

_No response_

### Additional context

```yml
// prev-next.d.ts
    prev: {
        text: any;
        link: string;
    };
    next: {
        text: any;
        link: string;
    };
```

### Validations

MarkusKeck:
close #1914
close https://github.com/vuejs/vitepress/discussions/1745

BREAKING CHANGE: The `prev` and `next` frontmatter options have been changed and are now nested.


MarkusKeck:
Only change the text of the link:

```yaml
---
prev:
  text: 'Getting Started'
---
```

Change the text and the link itself:

```yaml
---
prev:
  text: 'Markdown'
  link: '/guide/markdown'
---
```

If you don't want a link to be displayed on a specific page you can also disable it:

```yaml
---
prev:
  link: false
---
```

brc-dd (approved):

PR #1983, opened 2023-02-23
feat(theme): hero image now scales in all dimensions

### LINKED ISSUE #1982: VPHero image height does not scale for small viewports
### Describe the bug

The VPHero image does only set the 'max-width' inside the media querys but not 'max-height'.
This leads to an overlap of the hero image and the content that follows it.

https://pinia.vuejs.org/ had the same problem with its hero image, beeing 214px in width and 320px in height.
They solved it by adding `max-height` to their custom css.

```css
@media (min-width: 960px)
.VPHero .VPImage.image-src {
    max-height: 320px;
}
@media (min-width: 640px)
.VPHero .VPImage.image-src {
    max-height: 256px;
}
.VPHero .VPImage.image-src {
    max-height: 192px;
}
@media (min-width: 960px)
[data-v-be2bd16f] .image-src {
    max-width: 320px;
}
@media (min-width: 640px)
[data-v-be2bd16f] .image-src {
    max-width: 256px;
}
```
The better solution would be to just add it to the VPHero component.

image

### Reproduction

Take an image which is not cubical, so has more height than width and set it as hero image

```markdown
---
layout: home

hero:
      image:
        src: /assets/<image-name>.svg
        alt: alt-text
```

### Expected behavior

The VPHero image height should scale like the width.

image

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (32) x64 AMD Ryzen 9 3950X 16-Core Processor
    Memory: 36.46 GB / 127.82 GB
  Binaries:
    Node: 16.14.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (110.0.1587.50)    
  npmPackages:
    vitepress: ^1.0.0-alpha.13 => 1.0.0-alpha.47
```

### Additional context

_No response_

### Validations

MarkusKeck:
close #1982

### DIFF
--- a/src/client/theme-default/components/VPHero.vue
+++ b/src/client/theme-default/components/VPHero.vue
@@ -306,6 +306,7 @@ const heroImageSlotExists = inject('hero-image-slot-exists') as Ref<boolean>
   /*rtl:ignore*/
   left: 50%;
   max-width: 192px;
+  max-height: 192px;
   /*rtl:ignore*/
   transform: translate(-50%, -50%);
 }
@@ -313,12 +314,14 @@ const heroImageSlotExists = inject('hero-image-slot-exists') as Ref<boolean>
 @media (min-width: 640px) {
   :deep(.image-src) {
     max-width: 256px;
+    max-height: 256px;
   }
 }
 
 @media (min-width: 960px) {
   :deep(.image-src) {
     max-width: 320px;
+    max-height: 320px;
   }
 }
 </style>


MarkusKeck:
sry, should be fix not feat

brc-dd (approved):

PR #1989, opened 2023-02-24
feat(theme): enhance readability of custom containers (#1824)

### LINKED ISSUE #1824: Bad contrast for Custom Containers on light theme
### Describe the bug

Since the the theme update in [1.0.0-alpha.37](https://github.com/vuejs/vitepress/blob/main/CHANGELOG.md#100-alpha37-2023-01-17) the text to background contrast ration of [Custom Containers](https://vitepress.vuejs.org/guide/markdown#custom-containers) got worse. The text became much more difficult to read.

Here's a side-by-side comparison image:

comparison

### Reproduction

1. Go to https://vitepress.vuejs.org/guide/markdown#custom-containers
2. Switch to light theme

### Expected behavior

`--vp-custom-block-*-text` variables on light theme should be either set to `--vp-c-*-dark` or `--vp-c-*-darker` instead of `--vp-c-*`.

For Example: 

```css
:root {
  --vp-custom-block-warning-text: var(--vp-c-yellow-darker);
}

.custom-block.warning {
    color: var(--vp-custom-block-warning-text);
}
```

### System Info

```shell
-
```

### Additional context

_No response_

### Validations

kiaking:
close #1824

Increase contrast of custom containers. I've removed dimmed background and updated the yellow color to be more "amber" in order to have better contrast. It's extremely hard to have single color that works for both light and dark mode (especially yellow), but I think this is good enough?

It's kinda visually breaking change, but... it looks good though 👀

Screenshot 2023-02-24 at 20 47 42

Screenshot 2023-02-24 at 20 47 47

### DIFF
--- a/src/client/theme-default/styles/components/custom-block.css
+++ b/src/client/theme-default/styles/components/custom-block.css
@@ -58,7 +58,7 @@
 }
 
 .custom-block-title {
-  font-weight: 700;
+  font-weight: 600;
 }
 
 .custom-block p + p {
--- a/src/client/theme-default/styles/vars.css
+++ b/src/client/theme-default/styles/vars.css
@@ -29,11 +29,11 @@
   --vp-c-green-dimm-2: rgba(16, 185, 129, 0.2);
   --vp-c-green-dimm-3: rgba(16, 185, 129, 0.5);
 
-  --vp-c-yellow: #eab308;
-  --vp-c-yellow-light: #facc15;
-  --vp-c-yellow-lighter: #fde047;
-  --vp-c-yellow-dark: #ca8a04;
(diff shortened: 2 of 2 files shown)


kiaking:
OK let's go with this one for now. On light mode, the contrast score is still not above "A". However:

1. Sites like Tailwind also has similar score on light mode.

2. If we need to further increase the contrast, I think it's not possible by using the same color for both light/dark mode but rather we must have different colors for them. That is going to limit some customisability so I would like to avoid it.

Because of the above reason, I think it's fair to start with current contrast. If we get more feedback that we should increase the contrast, then I guess we should simply have different color sets for light and dark mode.

PR #1998, opened 2023-02-27
fix(theme): "copy code" button not readable on hover state (#819)

### LINKED ISSUE #819: copy to clipboard button style
### Describe the bug

If the `code sample` reached a bit long, the `copy button` and `copied tip` are not clear enough to recognize.

### Reproduction

image

### Expected behavior

`copy button` to be clear. 

### System Info

```shell
-
```

### Additional context

_No response_

### Validations

kiaking:
fix #819
fix #1892

This PR does 2 things.

1. Increase the code block height a bit and move "lang" label up to avoid lang label over wrapping with the code.
2. Adjust copy code button style so that it's more readable on hover style (#819).

Screenshot 2023-02-27 at 19 20 32

Screenshot 2023-02-27 at 19 20 39

Screenshot 2023-02-27 at 19 20 53

Screenshot 2023-02-27 at 19 21 05

### DIFF
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -468,8 +468,8 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  /*rtl:ignore*/
   border: 1px solid var(--vp-code-copy-code-hover-border-color);
+  /*rtl:ignore*/
   border-right: 0;
   border-radius: 4px 0 0 4px;
   width: 64px;
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -310,7 +310,7 @@
   position: relative;
   z-index: 1;
   margin: 0;
-  padding: 16px 0;
+  padding: 20px 0;
   background: transparent;
   overflow-x: auto;
 }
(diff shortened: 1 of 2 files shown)


src/client/theme-default/styles/components/vp-doc.css
```diff
@@ -460,11 +463,14 @@
...
   position: relative;
   /*rtl:ignore*/
+  top: -1px;
   left: -65px;
   display: flex;
   justify-content: center;
   align-items: center;
   /*rtl:ignore*/
+  border: 1px solid var(--vp-code-copy-code-hover-border-color);
+  border-right: 0;
```
brc-dd, line 473:
```suggestion
  border: 1px solid var(--vp-code-copy-code-hover-border-color);
  /*rtl:ignore*/
  border-right: 0;
```

src/client/theme-default/styles/components/vp-doc.css
```diff
@@ -460,11 +463,14 @@
 .vp-doc [class*='language-'] > button.copy:hover.copied::before {
   position: relative;
   /*rtl:ignore*/
+  top: -1px;
```
brc-dd, line 466:
```suggestion
  top: -1px;
  /*rtl:ignore*/
```

brc-dd:
Rest LGTM. Closes #1892 too IG?

kiaking:
Thanks! Ahhh #1892. I'll double check this to see if it works with this change 🤔

kiaking:
OK I've tweaked a bit and checked. Now the current one is not that bad but not perfect I would say. If we increase the top padding to 24px and then move the lang label up for another 2px looks perfect, when there is line hightlight on the first line.

But then the rest of code block with no line highlight looks a bit worse (in my opinion) 🤔 Top/Bottom padding would be too big, and lang label is too near the edge.

So, I say, having line highlight on the very first line should be rare, therefore the current stye is best for overall design 😳

brc-dd (approved):

PR #2011, opened 2023-02-28
Add `as` to Content component.

rgk:
I was looking to change the div to main, this would add a prop to do so.

### DIFF
--- a/src/client/app/components/Content.ts
+++ b/src/client/app/components/Content.ts
@@ -4,7 +4,8 @@ import { useRoute } from '../router.js'
 export const Content = defineComponent({
   name: 'VitePressContent',
   props: {
-    onContentUpdated: Function
+    onContentUpdated: Function,
+    as: { type: [Object, String], default: 'div' }
   },
   setup(props) {
     const route = useRoute()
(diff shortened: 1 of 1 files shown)


src/client/app/components/Content.ts
```diff
@@ -4,15 +4,16 @@ import { useRoute } from '../router.js'
 export const Content = defineComponent({
   name: 'VitePressContent',
   props: {
-    onContentUpdated: Function
+    onContentUpdated: Function,
+    as: { type: [Object, String], default: 'div' }
```
kiaking, line 8: I usually use `tag` for this type prop, but is there any standard for this? I've looked into Vuetify but couldn't find props that changes the element type 🤔 

It would be cool if we can follow some other ui libs for this kind of stuff to just align with them.

If we can't find it, I think we can go with `as`.

@brc-dd Happen to know any?

rgk: HeadlessUI uses `as` just like this, I thought it was standard at first.

brc-dd: Yeah, `as` is fine IG. It's more common in React land.

Issue #2021, opened 2023-03-02
'Switch Language' Button does not have an accessible name


sabicalija:
### Describe the bug

The button in the navigation for switching languages is not accessible for screen reader users.

![image](https://user-images.githubusercontent.com/13609768/222338409-311e1492-ffec-4a46-be1f-6f62b89dc58b.png)

Reviewing the HTML code reveals that property `aria-label` is not provided for the `button` element:

```html
<div class="VPFlyout VPNavBarTranslations translations">
  <button type="button" class="button" aria-haspopup="true" aria-expanded="false">
    <span class="text">
      <svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" class="option-icon" >
      <svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" class="text-icon">
    </span>
  </button>
  <div class="menu">
    <div class="VPMenu">
     <div class="items">
        <p class="title">Deutsch</p>
        <div class="VPMenuLink">
          <a class="VPLink link" href="...">English</a>
        </div>
      </div>
    </div>
  </div>
</div>
```

As far as I can tell, the reason for the missing `aria-label` property seems to be the missing `label` property for component `VPFlyout.vue` in the `VPNavBarTranslations.vue` component.

```html
<!-- src/client/theme-default/components/VPNavBarTranslations.vue -->
<template>
  <VPFlyout
    v-if="localeLinks.length && currentLang.label"
    class="VPNavBarTranslations"
    :icon="VPIconLanguages"
  >
    <div class="items">
      <p class="title">{{ currentLang.label }}</p>

      <template v-for="locale in localeLinks" :key="locale.link">
        <VPMenuLink :item="locale" />
      </template>
    </div>
  </VPFlyout>
</template>
```

I'm planning to provide a PR for this issue, if you agree. But I might need some guidance in how to provide the changes properly.

It seems like setting property `label` for component `VPFlyout` should resolve the issue. However, this value should be configurable for different locales IMO.

### Reproduction

Configure vitepress for 2 locales at least to enable the "Switch Language" button in the navbar. Using a screen reader, the button is not announced properly (e.g. "popup button collapse").

### Expected behavior

The `aria-label` property of the `button` element of `VPFlyout` in `VPNavBarTranslations` should be set.

```html
<!-- src/client/theme-default/components/VPNavBarTranslations.vue -->
<template>
  <VPFlyout
[...]

PR #2025, opened 2023-03-02
perf(a11y): add aria-label to language button

sabicalija:
close [#2021](https://github.com/vuejs/vitepress/issues/2021)


docs/guide/i18n.md
```diff
@@ -22,12 +22,14 @@ export default defineConfig({
   locales: {
     root: {
       label: 'English',
-      lang: 'en'
+      lang: 'en',
+      languageButtonAriaLabel: 'Change Language'
```
kiaking, line 26: It would be cool if we can have shorter name for this. Maybe `langMenuAriaLabel`?

types/shared.d.ts
```diff
@@ -85,5 +85,5 @@ export interface LocaleSpecificConfig<ThemeConfig = any> {
 
 export type LocaleConfig<ThemeConfig = any> = Record<
   string,
-  LocaleSpecificConfig<ThemeConfig> & { label: string; link?: string }
+  LocaleSpecificConfig<ThemeConfig> & { label: string; link?: string, languageButtonAriaLabel?: string }
```
brc-dd, line 88: Add this directly in ThemeConfig.

docs/guide/i18n.md
sabicalija: Thanks for the good suggestion. I've updated the commits accordingly.

types/shared.d.ts
sabicalija: Sorry, not sure if I understand you correctly. Are you referring to `types/default-theme.d.ts`?

If I understand your request correctly, you'd like to add this property for `themeConfig`. I think it's better to keep it as a _locale specific_ property, since labels should be different for every language, preferably.

sabicalija:
It would be nice, if the default configuration for the `deploy/netlify` deployment could include `locales` configuration to allow for testing features in that context.

sabicalija:
I've discovered related issues with the sidebar today. Probably it is relevant for other "group" and "dropdown" elements.

image
image

If you would merge fixes, I can prepare a PR or include fixes with this one. Pls just let me know. @kiaking @brc-dd

types/shared.d.ts
brc-dd: > Are you referring to types/default-theme.d.ts? If I understand your request correctly, you'd like to add this property for themeConfig. 

Yeah. 

> I think it's better to keep it as a locale specific property, since labels should be different for every language, preferably.

`themeConfig` is overridable inside locales. Refer the other labels that are present in that file. Things should not be kept at top level unless it's theme independent, which in your case is not true (Themes other than default theme might not respect that value at all.)

sabicalija: Ok. I thought it might not be suitable for `themeConfig`, since the settings would only be relevant if one has a `locales` configuration. Without other languages, the configuration has no effect.

I've updated the PR accordingly. Please have a look and let me know if I got everything right. Thanks!

brc-dd: Discard the changes in shared.d.ts, langs.ts, i18n.md. And update VPNavBarTranslations.vue to directly use `theme.langMenuAriaLabel`

sabicalija: Thanks! I've updated the commits.

I wonder if one can still change the `aria-label` for different locales this way. But if I understood you correctly, it should be possible nevertheless.

Hope the default value is ought to be set like that.

brc-dd: Yeah one can override them:

```ts

export default defineConfig({
  // shared properties and other top-level stuff...

  locales: {
    root: {
      label: 'English',
      lang: 'en',

      themeConfig: {
          // override labels here for english
      }
    },
    fr: {
      label: 'French',
      lang: 'fr', // optional, will be added  as `lang` attribute on `html` tag
      link: '/fr/guide', // default /fr/ -- shows on navbar translations menu, can be external

      // other locale specific properties...
      themeConfig: {
          // override labels here for french
      }
    }
  }
})

```

(We're working on updating the docs. Don't want conflicts there that's why leave that i18n.md unchanged for now.)

sabicalija: Yeah, I've seen it in the docs anyway. Just wasn't sure if there's something else that needs to be updated. Thanks!

PR #2032, opened 2023-03-04
Docs revamp

yyx990803:


yyx990803:
Note: this PR includes a number of features in addition to docs changes:

- (dynamic routes) `$params` template global
- (dynamic routes) `params` as root-level ref in `useData()`
- Add `isNotFound` in `PageData`
- deprecate `Theme.NotFound` and `Theme.setup`
- Add `Theme.extends`
- Pass files matched by `watch` to `loader()` in data loaders
- Add `defineLoader` type helper for data loaders
- Add `loadEnv` re-export from `vite`

PR #2040, opened 2023-03-06
perf(a11y): provide accessible header anchors

### LINKED ISSUE #2039: Header anchors not accessible to screen reader users.
### Describe the bug

The header anchors are not accessible to screen reader users.

```html
<h1 id="what-is-vitepress" tabindex="-1">
  What is VitePress?
  <a class="header-anchor" href="#what-is-vitepress" aria-hidden="true">#</a>
</h1>
```

Setting `aria-hidden` on focusable elements (that are not hidden from **all** users) fails test "[aria-hidden elements do not contain focusable elements](https://dequeuniversity.com/rules/axe/4.4/aria-hidden-focus?utm_source=lighthouse&utm_medium=devtools)".

I'm preparing a PR to fix this issue.

### Reproduction

No special configuration necessary. Header anchors are enabled by default.

This issue can be tested on the official [VitePress](https://vitepress.vuejs.org/guide/what-is-vitepress) site or any other basic vitepress deployment using the default-theme.

### Expected behavior

Provide an accessible default configuration for `markdown-it-anchor` and update styling of header-anchors accordingly.

A possible solution could look like this:

```html
<div class="header-wrapper">
  <h1 id="what-is-vitepress" tabindex="-1">What is VitePress?</h1>
  <a class="header-anchor" href="#what-is-vitepress" aria-label="Permalink to "What is VitePress?"">#</a>
</div>
```

A viable `markdown-it-anchor` configuration could look like this.
```js
{
  anchor: {
    permalink: MdItAnchor.permalink.linkAfterHeader({
      style: "aria-label",
      assistiveText: (title) => `Permalink to "${title}"`,
      visuallyHiddenClass: "visually-hidden",
      wrapper: ['<div class="header-wrapper">', "</div>"],
    }),
  },
}
```

Additionally, some updates for anchor styling are necessary too.
```css
.vp-doc .header-wrapper {
  position: relative;
}

.vp-doc .header-wrapper .header-anchor {
    line-height: 32px;
    position: absolute;
    top: 0;
}

.vp-doc .header-wrapper h1 + .header-anchor {
    line-height: 40px;
    font-size: 32px;
}

.vp-doc .header-wrapper h2 + .header-anchor {
    padding-top: 25px;
    line-height: 32px;
    font-size: 24px;
}

.vp-doc .header-wrapper h3 + .header-anchor {
    line-height: 28px;
    font-size: 20px;
}

.vp-doc .header-anchor:hover,
.vp-doc .header-anchor:focus,
.vp-doc h1:hover + .header-anchor,
.vp-doc h2:hover + .header-anchor,
.vp-doc h3:hover + .header-anchor,
.vp-doc h4:hover + .header-anchor,
.vp-doc h5:hover + .header-anchor,
.vp-doc h6:hover + .header-anchor {
  opacity: 1;
}
```

### System Info

```shell
System:
  OS: Linux 5.15 Ubuntu 20.04.5 LTS (Focal Fossa)
  CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1505M v5 @ 2.80GHz
  Memory: 19.26 GB / 31.13 GB
  Container: Yes
  Shell: 5.0.17 - /bin/bash
Binaries:
  Node: 18.14.0 - ~/.nvm/versions/node/v18.14.0/bin/node
  Yarn: 1.22.19 - ~/.nvm/versions/node/v18.14.0/bin/yarn
  npm: 9.3.1 - ~/.nvm/versions/node/v18.14.0/bin/npm
Browsers:
  Chrome: 110.0.5481.177
  Chromium: 110.0.5481.100
  Firefox: 110.0.1
```

### Additional context

More details about test "aria-hidden elements do not contain focusable elements" can be found at [https://dequeuniversity.com/rules/axe/4.4/aria-hidden-focus](https://dequeuniversity.com/rules/axe/4.4/aria-hidden-focus).

### Validations

sabicalija:
close #2039


brc-dd:
+1, it will break @vuejs/theme too 👀

sabicalija:
Yes, but...

Thank you for raising this question. It's actually quite intricate and I'll try to sum ti up to allow for further discussion (with links to more details). I can provide more details (and tests) depending on which direction we head to.

Are you referring to `h1` exclusively or did you mean all headers (`h1`, `h2`, etc.)?
Do you think about already existing styles would break or would it be problematic with this variant in general?

`markdown-it-anchor` provides four different [permalink](https://github.com/valeriangalliat/markdown-it-anchor#permalinks) (a.k.a. anchor link) styles with the latest releases (since v8.0.0) and a possiblity to create custom permalinks.

* [Header Link](https://github.com/valeriangalliat/markdown-it-anchor#header-link)
* [Link after header](https://github.com/valeriangalliat/markdown-it-anchor#link-after-header) (initial proposal)
* [Link inside header](https://github.com/valeriangalliat/markdown-it-anchor#link-inside-header)
* [ARIA hidden](https://github.com/valeriangalliat/markdown-it-anchor#aria-hidden)
* [Custom permalink](https://github.com/valeriangalliat/markdown-it-anchor#custom-permalink)

`Link inside header` and `ARIA hidden` have accessibility issues, therefore I would not consider them further.

`Header Link` would provide what you asked for:

```html
<h1 id="what-is-vitepress" tabindex="-1"><a class="header-anchor" href="#what-is-vitepress">What is VitePress?</h1>
```

The only problem with this variant is, that you cannot use any links inside the "heading".

`Link after header` looks like this:

```html
<div class="header-wrapper">
  <h1 id="what-is-vitepress" tabindex="-1">What is VitePress?</h1>
  <a class="header-anchor" href="#what-is-vitepress" aria-label="Permalink to &quot;What is VitePress?&quot;">#</a>
</div>
```

[valeriangalliat](https://github.com/valeriangalliat/) (package author) is using style `Header Link` himself and considers this an accessible variant. He also mentioned that this style was used with [MDN](https://developer.mozilla.org/en-US/docs/Web) and [HTTP Archive](https://httparchive.org/reports/state-of-the-web).

I've reviewed the current semantics of MDN. This is how it looks right now:

```html
<h2 id="try_it"><a href="#try_it">Try it</a></h2>
```

It (still) is just like permalink style `Header Link`.

However, MDN does not use permalinks with `h1` elements. Only with `h2`, `h3`, etc.

Additionally, the overall semantics of a page look slightly different, which would be even a more accessible variant. Any chances of using similar semantics with `theme-default`?

```html
<main id="content">
  <article lang="en-US">
    <h1>The HTML Section Heading elements</h1>
    <section aria-labeledby="try_it">
      <h2 id="try_it"><a href="#try_it">Try it</a></h2>
      <!-- ... -->
    </section/>
    <!-- ... -->
  </article>
  <!-- ... -->
</main>
```

This is the semantic of VitePress rendered pages at the moment:

```html
<main>
  <div>
    <h1>
      What is VitePress?
      <a class="header-anchor" href="#what-is-vitepress" aria-hidden="true">#</a>
    </h1>
    <!-- ... -->
    <h2>
      Use Cases
      <a class="header-anchor" href="#use-cases" aria-hidden="true">#</a>
    </h2>
    <!-- ... -->
  </div>
  <!-- ... -->
</main>
```

I guess it would require also changing some of the rendering by `markdown-it` and introduce more problems in terms of styles. But, the theme would be more accessible for everyone and since VitePress is used for a lot of documentation, many people would benefit from it.

sabicalija:
> Can we change it to something like what Docusaurus is doing? Example here https://docusaurus.io/docs
> 
> Screenshot 2023-03-13 at 5 29 22 PM
> 
> This would preserve the same DOM structure as we have now. They use `:before` + CSS to render the `#` instead of `aria-hidden`, maybe that's the trick?

I think we can achieve this relatively easily with the [Custom Renderer](https://github.com/valeriangalliat/markdown-it-anchor#custom-permalink) and I'll update the PR, after we find the best option.

But I don't think, that it is that relevant, if the `#` is rendered using `:before` or simply as the content of the anchor element.

The problem how `aria-hidden` is used right now, is, that there is content (namely, the anchor link) that is _hidden_ from users with visual impairments.

```html
<main>
  <div>
    <h1>
      What is VitePress?
      <a class="header-anchor" href="#what-is-vitepress" aria-hidden="true">#</a>
    </h1>
    <!-- ... -->
    <h2>
      Use Cases
      <a class="header-anchor" href="#use-cases" aria-hidden="true">#</a>
    </h2>
    <!-- ... -->
  </div>
  <!-- ... -->
</main>
```

A simple `aria-label` describing this link, instead of `aria-hidden`, which hides content from people with visual impairment, would already be sufficient and avoid excluding a certain user group.

```html
<main>
  <div>
    <h1>
      What is VitePress?
      <a class="header-anchor" href="#what-is-vitepress" aria-label="Permalink to &quot;What is VitePress?&quot;">#</a>
    </h1>
    <!-- ... -->
    <h2>
      Use Cases
      <a class="header-anchor" href="#use-cases" aria-label="Permalink to &quot;Use Cases&quot;">#</a>
    </h2>
    <!-- ... -->
  </div>
  <!-- ... -->
</main>
```

However, the best semantics are provided by MDN. I totally understand if you don't want to introduce breaking changes, but still I'd like to ask, if this would be an option. I guess it is still easier at this point, instead of introducing breaking changes at a later stage (after alpha).

I'm willing to provide PRs for all changes necessary (of course with some guidance by you).

sabicalija:
> Can we change it to something like what Docusaurus is doing? Example here https://docusaurus.io/docs
> 
> Screenshot 2023-03-13 at 5 29 22 PM
> 
> This would preserve the same DOM structure as we have now. They use `:before` + CSS to render the `#` instead of `aria-hidden`, maybe that's the trick?

You're suggestion was perfect! Yes, using `:before` would be better for rendering the `#`.

With the way Docusaurus solved it, the screen reader only reads the heading without the mention of `#`. But when anchor `#` has focus (e.g. by keyboard navigation using `Tab`), it's described correctly.

So this would be a perfectly accessible variant. I'll update this PR ASAP.

However, I'd argue that the MDN variant is semantically better/richer. Maybe something to keep in mind for the future.

sabicalija:
I've updated the PR with the proposed solution.

It's not made using a custom renderer but with the variant "Link After Header", although a bit hacky. But it's the best solution I could came up with, still respecting all possible user configurations (i hope :crossed_fingers:).

yyx990803:
There is one failing test `__tests__/e2e/home.test.ts` - the test case is expecting the extra `#` which is now properly hidden.

LGTM once we fix the test!

PR #2056, opened 2023-03-09
feat(theme): move to css logical properties

azat-io:
More info:
https://elad.medium.com/new-css-logical-properties-bc6945311ce7


yyx990803:
Thanks for the PR, but this seems like a rather significant change for little perceived benefits. Compatibility issues aside, contributors are likely more familiar with transitional properties.

PR #2059, opened 2023-03-09
fix(code-groups): fixes code-groups when using shiki-twoslash

### LINKED ISSUE #2066: Bug: Code-groups don't work correctly when using `markdown-it-shiki-twoslash`
### Describe the bug

Using code groups with `markdown-it-shiki-twoslash`

The current problem is that when finding the next element to set as active this `group?.querySelectorAll('div[class*="language-"]')` will grab the language-id class that gets injected by the twoslash lsp.

The fix should be simple enough by creating a filter that removes these classes before assigning the next element as active.

### Reproduction

Here is a example of this problem: [codesandbox](https://codesandbox.io/p/sandbox/determined-williams-y1s7pk?file=%2Fdocs%2Findex.md&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A25%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A25%7D%5D)

### Expected behavior

Code groups should toggle correctly.

### System Info

```shell
System:
    OS: Linux 6.1 Fedora Linux 36 (Workstation Edition)
    CPU: (16) x64 AMD Ryzen 7 5700X 8-Core Processor
    Memory: 22.70 GB / 31.22 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 19.3.0 - ~/.nvm/versions/node/v19.3.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v19.3.0/bin/yarn
    npm: 9.2.0 - ~/.nvm/versions/node/v19.3.0/bin/npm
  Browsers:
    Brave Browser: 110.1.48.164
    Chrome: 110.0.5481.100
    Firefox: 110.0
```

### Additional context

Opened a [PR](https://github.com/vuejs/vitepress/pull/2059) that should solve this problem.

### Validations

Raiden1411:
fixes #2066

### Problem: 
Using code groups with `markdown-it-shiki-twoslash`

This PR aims to fix this behavior. 

The current problem is that when finding the next element to set as active this `group?.querySelectorAll('div[class*="language-"]')` will grab the `language-id` class that gets injected by the twoslash lsp.

The fix should be simple enough by creating a filter that removes these classes before assigning the next element as active.

Here is a example of this problem: [codesandbox](https://codesandbox.io/p/sandbox/determined-williams-y1s7pk?file=%2Fdocs%2Findex.md&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A25%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A25%7D%5D)


brc-dd:
can you update it to directly use `:not` in the query selector?

Raiden1411:
I have added your suggestion and it also fixes the problem.

I think this would only become a problem if there is a actual language called `id` and this fix would filter that out.  I was also afraid that this could cause conflicts but so far I haven't seen any problems in the other projects that I've used this fix.

PR #2069, opened 2023-03-11
Fix/ssr hydrate

Reinhard2019:
**Problem:**

When you refresh the example page of the Vue docs web page, its left navigation is not highlighted.

The reason is because SSR Hydrate did not update the class correctly.

Here is the vue docs page: [https://cn.vuejs.org/examples/#hello-world](https://cn.vuejs.org/examples/#hello-world)


yyx990803:
Vue docs is not using the default theme. It uses a custom theme here: https://github.com/vuejs/theme

The issue also only occurs because the example page uses client-side hash to navigate between steps.

PR #2076, opened 2023-03-13
fix: theme.extends should cover default theme

jellyfishgh:
theme.extends is custom theme config.it should cover the default theme.


yyx990803:
The original code is working as intended.

PR #2100, opened 2023-03-17
fix: improve a11y

Akryum:


jd-solanki:
Does this bring back, Good old container styles?

image

Akryum:
They look like this: https://deploy-preview-2100--vitepress-docs.netlify.app/guide/markdown#custom-containers

PR #2110, opened 2023-03-20
feat: offline search

### LINKED ISSUE #670: Offline search functionality
ref #733

We are using VitePress for internal documentation. Especially for larger projects such as our entire knowledge base, it can be hard to find exactly what you're looking for. A search bar would greatly help out. An Algolia search bar is already available, but since our docs are hosted on a private network and for internal use only, this is not an option for us.
 
From Algolia DocSearch documentation:
> Your website must be publicly available. We do not host search indices for websites that are only available after authentication or are hosted on a private network.

A search bar that works offline, or at least on a private network. I believe VuePress [has a plugin](https://vuepress.vuejs.org/plugin/official/plugin-search.html#install), but the serach bar shows "Powered by Algolia" so I'm not sure.

- Hosting our docs publicly: not an option.
- Using VuePress or a different static site generator altogether, but since our apps are built on top of Vite we'd rather not.
- Using VuePress Next/V2 since it seems to have a local search option, and supports Vite as well.

### Additional context

_No response_

### Validations

Akryum:
Fix #670 

- [x] Generate index during dev
- [x] Generate index during build
- [x] Update index on file change
- [x] Search UI
- [x] Keyboard shortcuts
- [x] Translatable help text
- [x] Match highlighting
- [x] Mobile UI
- [ ] Dynamic routing
- [x] UI Localization
- [x] Filter by locale
- [x] Documentation

image


Akryum:
> Another thing, does it filter results based on the current locale?

Working on it!

src/node/plugins/offlineSearchPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+          }
+        ]
+      })
+    }
+  }
+
+  function getDocId(file: string) {
+    let id = file.replace(siteConfig.srcDir, siteConfig.userConfig.base ?? '')
+    id = id.replace(/\.md$/, siteConfig.cleanUrls ? '' : '.html')
+    return id
```
chgeo, line 103: While testing I figured 2 small issues.  The plugin 
- doesn't work w/ [route rewrites](https://vitepress.dev/guide/routing#route-rewrites)
- produces a double `//` in the `id`

The code below fixes both issues

```suggestion
    let relFile = path.relative(siteConfig.srcDir, file)
    relFile = siteConfig.rewrites.map[relFile] || relFile
    let id = path.join(siteConfig.userConfig.base ?? "", relFile);
    id = id.replace(/\.md$/, siteConfig.cleanUrls ? "" : ".html");
    return id;
```

package.json
```diff
@@ -152,6 +152,7 @@
     "markdown-it-emoji": "^2.0.2",
     "micromatch": "^4.0.5",
     "minimist": "^1.2.8",
+    "minisearch": "^6.0.1",
```
brc-dd, line 155: IG this needs to be in dependencies as it is being referenced by client side code (the search will likely break if someone installs just vitepress, it's working on the deploy preview because there we install everything).

docs/reference/default-theme-search.md
```diff
@@ -1,5 +1,60 @@
...
+    offlineSearch: true
+  }
+})
+```
+
+Example result:
+
+![screenshot of the search modal](/search.png)
+
+Alternatively, you can use [Algolia DocSearch](#algolia-search) or some community plugins like <https://github.com/emersonbottero/vitepress-plugin-search>.
```
brc-dd, line 21:
```suggestion
Alternatively, you can use [Algolia DocSearch](#algolia-search) or some community plugins like <https://www.npmjs.com/package/vitepress-plugin-search> or <https://www.npmjs.com/package/vitepress-plugin-pagefind>.
```

brc-dd:
rest of the stuff looks fine to me 🚀

PR #2133, opened 2023-03-25
feat: smooth scroll and preferences

userquin:
I notice preferences not using i18n, maybe we can add i18n later...

/cc @kiaking @brc-dd


userquin:
I need to review Firefox, it is not working...

Paper-Folding:
Wow, I was just about to post an issue until I see this pr, I think I really need it, could you please merge this as soon as possible? ❤️  Meanwhile, bootstrap also got a [smooth scroll](https://getbootstrap.com/docs/5.3/components/scrollspy), maybe you can take a look. Thank you!

kiaking:
I don't think having this button makes sense 🤔 What is the idea behind this feature? Never seen this in any site. Do you have any example?

PR #2135, opened 2023-03-25
feat: more flexible `ignoreDeadLinks`

antfu:
Sometimes user might have custom routes, custom subpath site, etc, which is out of the awareness of VitePress to consider as "Dead link". The current `false | true | 'localhostLinks'` option is a bit too extreme. Forcing those cases to disable the dead links check completely.

This PR adds more fine-grain control over them to make exceptions.

### DIFF
--- a/src/node/plugin.ts
+++ b/src/node/plugin.ts
@@ -70,7 +70,6 @@ export async function createVitePressPlugin(
     vue: userVuePluginOptions,
     vite: userViteConfig,
     pages,
-    ignoreDeadLinks,
     lastUpdated,
     cleanUrls
   } = siteConfig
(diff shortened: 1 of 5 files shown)


src/node/plugin.ts
```diff
@@ -195,7 +195,7 @@ export async function createVitePressPlugin(
     },
 
     renderStart() {
-      if (hasDeadLinks && !ignoreDeadLinks) {
+      if (hasDeadLinks && ignoreDeadLinks !== true) {
```
brc-dd, line 198: This can be removed IG 👀  (since you're now handling this in `shouldIgnoreDeadLink` itself). `hasDeadLinks` will be false.

```suggestion
      if (hasDeadLinks) {
```

brc-dd (approved):

PR #2138, opened 2023-03-25
Closes #2035: Adds left aside option

### LINKED ISSUE #2035: [Outline] Add support for left or right positioned Outline

I would like to move the Outline to the left of the main content which can be done relatively easily with CSS however the caveat is because when we use frontmatter to hide the outline on certain pages it creates a gap on the left side instead of completely hiding the Outline’s parent container.

Two things:

1. Theme Config to have something like “outlinePosition” that takes “left” or defaults to “right” and positions the outline correctly.
2. Change the logic so that when frontmatter has outline set to hidden for a page that the entire Outline element is removed instead of it’s child elements which will prevent the left aligned Outline from creating a left side gap.

Using CSS to move the Outline element but it not fully work with frontmatter setting an outline to false on a specific page by creating an empty gap because the parent element of Outline is not removed.

### Additional context

_No response_

### Validations

JohnCampionJr:
Made it as unobtrusive as I could think of.  Simply set aside to left in frontmatter and it changes.

### DIFF
--- a/src/client/theme-default/composables/sidebar.ts
+++ b/src/client/theme-default/composables/sidebar.ts
@@ -49,6 +49,12 @@ export function useSidebar() {
     )
   })
 
+  const leftAside = computed(() => {
+    if (hasAside && frontmatter.value.aside != null)
+      return frontmatter.value.aside === 'left'
+    return false
+  })
+
   const hasAside = computed(() => {
     if (frontmatter.value.layout === 'home') return false
     if (frontmatter.value.aside != null) return !!frontmatter.value.aside
@@ -80,6 +86,7 @@ export function useSidebar() {
     sidebarGroups,
     hasSidebar,
     hasAside,
+    leftAside,
     isSidebarEnabled,
     open,
     close,
--- a/src/client/theme-default/components/VPDoc.vue
+++ b/src/client/theme-default/components/VPDoc.vue
@@ -7,7 +7,7 @@ import VPDocFooter from './VPDocFooter.vue'
 import VPDocOutlineDropdown from './VPDocOutlineDropdown.vue'
 
 const route = useRoute()
-const { hasSidebar, hasAside } = useSidebar()
+const { hasSidebar, hasAside, leftAside } = useSidebar()
 
 const pageName = computed(() =>
   route.path.replace(/[./]+/g, '_').replace(/_html$/, '')
@@ -20,7 +20,7 @@ const pageName = computed(() =>
     :class="{ 'has-sidebar': hasSidebar, 'has-aside': hasAside }"
   >
     <div class="container">
-      <div v-if="hasAside" class="aside">
+      <div v-if="hasAside" class="aside" :class="{'left-aside': leftAside}">
         <div class="aside-curtain" />
         <div class="aside-container">
           <div class="aside-content">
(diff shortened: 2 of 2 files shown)


JohnCampionJr:
I couldn't find anywhere in the docs to update this one, so I fixed up the types instead

types/default-theme.d.ts
```diff
@@ -42,10 +42,11 @@ export namespace DefaultTheme {
 
     /**
      * Set to `false` to prevent rendering of aside container.
+     * Set to `left` to move aside to left of content.
      *
      * @default true
      */
-    aside?: boolean
+    aside?: boolean | string
```
brc-dd, line 49:
```suggestion
    aside?: boolean | 'left'
```

brc-dd (commented):
Added some suggestions. Let's make the type stricter.

src/client/theme-default/composables/sidebar.ts
```diff
@@ -49,6 +49,12 @@ export function useSidebar() {
     )
   })
 
+  const leftAside = computed(() => {
+    if (hasAside && frontmatter.value.aside != null)
+      return frontmatter.value.aside === 'left'
```
brc-dd, line 54: It should be something like this IG, can you test this locally? (Also run prettier once before committing)

```suggestion
    if (hasAside)
      return frontmatter.value.aside == null ? theme.value.aside === 'left' : frontmatter.value.aside === 'left'
```

CASES:

1. `aside` is set to `left` in `themeConfig`:
   1. if `aside` is not set in frontmatter, then render it on left
   3. if `aside` is set to `true` in frontmatter, then render it on right
   5. if `aside` is set to `false` in frontmatter, then don't render it
2. `aside` is set to `true` or not specified in `themeConfig`:
   1. if `aside` is not set in frontmatter, then render it on right
   2. if `aside` is set to `false` in frontmatter, then don't render it
   3. if `aside` is set to `left` in frontmatter, then render it on left
3. `aside` is set to `false` in `themeConfig`:
   1. if `aside` is not set in frontmatter, then don't render it
   2. if `aside` is set to `left` in frontmatter, then render it on left
   3. if `aside` is set to `true` in frontmatter, then render it on right

JohnCampionJr:
Tested locally :)

brc-dd (approved):

JohnCampionJr:
Improved the docs and the types to clearly explain what happens :)

docs/reference/default-theme-config.md
```diff
@@ -166,11 +166,13 @@ export type SidebarItem = {
 
 ## aside
 
-- Type: `boolean`
+- Type: `boolean | 'left'`
 - Default: `true`
 - Can be overridden per page via [frontmatter](./frontmatter-config#aside)
```
brc-dd, line 171: Cool, can you update `docs/reference/frontmatter-config.md` too?

docs/reference/default-theme-config.md
```diff
@@ -166,11 +166,13 @@ export type SidebarItem = {
...
 
 Setting this value to `false` prevents rendering of aside container.
+Setting this value to `true` renders the aside to the right.
+Setting this value to `left` renders the aside to the left.
```
brc-dd, line 175:
```suggestion
Setting this value to `false` prevents rendering of aside container.\
Setting this value to `true` renders the aside to the right.\
Setting this value to `left` renders the aside to the left.
```

(markdown doesn't preserve single line breaks)

haiting:
Is that feature release ？

brc-dd:
@haiting Yeah it will be there in alpha.64

haiting:
Thx! And do you mind if I know the schedule?

brc-dd:
2-3 days

PR #2140, opened 2023-03-26
feat: add `filePath` to `PageData`

### LINKED ISSUE #2088: `editLink` should be able to use the link of the route `rewrites` keys.

I use `editLink` for my GitHub repo URL. However, I also use the `rewrites` to rewrite my route. The edit link will link to the rewritten link which can't find the edit page.

I think the `:path` in `editLink` should use the original path before rewriting by default or add a boolean key to decide whether to use it or not.

Or add another mapping object to rewrite the path like the `rewrites` object.

### Additional context

_No response_

### Validations

chgeo:
In the edit link pattern function, the current `relativePath` (from `PageData`) is not a file path and hence does not allow to distinguish paths from different packages.
Add a new `relativeFilePath` string for this.  Keep `relativePath` for compatibility reasons.

Fixes #2088


chgeo:
@brc-dd any chance this get merged soon?  In our multi-package deployment of Vitepress, the _edit_ button doesn't work for half of the pages.

brc-dd:
> Keep relativePath for compatibility reasons.

I removed it actually. There aren't many people using this feature yet, it would be safer to drop this now instead of adding to the confusion in longer run.

Issue #2148, opened 2023-03-27
`lastUpdated` timestamp does not work for Git submodules


chgeo:
### Describe the bug

For file paths that lead inside a Git submodule, the lastUpdated timestamp is not set.

Reason is that `git log` does not deliver data if called with such a submodule path.

### Reproduction

Create a project w/ a Git submodule

### Expected behavior

Timestamps shall work for such paths.

### System Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 148.63 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
    npm: 8.19.3 - ~/.nvm/versions/node/v18.13.0/bin/npm
  Browsers:
    Chrome: 111.0.5563.110
    Firefox: 106.0.4
    Safari: 16.3
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

chgeo:
See fix: https://github.com/vuejs/vitepress/pull/2149

PR #2149, opened 2023-03-27
Fix `lastUpdated` to work w/ Git submodules

chgeo:
The previous approach was broken as `git log` does not deliver data if called with a path that leads inside a submodule.

The fix is to set the file's dirname as `cwd` and call git w/ just the file name.  This works in all circumstances.

Fixes https://github.com/vuejs/vitepress/issues/2148


src/node/markdownToVue.ts
```diff
@@ -187,7 +187,7 @@ export async function createMarkdownToVueRenderFn(
     }
 
     if (includeLastUpdatedData) {
-      pageData.lastUpdated = await getGitTimestamp(file)
+      pageData.lastUpdated = await getGitTimestamp(fileOrig)
```
chgeo, line 190: Note that I had to change this to get the true/un-rewritten file path.

PR #2152, opened 2023-03-27
fix: redirect locale link with hash

CHOYSEN:


brc-dd:
Hey, also consider adding some option to [`i18nRouting`](https://vitepress.dev/reference/default-theme-config#i18nrouting) like `'notForHash'` because because people may not care to override anchor links when they are translating.

Let me explain on that, if in a language `foo`, `bar` means `baz`, then if someone has code like this:

```md
<!-- index.md -->

# bar

here the anchor will be created as `#bar`
```

```md
<!-- foo/index.md -->

# baz

here the anchor will be created as `#baz` unless the user does something like this:

# baz {#bar}
```

Also that hash might need encoding/decoding (https://github.com/vuejs/vitepress/compare/main...feat/hash). Maybe check how it works with CJK characters?

CHOYSEN:
Ohh, you are right. I have been browsing the "configuration" page mostly, making me mistakenly believe that the anchor points for each language will be the same. I decided to close this PR, thanks!

brc-dd:
@CHOYSEN We can actually go ahead with this PR. Because *ideally* those anchors should be maintained to be same. Most of the projects that our teams are maintaining, they are taking care of that. But we should give an option to disable that if someone doesn't care.

CHOYSEN:
Sounds great, but I don't have time until the next two weeks, anyone is welcome to complete this!

PR #2162, opened 2023-03-29
styel: add favicon in docs

fyzhu:
use favicon of vite like vuepress do


brc-dd:
We are working on the rebranding; will add new logos and favicon, etc. once they are decided.

PR #2163, opened 2023-03-29
feat(build): provide `siteConfig` in `transformPageData` context

JohnCampionJr:
Similar to buildEnd, allows config options access in transformPageData

### DIFF
--- a/docs/reference/site-config.md
+++ b/docs/reference/site-config.md
@@ -552,18 +552,18 @@ export default {
 
 ### transformPageData
 
-- Type: `(pageData: PageData) => Awaitable<Partial<PageData> | { [key: string]: any } | void>`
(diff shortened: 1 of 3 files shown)


JohnCampionJr:
PS if there's any way you could merge this for the next release, I'd be very thankful :)

brc-dd:
siteConfig wasn't added because you'll anyway be writing that hook in your config file, so you can just access the variables from there?

```ts
const config = defineConfig({
  // ...
  async transformPageData(pageData) {
    console.log(config.something)
  }
})

export default config
```

Or maybe if you don't wanna self-reference the config, just store what you want to access in a variable?

```ts
const foo = 'something'

export default defineConfig({
  title: foo,

  async transformPageData(pageData) {
    console.log(foo)
  }
})
```

brc-dd (approved):

PR #2167, opened 2023-03-30
fix(theme): fix copy button display

btea:
image

### DIFF
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -426,7 +426,7 @@
   /*rtl:ignore*/
   right: 12px;
   z-index: 3;
-  display: block;
+  display: flex;
   justify-content: center;
   align-items: center;
   border: 1px solid var(--vp-code-copy-code-border-color);


brc-dd:
Those justify/align center need to be removed not display block. This is what happens with your PR (before it everything was working fine although those rules are redundant):

image

btea:
Maybe `display`, `justify-content` and `align-items` can all be removed.

brc-dd:
display isn't `block` or `inline-block` by default for every browser.

src/client/theme-default/styles/components/vp-doc.css
```diff
@@ -426,9 +426,6 @@
   /*rtl:ignore*/
   right: 12px;
   z-index: 3;
```
brc-dd, line 428:
```suggestion
  z-index: 3;
  display: block;
```

btea: `button` has already set `position:absolute`, at this time the computed value of the `display` property becomes 'block', so I think it can be removed here.

brc-dd: > button has already set position:absolute, at this time the computed value of the display property becomes 'block'

How?? On Chrome it's using the user agent stylesheet that has display: inline-block

PR #2172, opened 2023-03-30
Cms/posts/sample 1

Ashley1985:


pikax:
@Ashley1985 you're creating Pull Request on the Vitepress, please don't do that unless you're legit contributing for the project, otherwise you're just wasting time from people maintaining the repo.

PR #2186, opened 2023-04-03
feat(lastUpdate): Added option to specify desired date format.

mllull:
When you write documentation in Spanish, French or any language that has different "date formats", it's very confusing the date that is displayed without this PR.

For exemple, as per today, you can see the *Last updated* as: `4/3/2023, 15:00`

If the target people of the documentation is not from an English-spoken country, they will think that de documentation was written the past month.

This PR is brings the possibility of specifying a *date format* to the `lastUpdated` option. That's useful in *non-english*  language.

### DIFF
--- a/docs/reference/default-theme-config.md
+++ b/docs/reference/default-theme-config.md
@@ -304,6 +304,21 @@ export default {
 }
 ```
 
+## lastUpdatedFormat
+
+- Type: `string`
+- Default: `null`
+
+The string to be applied to format last updated time using [VueUse/useDateFormat](https://vueuse.org/shared/useDateFormat/).
+
+```ts
+export default {
+  themeConfig: {
+    latdUpdatedFormat: 'DD/MM/YYYY HH:mm:ss'
+  }
+}
+```
+
 ## algolia
 
 - Type: `AlgoliaSearch`
--- a/docs/reference/site-config.md
+++ b/docs/reference/site-config.md
@@ -343,7 +343,7 @@ This option injects an inline script that restores users settings from local sto
 
 Whether to get the last updated timestamp for each page using Git. The timestamp will be included in each page's page data, accessible via [`useData`](./runtime-api#usedata).
 
-When using the default theme, enabling this option will display each page's last updated time. You can customize the text via [`themeConfig.lastUpdatedText`](./default-theme-config#lastupdatedtext) option.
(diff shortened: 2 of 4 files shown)


mllull:
P.S.: I'm using `useDateFormat` from  `vueuse/core` lib, which is a project dependency. There is no reason to reinvent the wheel by writing a new *data-parsing* function.

chgeo:
Wouldn’t it be more natural and easier to let users decide on the format, i.e. to use the browser/OS preference?
With the proposed change, as a site owner, I would not know which format to choose…

mllull:
@chgeo Right, maybe we can determine the right format from browser by executing this function:

```js
const getDateFormatString = () => {
  const formatObj = new Intl.DateTimeFormat('default').formatToParts(new Date());

  return formatObj
    .map(obj => {
      switch (obj.type) {
        case "day":
          return "DD";
        case "month":
          return "MM";
        case "year":
          return "YYYY";
        default:
          return obj.value;
      }
    })
    .join("");
}
```

This will return a `DD/MM/YYYY` or `MM/DD/YYYY` depending on the browser's language.

This solution takes us to a new question: How can we display the right hour format? Maybe adding new option to format hour and then mix it with the result from `getDateFormatString` funcion?

chgeo:
You can use `Intl.DateTimeFormat` with `undefined` to get the format in the user's locale.  I have tested it with Chrome and Safari, and they pick up the OS settings (after browser restart, potentially).

This is what the script in `VPDocFooterLastUpdated.vue` looks like.  Note that the format is only defined in abstract terms like `medium` and `short`.
For the `<time datetime="...">`, I think the ISOString is still best to express a 'neutral' format.

```diff
<script setup lang="ts">
 import { ref, computed, watchEffect, onMounted } from 'vue'
 import { useData } from '../composables/data'

-const { theme, page, lang } = useData()
+const { theme, page } = useData()

 const date = computed(() => new Date(page.value.lastUpdated!))
+const format = new Intl.DateTimeFormat(undefined/*user locale*/, { dateStyle: 'medium', timeStyle: '>
 const isoDatetime = computed(() => date.value.toISOString())
 const datetime = ref('')

@@ -12,7 +13,7 @@ const datetime = ref('')
 // potential differences in timezones of the server and clients
 onMounted(() => {
   watchEffect(() => {
-    datetime.value = date.value.toLocaleString(lang.value)
+    datetime.value = format.format(date.value)
   })
 })
 </script>
```

mllull:
Closing this PR because https://github.com/vuejs/vitepress/pull/2332 is a rewrite of it by introducing changes of the above comments.

PR #2189, opened 2023-04-04
fix(theme): ensure correct outline state

microdev1:
Fixes a couple of edge cases in outline:
- The first outline link remains active when content is scrolled back to top.
- The state is incorrect when `isAsideEnabled` becomes `true` after being set to `false` and the content was scrolled/updated.

Background changes:
- Use `onContentUpdated` instead of `onUpdated` as the later is triggered twice.
- Calculate `anchors` only when the content is updated or `isAsideEnabled` becomes `true`.
- Remove `scroll` listener when the content has outline disabled or `isAsideEnabled` becomes `false`.


bluwy:
This PR is quite old, but I think most of the issue explained is now fixed.

PR #2198, opened 2023-04-06
feat: allow passing props and children/slots to defineClientComponent

Liberty-liu:
When using the defineClientComponent method, can pass in prop and children for the component


brc-dd:
Hey, just a question here, why is this required? You can pass them directly like this right:

```vue
<script setup>
import { defineClientComponent, ref, h } from 'vitepress'

const customCompRef = ref(null)
const ClientComp = defineClientComponent(() => import('component-that-access-window-on-import'))
</script>

<template>
  <ClientComp ref="customCompRef">
    <template #foo>
      <div>foo</div>
    </template>
  </ClientComp>
</template>
```

PR #2223, opened 2023-04-15
fix: double base path when navigating to local search results

ChuHoMan:
Hello, first of all, thank you for implementing the local search feature. I just discovered the double base path issue, and I was planning to open a PR to fix it, but it was already resolved shortly after. 🚀
After comparing my fix with the existing one, I think using `siteConfig.site.base` is a better option, as currently `siteConfig.site.base` is set to `/` and it aligns with the purpose here. 😊

### DIFF
--- a/src/node/plugins/localSearchPlugin.ts
+++ b/src/node/plugins/localSearchPlugin.ts
@@ -97,10 +97,7 @@ export async function localSearchPlugin(
   function getDocId(file: string) {
     let relFile = path.relative(siteConfig.srcDir, file)
     relFile = siteConfig.rewrites.map[relFile] || relFile
-    let id = path.join(
-      siteConfig.userConfig.base ?? siteConfig.site.base,
-      relFile
-    )
+    let id = path.join(siteConfig.site.base, relFile)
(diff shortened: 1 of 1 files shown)


src/node/plugins/localSearchPlugin.ts
```diff
@@ -97,7 +97,10 @@ export async function localSearchPlugin(
   function getDocId(file: string) {
     let relFile = path.relative(siteConfig.srcDir, file)
     relFile = siteConfig.rewrites.map[relFile] || relFile
-    let id = path.join(siteConfig.userConfig.base ?? '/', relFile)
+    let id = path.join(
+      siteConfig.userConfig.base ?? siteConfig.site.base,
```
brc-dd, line 101: ah, just `siteConfig.site.base` should be sufficient, no need to use userConfig.base

src/node/plugins/localSearchPlugin.ts
```diff
@@ -97,7 +97,10 @@ export async function localSearchPlugin(
...
+      relFile
+    )
```
brc-dd, line 103:
```suggestion
    let id = path.join(siteConfig.site.base, relFile)
```

PR #2235, opened 2023-04-16
fix(search): avoid getting double base when fetchExcerpt

ChuHoMan:
After setting config `base: '/base/'`, avoid getting double base when fetch excerpt.


brc-dd:
already fixed in 185213c

ChuHoMan:
ok, I saw it.

PR #2236, opened 2023-04-16
fix(search): avoid body scroll when using local search

ChuHoMan:
This behavior aligns with Algolia Search.

### DIFF
--- a/src/client/theme-default/components/VPLocalSearchBox.vue
+++ b/src/client/theme-default/components/VPLocalSearchBox.vue
@@ -23,7 +23,7 @@ import {
   watch,
   watchEffect,
   type Ref,
-onBeforeUnmount
+  onBeforeUnmount
 } from 'vue'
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPLocalSearchBox.vue
```diff
@@ -312,6 +314,16 @@ useEventListener('popstate', (event) => {
   event.preventDefault()
   emit('close')
 })
+
+/** Lock body */ 
+const isLocked = useScrollLock(document.body)
```
brc-dd, line 319: directly accessing document will cause issues with ssr. make it something like

```
const body = shallowRef<HTMLElement | null>(null)

onMounted(() => {
  body.value = document.body
})

const isLocked - useScrollLock(body)
```

ChuHoMan: ok, I will fix it.

brc-dd (approved):

Issue #2245, opened 2023-04-17
detailed view in local search breaks if fetched excerpt has useData call


LanXiaoChr:
### Describe the bug

When I create a new folder in the root directory and create an md file in the new folder. Then click the 'Display detailed list' button and search. Details are not displayed in the search results. And an error appears in the console.
```
GET http://localhost:4173/assets/guide/test_index.md.undefined.js net::ERR_ABORTED 404 (Not Found)
TypeError: Failed to fetch dynamically imported module: http://localhost:4173/assets/guide/test_index.md.undefined.js
```

### Reproduction

**1. Install (Default configuration)**
```
npm init -y
npm i -D vitepress
npx vitepress init
```
File structure:
```
.
├─ .vitepress
│  └─ config.ts
├─ api-examples.md
├─ markdown-examples.md
├─ index.md
└─ package.json
```

**2. Create folders and files**
File structure:
```
.
├─ guide
│  └─ test.md
├─ .vitepress
│  └─ config.ts
├─ api-examples.md
├─ markdown-examples.md
├─ index.md
└─ package.json
```
guide/test.md
```
# vite
vitepress
```

**3. Turn on local search**
config.ts
```
import { defineConfig } from 'vitepress'

// https://vitepress.dev/reference/site-config
export default defineConfig({
  title: "My Awesome Project",
  description: "A VitePress Site",
  themeConfig: {
    search: {
      provider: 'local'
    },
    // https://vitepress.dev/reference/default-theme-config
    nav: [
      { text: 'Home', link: '/' },
      { text: 'Examples', link: '/markdown-examples' }
    ],

    sidebar: [
      {
        text: 'Examples',
        items: [
          { text: 'Markdown Examples', link: '/markdown-examples' },
          { text: 'Runtime API Examples', link: '/api-examples' }
        ]
      }
    ],

    socialLinks: [
      { icon: 'github', link: 'https://github.com/vuejs/vitepress' }
    ]
  }
})
```

**4. Build and preview**
```
npm run docs:build
npm run docs:preview
```

**5. Search**
![111111111111](https://user-images.githubusercontent.com/48700158/232397128-a0480633-ac91-4c68-a12f-258e6888c7c4.png)

### Expected behavior

Show the details.

### System Info

```shell
Windows 10 22H2
NodeJS v16.14.0
Edge 112.0.1722.48
Vitepress 1.0.0-alpha.72
```

### Validations

[...]

PR #2253, opened 2023-04-18
fix(search): local search showDetailedList not working in pro env in windows

zonemeen:
fix https://github.com/vuejs/vitepress/issues/2245

The function `slash` should be inside the function `pathToFile`.


brc-dd:
Yeah, but it still won't fix that issue.

zonemeen:
I'll try to fix this when I have time later.

PR #2254, opened 2023-04-18
feat(search): clear search text when localSearchBox exit

zonemeen:
It looks weird if the search text is not cleared when localSearchBox exits.

### DIFF
--- a/src/client/theme-default/components/VPLocalSearchBox.vue
+++ b/src/client/theme-default/components/VPLocalSearchBox.vue
@@ -6,8 +6,7 @@ import {
   onKeyStroke,
   useEventListener,
   useLocalStorage,
-  useScrollLock,
-  useSessionStorage
+  useScrollLock
 } from '@vueuse/core'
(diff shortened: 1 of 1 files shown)


brc-dd:
Hey, this was intentional, so you can sort of resume search. Also when the box is reopened, that text is already selected so if you start typing it gets replaced. It will be better to disable this only behind some flag like `disableQueryPersistence` (defaults false).

zonemeen:
Thanks for your reply. So how should I configure to disable this?

brc-dd:
You can add an option similar to this:

https://github.com/vuejs/vitepress/blob/1769d329cc1f517136d9a543385edce335e5dd24/types/default-theme.d.ts#L299

Then before close is emitted just set `filterText` to `''`.

zonemeen:
OK, I understand. Can I submit this new feature?

brc-dd:
Yeah sure.

zonemeen:
Ok, I will resubmit the code on this branch when I have time later. Thanks for your patience and guidance.😄

PR #2265, opened 2023-04-21
fix(router): scroll back to the hash anchor even if it is already selected

lchzh3473:
即使hash与当前hash相等，我们仍然需要将其滚动到对应锚点。
否则当我们点击锚点后手动滚回原位，再次点击相同的锚点时将没有任何反应。
Even if the hash is equal to the current hash, we still need to scroll it to the corresponding anchor.
Otherwise, when we click on the anchor and then manually roll back and click on the same anchor again, there will be no response.

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -188,7 +188,7 @@ export function createRouter(
             ) {
               // scroll between hash anchors in the same page
               if (hash) {
-                // avoid non-response with the same hash anchor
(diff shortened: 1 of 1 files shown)


brc-dd (approved):

brc-dd:
Hey can you run `pnpm format` and commit again?

lchzh3473:
> Hey can you run `pnpm format` and commit again?

OK, I've run pnpm format and committed the changes.

PR #2267, opened 2023-04-21
fix: validate directory when creating project

zuixinwang:


zuixinwang:
@brc-dd 
hello, is this PR useful? 😃

PR #2269, opened 2023-04-21
feat: Change the member introduction to html

Liberty-liu:
This is handy when I want to write some html in member introduction, such as a tag

### DIFF
--- a/src/client/theme-default/components/VPTeamMembersItem.vue
+++ b/src/client/theme-default/components/VPTeamMembersItem.vue
@@ -31,9 +31,7 @@ defineProps<{
             {{ member.org }}
           </VPLink>
         </p>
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPTeamMembersItem.vue
```diff
@@ -180,6 +178,13 @@ defineProps<{
   margin: 0 auto;
 }
 
+.desc ::v-deep(a) {
```
brc-dd, line 181: `::v-deep` is deprecated. Use `:deep` instead.

Liberty-liu: I'm sorry, it's been corrected

brc-dd (approved):

PR #2272, opened 2023-04-22
fix: allow data files in packages to be found

### LINKED ISSUE #2242: data loader files from package not working in dev
### Describe the bug

I've built a fairly comprehensive blog theme for my use.  

https://github.com/jcamp-code/vitepress-blog-theme
https://vitepress-blog.netlify.app/blog/

It extends the default theme by using the slots and built in features of VitePress.  As you can see from the demo, it works just fine.  I was hoping to package it up and be able to use it across a few sites so that I could make updates in one place and just update the package of my sites.

However, when I try that, the new site throws 

```
Unknown file extension ".css"
Unknown file extension ".vue"
```

It's not processing these files from the package correctly

### Reproduction

https://github.com/jcamp-code/starter-vitepress-blog

### Expected behavior

Would like VitePress to properly pull in Vue and CSS files from packages

### System Info

```shell
NA
```

### Additional context

_No response_

### Validations

JohnCampionJr:
Fixes #2242 

This allows them to work inside the packaged theme, but not outside.

This is only a start, but its something.

### DIFF
--- a/src/node/plugins/staticDataPlugin.ts
+++ b/src/node/plugins/staticDataPlugin.ts
@@ -76,7 +76,7 @@ export const staticDataPlugin: Plugin = {
 
         // use vite's load config util as a away to load Node.js file with
         // TS & native ESM support
-        const res = await loadConfigFromFile({}, configFile)
+        const res = await loadConfigFromFile({} as any, configFile)
 
         // record deps for hmr
         if (server && res) {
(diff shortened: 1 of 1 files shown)


JohnCampionJr:
Marked as ready as even just this part of it would be helpful; I can live without the composables for the moment.

PR #2286, opened 2023-04-25
fix(cli): validate path when creating project

zuixinwang:


zuixinwang:
@brc-dd 
😃 please check this PR

brc-dd:
I don't think we need the path to exist. Running `init` and providing the target directory as `docs` automatically creates the directory if it doesn't exist.

zuixinwang:
aha, so should I close this PR ? 😹 
If merged, it is necessary to update the vitepress docs to inform readers that they need to ensure that the directory exists before initialization

PR #2292, opened 2023-04-25
feat: allow use `@` prefix with `@include:`

### LINKED ISSUE #2291: inconsistent behavior importing markdown

We can import code snippets using `@` prefix as root, `@` prefix not available when using markdown file inclusion, this is is not working: `<!--@include: @/partials/why.md-->`

Use the `@` prefix also when using mardown file inclusion

None.

### Additional context

I have a local PR, I will send it in a few minutes.

### Validations

userquin:
closes #2291


src/node/markdownToVue.ts
```diff
@@ -88,9 +88,13 @@ export async function createMarkdownToVueRenderFn(
     // resolve includes
     let includes: string[] = []
     src = src.replace(includesRE, (m, m1) => {
+      const atPresent = m1[0] === '@'
```
userquin, line 91: move inside try/catch block or check length > 0

userquin: or just return m1 if no length

brc-dd (approved):

PR #2310, opened 2023-04-28
feat: preserve user log level

### LINKED ISSUE #2309: VitePress not preserving log level when building, using always warn
### Describe the bug

Sometimes we need a custom build log level to debug other plugins or VP build end hooks, since VP has 'warn' when resolving vite config, there is no way to show info level: for example, `@vite-pwa/vitepress` logs `workbox` build result only with `info`.

### Reproduction

Configure `vite: { logLevel: 'info' }`  in docs VP config file here and run build docs script.

### Expected behavior

User log level should be preserved.

### System Info

```sh
NA
```

### Additional context

I'll send a PR to fix this issue.

### Validations

userquin:
closes #2309

### DIFF
--- a/src/node/build/buildMPAClient.ts
+++ b/src/node/build/buildMPAClient.ts
@@ -16,7 +16,7 @@ export async function buildMPAClient(
     root: config.srcDir,
     cacheDir: config.cacheDir,
     base: config.site.base,
-    logLevel: 'warn',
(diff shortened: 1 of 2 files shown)


brc-dd (approved):

userquin:
Configuring log level inside `vite.config` file will not work, will work only when using `vite.logLevel` inside VP config module.

PR #2311, opened 2023-04-28
feat: extract metadata to static asset

bojanrajh:
This PR extracts metadata (`__VP_HASH_MAP__` and `__VP_SITE_DATA__`) to a static asset so it's not included in HTML of every page and can be cached in browser.

Our build size for 1250 markdown files was reduced from 800MB to 480MB (-40%), from 100s to 93s (-7%), from 650kB to 380kB per page in average.


brc-dd:
This one we'll first discuss internally. Regarding #2297, I was taking break for sometime, will try to get it merged this week.

brc-dd:
Some of the suggestions from Evan:

1. generate a `<link rel="preload">` at the very top for the meta chunk
2. keep this behind some option like `metaChunk: true` for now (so that we can profile / compare the perf in production on a few sites)

bojanrajh:
Implemented the `metaChunk` config. Should I also add the `preload` tag to the [`shouldPreload`](https://github.com/vuejs/vitepress/blob/dfe5246f283a8c31a26cde3171292a8625ff4154/src/node/build/render.ts#L93) block instead? And should we do the same for the [`appChunk.fileName`](https://github.com/vuejs/vitepress/blob/dfe5246f283a8c31a26cde3171292a8625ff4154/src/node/build/render.ts#L177)?

PR #2319, opened 2023-04-29
feat(theme): support light shiki themes

brc-dd:
closes #2200


brc-dd:
Currently it checks if the user has specified light/dark mode theme for shiki and adds a class to wrapper. We can use this class to override styles in light mode. CSS vars and code needs slight cleanup. Ignore the failing tests, they will be fixed on removing adaptive themes from default.

/cc @userquin

Issue #2312, opened 2023-04-29
'Return to top' button is always visible


Unbinilium:
### Describe the bug

Hello, I noticed that in `home` layout, on mobile devices (with narrow viewport width), the 'Return to top' button is always visible even you're already at the top, and I have difficulties in hiding this button.

### Reproduction

1. Use a phone browser (with narrow viewport width).
2. Open [https://vitepress.dev](https://vitepress.dev), then you'll see the 'Return to top' button is visible.

### Expected behavior

1. It can be automatically hide or show depend on the page location.
2. It can be explicitly disable in configurations, like `themeConfig: { ... returnToTop: false, ...}`.

### System Info

```sh
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1
    Memory: 106.47 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.0.0 - /opt/homebrew/bin/node
    npm: 9.6.4 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Safari: 16.4
  npmPackages:
    vitepress: ^1.0.0-alpha.74 => 1.0.0-alpha.74
```

### Additional context

https://user-images.githubusercontent.com/15633984/235286646-a19d669e-8274-4b2a-b55a-9f372d978d87.mov

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

ATQQ:
<img width="507" alt="image" src="https://user-images.githubusercontent.com/42485491/235334918-187a4805-a820-4a4b-83f6-3cb4d6df315d.png">

I found out today, too

PR #2320, opened 2023-04-30
fix(theme): 'Return to top' button is always visible in the home page

zonemeen:
fix https://github.com/vuejs/vitepress/issues/2312


brc-dd:
Hey can you explain the changes about `has-borderTop` and removing `home` condition from `hasSidebar`? IMO just `v-if="frontmatter.layout !== 'home'"` thing on local nav would've fixed that issue.

zonemeen:
`has-borderTop` is added because when there is no sidebar, the line disappears in the mobile view as shown in the figure below.

Normal:
image

No sidebar:
image

PR #2328, opened 2023-05-01
fix(theme): add slash to hotKey search

viniciusteixeiradias:
### What is the purpose of this pull request? 

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

Solving this problem [here (Vuejs)](https://github.com/vuejs/docs/issues/2240).

### DIFF
--- a/src/client/theme-default/components/VPNavBarSearch.vue
+++ b/src/client/theme-default/components/VPNavBarSearch.vue
@@ -54,25 +54,6 @@ const preconnect = () => {
   })
 }
 
-const isEditingContent = (e: KeyboardEvent): boolean => {
-  const element = e.target as HTMLElement;
-  const tagName = element.tagName;
-
-  return (
-    element.isContentEditable ||
-    tagName === 'INPUT' ||
-    tagName === 'SELECT' ||
-    tagName === 'TEXTAREA'
-  );
-}
-
-const isHotKey = (e: KeyboardEvent): boolean => {
-  const isBindKey = e.key === 'k' && (e.ctrlKey || e.metaKey)
-  const isSlashKey = e.key === '/'
-
-  return isBindKey || isSlashKey
-}
-
 onMounted(() => {
   if (!__ALGOLIA__) {
     return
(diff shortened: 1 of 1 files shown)


brc-dd:
`/` needs specially handling. E.g. if you've a live code component in your page and you're typing slash there, it should not open the search box. We should copy the logic from here [`onKeyDown`](https://github.com/algolia/docsearch/blob/bae117ce6db0e51f7c0958215705dc849b25212c/packages/docsearch-react/src/useDocSearchKeyboardEvents.ts#L40-L45), [`isEditingContent`](https://github.com/algolia/docsearch/blob/bae117ce6db0e51f7c0958215705dc849b25212c/packages/docsearch-react/src/useDocSearchKeyboardEvents.ts#L11-L21)

viniciusteixeiradias:
@brc-dd Okay, I agree with this special treatment, but that would be a new PR, wouldn't it? My PR was only to resolve the opening after loading with `/`, this special treatment was already an existing need before my change.

I can do it here, but I would like to know your opinion, in my view they are different implementations that should be done in different PRs.

brc-dd:
That comment is for `!isOpen` check not `!isEditingContent`. Your PR will not behave as expected when there is any content editable or input/textarea element and one presses slash there.

viniciusteixeiradias:
@brc-dd I did not understand your comment, could you explain again? 

This treatment was already necessary before the opening of my PR, since the opening with `/` worked after the first opening. My PR was objective, making it possible to open the dialog with `/` after rendering the component.

Suppose treatment should exist, but is it something that should be done in this PR?

Obs: I tested with `input`, `text-area` and `select`, slash only opened the dialog when typed in the text-area.

brc-dd:
@viniciusteixeiradias 

https://user-images.githubusercontent.com/40380293/235582978-3b0b9786-ebbe-47e7-ae05-d0ba47c207db.mov

viniciusteixeiradias:
@brc-dd Sorry for my mistake, I performed the test before the build so it worked for me. It's my first time working in open source, I'm learning with you. Anyway, I uploaded the corrections

viniciusteixeiradias:
@brc-dd All good now?

PR #2329, opened 2023-05-01
feat(theme): use inert to avoid traverse menus and content with keyboard

### LINKED ISSUE #1332: menu (navigation drawer) should not be traversable with keyboard when not shown

userquin:
This PR includes `inert` attribute to prevent traverse content and menu on narrow sreens:
- moved `isScreenOpen` outside the composable
- included logic in `Layout` to deal with `inert` attribute, right now the attribute is not being handled by `vue`, if present handled as present (vue is rendering the attribute as `inert="false"`)

This PR fixes #1332 (closed by #1491, but not working in current version, check video)

https://streamable.com/tf9qtu

`inert` attribute supported by all major browsers: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert#browser_compatibility

~~TODO: remove `undefined` when https://github.com/vuejs/core/pull/8209 merged~~

### DIFF
--- a/src/client/theme-default/Layout.vue
+++ b/src/client/theme-default/Layout.vue
@@ -2,6 +2,7 @@
 import { computed, provide, useSlots, watch } from 'vue'
 import { useRoute } from 'vitepress'
 import { useData } from './composables/data'
+import { useNav } from './composables/nav'
 import { useSidebar, useCloseSidebarOnEscape } from './composables/sidebar'
 import VPSkipLink from './components/VPSkipLink.vue'
 import VPBackdrop from './components/VPBackdrop.vue'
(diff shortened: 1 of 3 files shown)


src/client/theme-default/Layout.vue
```diff
@@ -36,27 +39,27 @@ provide('hero-image-slot-exists', heroImageSlotExists)
 <template>
   <div v-if="frontmatter.layout !== false" class="Layout">
     <slot name="layout-top" />
```
userquin, line 41: should we wrap `layout-top` and `layout-bottom` slots?

userquin: should we also provide some `inert-*`?

PR #2347, opened 2023-05-06
feat(theme): add appearance transition feature

hooray:
Use [View Transition API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) to make appearance switch more dynamic.

Kapture 2023-05-07 at 05 36 16

But these APIs needs Chrome 111, so I added a configuration item `appearanceTransition` that can be selectively enabled.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -60,7 +60,9 @@ export default defineConfig({
       code: 'CEBDT27Y',
       placement: 'vuejsorg'
     }
-  }
+  },
+
+  appearanceTransition: true
 })
 
 function nav() {
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -60,9 +60,7 @@ export default defineConfig({
       code: 'CEBDT27Y',
       placement: 'vuejsorg'
     }
-  },
-
-  appearanceTransition: true
+  }
 })
 
 function nav() {
--- a/src/client/theme-default/components/VPSwitchAppearance.vue
+++ b/src/client/theme-default/components/VPSwitchAppearance.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { ref, onMounted, watch } from 'vue'
+import { ref, onMounted, watch, computed } from 'vue'
 import { useData } from '../composables/data'
 import { APPEARANCE_KEY } from '../../shared'
 import VPSwitch from './VPSwitch.vue'
@@ -12,6 +12,11 @@ const toggle = typeof localStorage !== 'undefined' ? useAppearance() : () => {}
 
 onMounted(() => {
   checked.value = document.documentElement.classList.contains('dark')
+})
+
+const isAppearanceTransition = computed(() => {
+  // @ts-expect-error: Transition API
+  return document.startViewTransition && site.value.appearanceTransition
 })
 
 function useAppearance() {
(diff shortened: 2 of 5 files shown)


hooray:
> Brilliant idea! I posted it on Twitter to collect feedback: [twitter.com/antfu7/status/1655588888105172994](https://twitter.com/antfu7/status/1655588888105172994)
> 
> Do you think we could enable `appearanceTransition` in this PR so we could play with it with the deploy preview?

Of course

brc-dd:
Might wanna consider some other factors too:

- Disable this on `prefers-reduced-motion`
- Maybe decrease the duration to something like 200 or 300. Currently, it seems slightly slow to me 👀

hooray:
> Might wanna consider some other factors too:
> 
> * Disable this on `prefers-reduced-motion`
> * Maybe decrease the duration to something like 200 or 300. Currently, it seems slightly slow to me 👀

I made some adjustments based on your suggestions

antfu:
Playing it on antfu.me, a few feedback

- I feel it's better to reverse the out/in animation for light/dark. Making the light icon more like a "sun" to light up the canvas. (already applied to https://antfu.me/ that you can play with)
- The animation is not very smooth, especially with low `duration` on a large screen (tho the FPS stays consistent according to Chrome DevTools). I am not sure if this is something we could improve. Maybe we can change the duration based on distance?

kiaking:
This is very cool and I'm surprised how simple it's to implement 😳  Well done!

However, my personal opinion is I'm against adding this to the core feature. Love to see this as some kind of plugin feature or recipe. Or tweak the core code to make VitePress easy to implement this effect.

The reason why I'm against adding this specific animation effect in the core 😓 

1. The growing circle animation I don't think is aligned with Vue or Vite tone.
2. Without having disciplined brand guideline, this kind of animation will almost certainly look cool at first but become off trend and not that cool very very quickly (do you remember what happened to polygon design trend).
3. If we add this, there's no turning back. If we undo or change the effect, some forks would not like it.
4. If we add this, I'm pretty sure we'll gets requests to add different types of effect (crossfade, slides, etc.). I don't think VitePress core wants to focus so much on animation effects.

That's been said and as I mentioned, it would be super cool if default theme has a cool way for users to hook in this kind of animation 👍

yuhengshen:
speed up the transition and simplify `[...clipPath].reverse()` to `clipPath.reverse()`

hooray:
> This is very cool and I'm surprised how simple it's to implement 😳 Well done!
> 
> However, my personal opinion is I'm against adding this to the core feature. Love to see this as some kind of plugin feature or recipe. Or tweak the core code to make VitePress easy to implement this effect.
> 
> The reason why I'm against adding this specific animation effect in the core 😓
> 
> 1. The growing circle animation I don't think is aligned with Vue or Vite tone.
> 2. Without having disciplined brand guideline, this kind of animation will almost certainly look cool at first but become off trend and not that cool very very quickly (do you remember what happened to polygon design trend).
> 3. If we add this, there's no turning back. If we undo or change the effect, some forks would not like it.
> 4. If we add this, I'm pretty sure we'll gets requests to add different types of effect (crossfade, slides, etc.). I don't think VitePress core wants to focus so much on animation effects.
> 
> That's been said and as I mentioned, it would be super cool if default theme has a cool way for users to hook in this kind of animation 👍

It makes sense, I think I need to think about whether to continue this PR 🤔

yuhengshen:
@brc-dd Every clipPath is a new array, so it’s okay to mutate the original array directly.

AgustinBanchio:
Would like to have a look into this as I would like this feature.

Do you think the most acceptable solution would be:

A. Passing a transition function through the themeConfig and handling it in VPSwitchAppearance in a way similar to this PR? (which also modifies the component to look different if a transition exists and is possible)
B. Allowing users to replace VPSiwtchAppearance with a custom component that has access to a version of `toggle` or `setClass` and they are free to wrap those around anything they want (including transitions and or having a different toggle look?

I feel like option B is more of a custom theme solution, so think option A is more likely to make sense

PR #2355, opened 2023-05-10
feat: code group tab scrollIntoView

izayl:
add smooth scrollIntoView when code group is scrollable

### DIFF
--- a/src/client/app/composables/codeGroups.ts
+++ b/src/client/app/composables/codeGroups.ts
@@ -8,6 +8,13 @@ export function useCodeGroups() {
       if (el.matches('.vp-code-group input')) {
         // input <- .tabs <- .vp-code-group
         const group = el.parentElement?.parentElement
+
+        const label = group?.querySelector(`label[for="${el.id}"]`)
+        label?.scrollIntoView({
+          behavior: 'smooth',
+          block: 'nearest',
+          inline: 'nearest'
+        })
         const i = Array.from(group?.querySelectorAll('input') || []).indexOf(el)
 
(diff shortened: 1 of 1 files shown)


brc-dd:
Ah, for some reason the scrolling isn't smooth if you click the left label, and it does weird scrolling if one clicks the currently selected element.

https://github.com/vuejs/vitepress/assets/40380293/fc8fed46-1e72-4ec3-a028-e3f76c28b49f

izayl:
let me fix it

izayl:
@brc-dd I found this is a bug of vitepress, see the recurrence demo, and I'll fix it in another pr

https://vitepress-codegroup-scroll-issue-izayl.vercel.app/markdown-examples.html

https://github.com/vuejs/vitepress/assets/10740043/53fae0ce-e76b-4313-8d00-0927c5547d59

izayl:
it now much better

https://github.com/vuejs/vitepress/assets/10740043/0cbde762-b3c6-456f-a10f-161238da9a31

izayl:
@brc-dd I've made some changes to the code based on your feedback, and I'd appreciate it if you could take another look.

brc-dd:
> for some reason the scrolling isn't smooth if you click the left tab

This thing still doesn't appear to be fixed. For example click the last tab in your example (`decode-array-primitive`) -- it will scroll smoothly. Then click the first tab (`decode-primitive`) -- it won't scroll smoothly.

The fix at #2362 too isn't valid (left comment there). It will be better to not set scroll behavior smooth. Just something like this should be fine IMO:

```ts
label?.scrollIntoView({ block: 'nearest' })
```

PR #2373, opened 2023-05-15
fix(build): display file name when rendering fails

Pandapip1:
This small patch will save many people a LOT of time debugging. It would have saved me at least an hour.

I can't imagine that it's intentional that there is no information regarding which file caused rendering to fail, so that's why I'm calling this a fix.


brc-dd:
Can you provide me a minimal reproducible example where the error stack has no mention of corresponding file?

Pandapip1:
I made this PR a few months ago, so I can't remember the exact issue I had. I believe that I made a typo and tried to acccess a property of an undefined variable. This should repro:

```
{{ shsghs.hfhfh  }}
```

brc-dd:
That already shows the file name:

image

\
I agree it shows that in a way that might not be clear (replaces `/` with `_` and the paths are corresponding to temp directory).

Pandapip1:
I think you misunderstand. This shows the file name of the markdown file where the error occurred, *not* the file name of the script in which the error occurred.

Pandapip1:
I'm kinda stuck with using my own patch for debugging until this is merged.

brc-dd:
> I think you misunderstand. This shows the file name of the markdown file where the error occurred, not the file name of the script in which the error occurred.

But the name of markdown file is easily inferable from the script name. an error in `.temp/foo_bar.md.js` means an error in `foo/bar.md`

PR #2377, opened 2023-05-16
docs: Expanded sections on applying custom CSS and SCSS to the default theme, and working with dark mode

SimonEast:
Hi VitePress team,

I spent some time figuring out how to apply style customizations to the default theme, and would like to contribute back to the official documentation to assist others in this process.

Here's my pull request, which relates only to the docs.

Since this is my first one, please feel free to provide feedback or tweaks/adjustments to what I have written, and I can make any necessary changes if required.

Cheers.

-----
<a href="https://stackblitz.com/~/github.com/simoneast/vitepress/tree/SimonEast%2Fpatch-81677">Review PR in StackBlitz Codeflow</a> _Submitted with [StackBlitz Codeflow](https://stackblitz.com/~/github.com/simoneast/vitepress/tree/SimonEast%2Fpatch-81677)._


docs/guide/extending-default-theme.md
```diff
@@ -26,15 +26,109 @@ import './custom.css'
...
+
 ```css
 /* .vitepress/theme/custom.css */
 :root {
+  /* The main brand color:
+     Typically used for Hero text, links, "brand" buttons, 
+     search result boxes, and other highlights */
   --vp-c-brand: #646cff;
-  --vp-c-brand-light: #747bff;
+
```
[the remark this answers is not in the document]
SimonEast: Thanks. I've added it back in. If you have a suggestion for a better description of where `brand-light` is used in the default theme, I can add that in also. I don't think I saw an obvious use of it the last time I looked.

docs/guide/extending-default-theme.md
```diff
@@ -26,15 +26,109 @@ import './custom.css'
...
+See [`vars.css`](https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/styles/vars.css) in the default theme for other variables that can be overridden.
+
+You can also apply any other CSS style rules here. While using CSS variables is good practice to help reduce repetition, plain CSS is also fine.
+
+Just remember to choose colors that work well in both dark and light modes, unless you disable the dark mode toggle or specify unique colors for both light and dark modes. See [Dark Mode](#dark-mode) below for more.
+
+## Using CSS Pre-Processors
+
+If you'd like to use SASS, LESS, or Stylus within your theme, first make sure the relevant pre-processor is installed by running the applicable command:
+
```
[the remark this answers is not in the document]
SimonEast: What's the benefit of the `@latest` suffix? It's not something I've really used before. Doesn't `npm install` typically fetch the latest stable version by default? Or does `@latest` grab alpha/beta versions?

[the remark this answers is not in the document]
SimonEast: Just circling back to your comment from last year. Is there any benefit in specifying `@latest` when running `npm install`? 

According to the [npm documentation](https://docs.npmjs.com/cli/v10/commands/npm-dist-tag#description):

> By default, `npm install <pkg>` (without any `@<version>` or `@<tag>` specifier) installs the latest tag.

Is this not necessarily the case with other package managers?

Or is `@latest` a way of overriding whatever version might already be present in your `package.json`?

I couldn't see any places where `@latest` is used in the current VitePress docs, but if that is the preferred syntax we can go with that for consistency. But it would just be good to understand the intention behind that.

SimonEast:
I've updated my PR so that the docs match the latest VitePress theme. Would be great to get any final feedback/tweaks that might be preventing it from being merged in.

I think it's important to get people up and running fast with the default theme but allow them to easily swap the color scheme for something that matches their project, and that's what this PR is intended to do.

docs/reference/default-theme-config.md
```diff
@@ -184,8 +184,10 @@ export type SidebarItem = {
 - Default: `true`
 - Can be overridden per page via [frontmatter](./frontmatter-config#aside)
 
+This is typically used to show the "On this page" section, containing links to major headings within the current page.
+
 Setting this value to `false` prevents rendering of aside container.\
-Setting this value to `true` renders the aside to the right.\
+Setting this value to `true` (the default) renders the aside to the right.\
```
brc-dd, line 190: no need to specify again,`` - Default: `true` ``  is already written above.

docs/reference/default-theme-config.md
```diff
@@ -17,7 +17,7 @@ export default {
 }
 ```
 
-**The options documented on this page only apply to the default theme.** Different themes expect different theme config. When using a custom theme, the theme config object will be passed to the theme so the theme can define conditional behavior based on it.
+**The options documented on this page only apply to the default theme.** Different themes expect different theme configuration options. When using a custom theme, the theme config object will be passed to the theme so the theme can define conditional behavior based on it.
```
brc-dd, line 20: the earlier one was already clear IMO

docs/reference/default-theme-config.md
```diff
@@ -1,6 +1,6 @@
 # Default Theme Config
 
-Theme config lets you customize your theme. You can define theme config via the `themeConfig` option in the config file:
+Theme config lets you customize your theme. You can define theme config via the `themeConfig` option inside [`.vitepress/config.[extension]`](./site-config):
```
brc-dd, line 3: don't think this is needed either

docs/guide/using-vue.md
```diff
@@ -200,9 +200,9 @@ Note that this might prevent certain tokens from being syntax highlighted proper
 
 ## Using CSS Pre-processors
 
-VitePress has [built-in support](https://vitejs.dev/guide/features.html#css-pre-processors) for CSS pre-processors: `.scss`, `.sass`, `.less`, `.styl` and `.stylus` files. There is no need to install Vite-specific plugins for them, but the corresponding pre-processor itself must be installed:
+VitePress has [built-in support](https://vitejs.dev/guide/features.html#css-pre-processors) for CSS pre-processors: `.scss`, `.sass`, `.less`, `.styl` and `.stylus` files. Simply install the pre-processor itself using the applicable command:
 
-```
+```bash
```
brc-dd, line 205: use ` ```sh ` instead to align with other docs

docs/guide/using-vue.md
```diff
@@ -200,9 +200,9 @@ Note that this might prevent certain tokens from being syntax highlighted proper
 
 ## Using CSS Pre-processors
 
-VitePress has [built-in support](https://vitejs.dev/guide/features.html#css-pre-processors) for CSS pre-processors: `.scss`, `.sass`, `.less`, `.styl` and `.stylus` files. There is no need to install Vite-specific plugins for them, but the corresponding pre-processor itself must be installed:
+VitePress has [built-in support](https://vitejs.dev/guide/features.html#css-pre-processors) for CSS pre-processors: `.scss`, `.sass`, `.less`, `.styl` and `.stylus` files. Simply install the pre-processor itself using the applicable command:
```
brc-dd, line 203: again unnecessary change. please avoid just rephrasing and moving things around

docs/guide/extending-default-theme.md
```diff
@@ -4,7 +4,7 @@ outline: deep
 
 # Extending the Default Theme
 
-VitePress' default theme is optimized for documentation, and can be customized. Consult the [Default Theme Config Overview](../reference/default-theme-config) for a comprehensive list of options.
+VitePress' default theme is optimized for documentation, and a number of elements can be customized via `.vitepress/config.[extension]`. Consult the [Default Theme Config Overview](../reference/default-theme-config) for a comprehensive list of options.
```
brc-dd, line 7: site config has a separate section, no need to keep putting same thing everywhere.

docs/guide/extending-default-theme.md
```diff
@@ -20,25 +20,129 @@ Before proceeding, make sure to first read [Using a Custom Theme](./custom-theme
 
 ## Customizing CSS
 
-The default theme CSS is customizable by overriding root level CSS variables:
+To apply custom CSS, you first need to create a custom VitePress theme. Place the following code into `.vitepress/theme/index.js` which imports the default theme and adds an additional CSS file to the bundle:
```
brc-dd, line 23: it can be index.ts too, maybe worth a mention

docs/guide/extending-default-theme.md
```diff
@@ -20,25 +20,129 @@ Before proceeding, make sure to first read [Using a Custom Theme](./custom-theme
...
 
-```js
-// .vitepress/theme/index.js
+::: code-group
+```js [.vitepress/theme/index.js]
```
brc-dd, line 26: code groups are not meant to just show title, keep the original formatting

docs/guide/extending-default-theme.md
```diff
@@ -20,25 +20,129 @@ Before proceeding, make sure to first read [Using a Custom Theme](./custom-theme
...
+   --vp-c-brand-3: #{$brand-color};
+   /* Brand color soft: for tip boxes etc. 
+      Must have transparency to account for nested boxes. */
+   --vp-c-brand-soft: #{transparentize($brand-color, 0.86)};
+ }
+```
+
+::: tip
+When SASS variables or functions are used to define a CSS variable, they need to be wrapped in `#{ ... }`. See the SASS documentation [here](https://sass-lang.com/documentation/breaking-changes/css-vars) and [here](https://sass-lang.com/documentation/variables).
+:::
```
brc-dd, line 128: remove this, we don't need to teach sass/scss to users. if someone is using a pre processor they should be aware of it's syntax

brc-dd (changes requested):

docs/reference/default-theme-config.md
SimonEast: Perhaps it's obvious once you've used VitePress for a while, but as a newcomer it's not clear which config file this is referring to and where it's located. I initially wondered "_Does VitePress have more than one config file?_" and "_Is there a theme config file, or is it the same as the VitePress config file?_" and "_Which folder should I be looking in?_".

I think having the path here reduces that initial confusion.

docs/guide/extending-default-theme.md
SimonEast: Well, I find it cleaner and clearer to have the filename outside the actual code block. It also means that when you click the "copy" button, you don't have the additional comment included. I think using a code-group tab for this purpose is a clean solution, but if you're against that I can revert to the original.

docs/guide/extending-default-theme.md
SimonEast: Yes, I agree that it's not necessary to explain basic SCSS syntax for the most part, *however* in this case I believe it's warranted since:

* It's difficult to find within the SASS documention. It's hidden under a "breaking changes" document, and not mentioned at all on the main [SASS Variables](https://sass-lang.com/documentation/variables/) doc page where you would expect.
* Combining CSS variables and SASS variables may not be standard practice for many devs (generally you would stick to one or the other), but when customizing VitePress in this way it becomes necessary

Having this tip here would have saved me 20mins of hunting through my CSS inspector for why it wasn't working, and then paging through multiple Google results until I found the issue.

docs/reference/default-theme-config.md
SimonEast: Yes, that's true. I've removed it.

docs/guide/using-vue.md
SimonEast: Done

docs/guide/using-vue.md
SimonEast: I felt the "Vite-specific plugin" point was a side-note that just added confusion when I initially read it (it led me to ask "_Am I supposed to understand the difference between a Vite-plugin and a pre-processor?_"). The revised version jumps straight to the commands necessary to get them up and running, and puts the side-note below, for those needing it.

But if you feel strongly about keeping it as is, I can revert.

docs/guide/extending-default-theme.md
SimonEast: Done. 👍

docs/reference/default-theme-config.md
SimonEast: Well "config" is a bit more vague and also not an actual word in proper English - more like developer slang that might refer to some insider knowledge. I didn't really understand what kind of configuration it referred to when I first read it. I thought "configuration options" would be a more accurate term.

But I can skip this change if necessary.

docs/guide/extending-default-theme.md
SimonEast: As mentioned in another comment, I initially found it confusing to work out which config file I was supposed to modify. Just trying to help new users get up to speed quickly.

If you're opposed to the repetition, perhaps it could just be a hyperlink instead, like this:

```suggestion
VitePress' default theme is optimized for documentation, and a number of elements can be customized via [the site config file](../reference/site-config). Consult the [Default Theme Config Overview](../reference/default-theme-config) for a comprehensive list of options.
```

Issue #2318, opened 2023-04-29
Home Page/Features Section: layout is inconsistent


demonGeek:
### Describe the bug

A grid layout is used when there are multiples of 2, 3, or 4 features but for any other number of features (1, 5, 7, 11 etc.) the grid is no longer displayed and each feature uses 100% width.
![6-features](https://user-images.githubusercontent.com/3934839/235317452-f393e41c-3bfc-45f6-852d-f0020d75931e.png)
![5-features](https://user-images.githubusercontent.com/3934839/235317453-7cf1178b-7ffc-4a17-9f13-54c4b7046259.png)

### Reproduction

1. Create a markdown document and set the frontmatter `layout` to `home`.
2. Create a `features` section .
3. Add a feature.
4. View the result.
5. Repeat steps 3-4.

### Expected behavior

Provided the width > 640px, the grid layout should be maintained with each feature occupying the same amount of horizontal space.

### System Info

```sh
System:
    OS: Linux 5.15 Ubuntu 22.04.2 LTS 22.04.2 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-2600K CPU @ 3.40GHz
    Memory: 8.96 GB / 15.59 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.12.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Brave Browser: 112.1.50.125
    Chromium: 112.0.5615.49
    Firefox: 112.0.2
  npmPackages:
    vitepress: ^1.0.0-alpha.74 => 1.0.0-alpha.74
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
You can adjust it via CSS for now.

PR #2382, opened 2023-05-18
feat(theme): home features section layout consistent

zonemeen:
close https://github.com/vuejs/vitepress/issues/2318

### DIFF
--- a/src/client/theme-default/components/VPFeatures.vue
+++ b/src/client/theme-default/components/VPFeatures.vue
@@ -26,7 +26,7 @@ const grid = computed(() => {
     return 'grid-3'
   } else if (length % 3 === 0) {
     return 'grid-6'
-  } else if (length % 2 === 0) {
+  } else if (length % 2 === 0 || length % 2 === 1) {
     return 'grid-4'
   }
 })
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPFeatures.vue
```diff
@@ -26,7 +26,7 @@ const grid = computed(() => {
     return 'grid-3'
   } else if (length % 3 === 0) {
     return 'grid-6'
-  } else if (length % 2 === 0) {
+  } else if (length % 2 === 0 || length % 2 === 1) {
```
brc-dd, line 29: Just a simple else should work here, right? (as `length % 2 === 0 || length % 2 === 1` will always be true)

```suggestion
  } else {
```

zonemeen:
@brc-dd  And when the length is one, should it be set to `grid-1`? Like this:

image

Otherwise it is like this:

image

I think it is more reasonable to set the style as `grid-1` when the length is one.

brc-dd:
Yeah for length = 1, `grid-1` seems fine.

zonemeen:
Yeah, I have resubmitted the code.😀

brc-dd (approved):

brc-dd:
Removed the grid-1 thing, because there is no CSS corresponding to that -- it is the default behavior. We just need to make sure to not apply grid-4 on that case. Updated the condition accordingly.

PR #2384, opened 2023-05-19
refactor: standardize import sort

zonemeen:


zonemeen:
This is indeed a bit inefficient.😂 So have you considered using eslint?

brc-dd:
ESLint was quite an overhead, most of the code was being maintained by the team and also things were changing too frequently. We can try having some good config once the things get more stable.

PR #2399, opened 2023-05-21
docs: Update site-config.md fix typo

manchan4869:
fix site-config#markdown


brc-dd:
It's TS interface. Both `;` and `,` are valid there.

PR #2400, opened 2023-05-22
fix(theme): support custom target and rel in navbar links for mobile

sapphi-red:
#1993 was lacking support for `VPNavScreenMenu` (shown when the screen width is narrow). This PR add that.


brc-dd (approved):

lslzl3000:
can we also add this feature to locale links? 
beta.1 not working on locales

PR #2403, opened 2023-05-22
feat(node): cli shortcuts add restart command

alexzhang1030:
add `restart` shortcut, copied from [vite](https://github.com/vitejs/vite/blob/e444375d34db1e1902f06ab223e51d2d63cd10de/packages/vite/src/node/shortcuts.ts#L85-L91)

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -34,7 +34,7 @@ if (!command || command === 'dev') {
     await server.listen()
     logVersion(server.config.logger)
     server.printUrls()
-    bindShortcuts(server)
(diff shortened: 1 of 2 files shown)


brc-dd:
Ah, restart might not work like this. It'll restart the vite server, but won't actually rerun the vitepress part. You'll need to pass `createDevServer` to `bindShortcuts` and run `await server.close(); await createDevServer()` in action.

alexzhang1030:
Thanks for reminding me, is this correct now? /cc @brc-dd

brc-dd (approved):

PR #2418, opened 2023-05-25
fix(release): not support direct import of json in windows

zonemeen:


brc-dd:
Seems weird. Which Node version are you on 👀  It should've worked on v16+ (irrespective of OS).

zonemeen:
node v16.14.0. 

But it keeps throwing errors, which I also find weird.🤣

userquin:
I'm using latest Node 18 LTS and it is working  (previously using Node 16.17.0 LTS and also working).

PR #2423, opened 2023-05-26
fix: import docsearch css only when provider is algolia

sxzz:


brc-dd:
Nah, we need that CSS to be present. That `VPAlgoliaSearchBox.vue` is dynamically imported. Open the deploy preview, it is showing this:

image

brc-dd:
PS: the local search provider mode also uses docsearch styles for the search button

PR #2428, opened 2023-05-27
perf(build): speed up copying of shared and client files

zonemeen:
1.Use `fast-glob` stream mode instead of synchronous mode. Synchronous mode blocks the main thread and affects performance.
2.Use `Promise.all` to copy files in parallel, not sequentially.
3.This code has some performance issues, using `await Promise.all` in the `forEach` will still block the thread.
https://github.com/vuejs/vitepress/blob/86a3c24d2a0efee2df4a742c7d5c329a0b6e4b10/scripts/copyShared.js#L4-L9

After each five times of docs-build time test, the build time without optimization is:
 1. 11.84s
 2. 12.5s
 3. 12.94s
 4. 12.56s
 5. 11.76s

the build time after optimization is:
1. 10.25s
2. 11.81s
3. 11.73s
4. 10.04s
5. 10.36s

windows 10
node v16.14.0

### DIFF
--- a/scripts/copyClient.js
+++ b/scripts/copyClient.js
@@ -5,10 +5,9 @@ function toDest(file) {
   return file.replace(/^src\//, 'dist/')
 }
 
-const copyPromises = []
-fg.stream('src/client/**').on('data', (file) => {
-  if (/(\.ts|tsconfig\.json)$/.test(file)) return
-  copyPromises.push(copy(file, toDest(file)))
-})
-
-await Promise.all(copyPromises)
+await Promise.all(
+  fg.sync('src/client/**').map((file) => {
+    if (/(\.ts|tsconfig\.json)$/.test(file)) return
+    return copy(file, toDest(file))
+  })
+)
--- a/scripts/copyClient.js
+++ b/scripts/copyClient.js
@@ -5,7 +5,10 @@ function toDest(file) {
   return file.replace(/^src\//, 'dist/')
 }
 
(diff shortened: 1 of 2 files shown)


brc-dd:
This might actually lead to race conditions. Suppose `fg.stream` emits `data` event for some file late, while the control reaches `await Promise.all`. In that case those file will not present in `copyPromises`.

zonemeen:
I don't think it will lead to a race condition, because `Promise.all(copyPromises)` waits for all the Promise in the `copyPromises` array to finish executing.

brc-dd:
Yeah that part I agree, but what if a promise is not there in `copyPromises`. `fg.stream(...).on('data', ...)` is async in itself.

zonemeen:
Do you mean there is a potential danger of missing some files?🤔

brc-dd:
Yeah. See:

* at time = 0, we get a file from stream, we push it to `copyPromises`
* at time = 1, the control reaches that `await Promise.all(...)` statement
* at time = 2, we get another file from stream, we push it to `copyPromises`, but that won't be copied

zonemeen:
Yeah, I guess I have to write it like this.

```js
stream.on('end', async () => { 
   await Promise.all(copyPromises);
})
```

brc-dd:
Ah yeah that should work, but is there any need 😅 these scripts are only used while locally debugging/releasing stuff.

zonemeen:
Yeah, there really is no need for this.😂

Issue #2262, opened 2023-04-20
Local Search has performance issues when detailed list is enabled


RaniAgus:
### Describe the bug

If I edit the search terms with detailed list enabled, the whole page gets frozen until the search is complete.

### Reproduction

1. Click on search
2. Enable detailed list
3. Slowly enter a search term, letter by letter

### Expected behavior

The query shouldn't freeze the whole page until it's completed. Also, if the query isn't complete, I should be able to click somewhere else to cancel it and close the search dialog.

Also, it would be nice to have a spinner when the query is not complete.

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 16.21 GB / 19.44 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.19 - /mnt/c/Program Files/nodejs/yarn
    npm: 8.13.2 - ~/.nvm/versions/node/v16.14.0/bin/npm
  npmPackages:
    vitepress: 1.0.0-alpha.73 => 1.0.0-alpha.73
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

arcqiufeng:
The page freezes when I enter text in the search box. I have the same problem when detail is enabled.

brc-dd:
For now consider disabling the detailed view - `search.options.disableDetailedView: true`

PR #2438, opened 2023-05-29
fix(theme): localSearch detailed list performance issue

zonemeen:
Fix local search display detailed list performance problem mentioned in https://github.com/vuejs/vitepress/issues/2262

This is mainly due to performance problem caused by this regular expression. Manipulating characters with regular expression can obviously cause performance problem.

https://github.com/vuejs/vitepress/blob/62d1110848e9b8944d920232bee185d8066194dd/src/client/theme-default/components/VPLocalSearchBox.vue#L122

Once the page has more anchors and more types of anchors or contents, it is easy to cause the page to freeze when searching.

The problem can be reproduced in this link by entering a `j` letter when searching: https://stackblitz.com/edit/vite-h3557a

After testing, the performance improvement is particularly huge.


src/client/theme-default/components/VPLocalSearchBox.vue
```diff
@@ -182,14 +181,22 @@ debouncedWatch(
...
-        }
+        const headings = $('h1, h2, h3, h4, h5, h6')
+        headings.each((_, el) => {
+          const anchor = $(el).find('a').attr('href')?.slice(1)
+          let next = $(el).next()
+          let html = ''
+          while (next.length && next[0].name !== el.name) {
+            const nonHTagHtml = `<${next[0].name} class="${
+                next[0].attribs.class
+            }">${next.html()}</${next[0].name}>`
```
brc-dd, line 194: what's the reason behind selecting only class here?

zonemeen: Since class is separated by cheerio, the elements with class will lose their class style if they are not written this way.

zonemeen:
> Hey can you check if the solution at #2486 works fine? I don't think we need cheerio on browser.

The reason why I introduced cheerio is because the node side has the same problem.

https://github.com/vuejs/vitepress/blob/c995b9f61d90aa7671371373c5772ab59b516fc5/src/node/plugins/localSearchPlugin.ts#L216

https://github.com/vuejs/vitepress/blob/c995b9f61d90aa7671371373c5772ab59b516fc5/src/node/plugins/localSearchPlugin.ts#L229

brc-dd:
Yeah, I meant we don't need cheerio for browser-side code. We can add it for node-side code. I'm saying this because adding cheerio to client will add another 200-300kB JS there.

zonemeen:
🤔Yeah, so there is really no need to introduce the browser-side cheerio.

PR #2461, opened 2023-06-01
refactor: simplify the import path

zonemeen:


brc-dd:
Hey a suggestion here, don't do merge. Use "squash and merge" option instead. That way master can have linear history.

zonemeen:
Ok, sorry about that. Thanks for the suggestion.

zonemeen:
Does this need to reset?

brc-dd:
Nah leave it.  No need to revert. Just remember to do squash next time.

PR #2466, opened 2023-06-01
feat(theme): add custom label for social links

AgustinBanchio:
Currently, the accessibility label for social links is derived from the icon name. When using a custom icon with an SVG, the aria-label is empty. 

This PR adds an optional field to SocialLink called `label`.

When a SocialLink includes the `label` string, it will be used as the aria-label. This allows custom icons to have a label and improve accessibility, and also to overwrite the default labels if so desired.

Behaviour is unchanged if the new optional field is not present.

This PR also updates the documentation and includes an example.

I understand that the word label might be confusing and users could be expecting the default theme to render this label as an icon subtext instead of it being an accessibility label. Please let me know if you think a different name would be more suitable.

### DIFF
--- a/docs/reference/default-theme-config.md
+++ b/docs/reference/default-theme-config.md
@@ -214,7 +214,9 @@ export default {
         icon: {
           svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>Dribbble</title><path d="M12...6.38z"/></svg>'
         },
-        link: '...'
+        link: '...',
+        // You can include a custom label for accessibility.
+        label: 'cool link'
       }
     ]
   }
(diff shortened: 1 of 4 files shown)


brc-dd (approved):

docs/reference/default-theme-config.md
```diff
@@ -214,7 +214,9 @@ export default {
         icon: {
           svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>Dribbble</title><path d="M12...6.38z"/></svg>'
         },
-        link: '...'
+        link: '...',
+        // You can include a custom label for accessibility.
+        label: 'cool link'
```
kiaking, line 219: Maybe this should be named `aria-label`? `label` sounded to me like if it were to add actual text after the icon 👀

PR #2486, opened 2023-06-07
perf(search): use dom apis instead of regex-based section parsing

brc-dd:
x-ref: #2262, #2438

### DIFF
--- a/src/client/theme-default/components/VPLocalSearchBox.vue
+++ b/src/client/theme-default/components/VPLocalSearchBox.vue
@@ -119,8 +119,6 @@ watchEffect(() => {
 
 const results: Ref<(SearchResult & Result)[]> = shallowRef([])
 
-const headingRegex = /<h(\d*).*?>.*?<a.*? href="#(.*?)".*?>.*?<\/a><\/h\1>/gi
-
 const enableNoResults = ref(false)
 
 watch(filterText, () => {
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPLocalSearchBox.vue
```diff
@@ -182,14 +178,18 @@ debouncedWatch(
...
+        const promises = Array.from(headings).map(async (el) => {
+          const href = el.querySelector('a')?.getAttribute('href')
+          const anchor = href?.startsWith('#') && href.slice(1)
+          if (!anchor) return
+          let html = ''
+          while ((el = el.nextElementSibling!) && !/^h[1-6]$/i.test(el.tagName))
+            html += el.outerHTML
+          map!.set(anchor, html)
+        })
+        await Promise.all(promises)
```
zonemeen, line 191: Wouldn't Promise.all be a bit redundant?

brc-dd: Yeah, might be. Can we benchmark that somehow? this async thing vs headings.forEach 👀

zonemeen: > Can we benchmark that somehow?

All I can think of is the difference in multiple execution times between these two methods as a benchmark test👀There is no difference in execution time after I measured it several times locally.

And what I have learned is that the `forEach` or `map` loop executes all of the iterations simultaneously.

PR #2488, opened 2023-06-07
docs: add root path to gitignore suggestion

Lehoczky:
We currently suggest the user to add the `.vitepress/dist` and `.vitepress/cache` folders to their `.gitignore` file, but if they just copy-paste these values, it won't work. They will either need to prefix this path with the vitepress root folder, or use `**/.vitepress/dist`.

I've implemented the formal option in this PR, but happy change it to the latter if that sounds better to you guys.

Before:
image

After:
image


Lehoczky:
Or we can add these folders directly to the user's gitignore file

PR #2494, opened 2023-06-12
fix(cli): `--minify false` not working

luoway:
> https://www.npmjs.com/package/minimist
> Numeric-looking arguments will be returned as numbers unless opts.string or opts.boolean is set for that argument name.

When the argument is boolean-looking, it is necessary to configure it to be returned as boolean.

### DIFF
--- a/src/client/theme-default/components/VPLocalSearchBox.vue
+++ b/src/client/theme-default/components/VPLocalSearchBox.vue
@@ -421,7 +421,7 @@ function formMarkRegex(terms: Set<string>) {
             <button
               class="back-button"
               :title="$t('modal.backButtonTitle')"
-              @click="selectedIndex > -1 && $emit('close')"
+              @click="$emit('close')"
             >
               <svg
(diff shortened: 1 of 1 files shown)


luoway:
It's recommended to use `cac` instead of `minilist`, just like what [`vitejs/vite`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/cli.ts#L14) did.

src/node/cli.ts
```diff
@@ -6,7 +6,9 @@ import { init } from './init/init'
 import { version } from '../../package.json'
 import { bindShortcuts } from './shortcuts'
 
-const argv: any = minimist(process.argv.slice(2))
+const argv: any = minimist(process.argv.slice(2), {
+  boolean: ['minify']
```
brc-dd, line 10: this will set minify to be false by default

brc-dd, line 10: this will set minify to false if not provided

PR #2495, opened 2023-06-12
Improve rendering and performance by adding decoding="async" and loading="lazy" attributes to <img> tags 🖼️

sanjaiyan-dev:
This pull request implements two optimizations to enhance the rendering and improve the performance of image loading on our application. By adding the `decoding="async"` attribute to the `` tag and using `loading="lazy"`, we can achieve faster and more efficient image rendering.

**Changes Made:**

- Added `decoding="async"` attribute to the `` tag, enabling asynchronous decoding of images.
- Implemented `loading="lazy"` attribute to the `` tag, enabling lazy loading of images.

**Benefits:**

- Improved Rendering: Asynchronous decoding allows the browser to decode images in the background, enhancing the rendering speed and ensuring a smoother user experience.
- Enhanced Performance: Lazy loading defers the loading of images until they are actually needed, reducing initial page load time and saving bandwidth for users with slower connections.

Ref-: https://web.dev/learn/design/responsive-images/#image-decoding

### DIFF
--- a/src/client/theme-default/components/VPImage.vue
+++ b/src/client/theme-default/components/VPImage.vue
@@ -19,6 +19,7 @@ export default {
     <img
       v-if="typeof image === 'string' || 'src' in image"
       class="VPImage"
+      decoding="async"
       v-bind="typeof image === 'string' ? $attrs : { ...image, ...$attrs }"
       :src="withBase(typeof image === 'string' ? image : image.src)"
       :alt="alt ?? (typeof image === 'string' ? '' : image.alt || '')"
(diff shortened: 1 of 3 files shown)


sanjaiyan-dev:
🤔

you-hengh:
I agree with this proposal. I observed that when I click on an article, all the images of this article are requested. In fact, I only want to see the head of the article, but requesting all the images causes unnecessary waste of CDN traffic. , hoping to support this proposal to reduce unnecessary waste of traffic.

you-hengh:
@brc-dd Do you have time to look at this proposal?

sanjaiyan-dev:
👀

bluwy (commented):
So as I understand, this PR only adds the decoding and loading attributes to certain VP components, and not everywhere. Images in markdown for example isn't handled here. Plus there's now the `markdown.image.lazyLoad` option for that.

About `decoding="async"`, from https://www.tunetheweb.com/blog/what-does-the-image-decoding-attribute-actually-do/, it doesn't feel to me there's a strong reason to add or not to add, personally I'm inclined to keep things as is.

About `loading="lazy"`, which is really only for VPTeamMembersItem in this PR, I don't think we should add it as team member images can appear above the fold for full-page team pages.

Issue #2404, opened 2023-05-22
Change the `dist/assets` directory name


mtermoul:
### Describe the bug

When you run the build command all js, css, images are copied to `/.vitepress/dist/assets/`. Is there a way to change the `assets` directory? I know that using the `outDir` config param can be used to change the the build output directory. However I could not find any param to allow me to change the output directory for the assets.

Vite have a build config parameter to change the the assets directory called `site: { build: { assetsDir: 'myBlueFolder' } }`. So I tried to pass this vite config parameter to vitepress, but it doesn't seem to work.
This is my vitepress config along with the site config.

```
export default defineConfig({
  title: 'My first vitepress site',
  description: '...',
  themeConfig: {
 ...  
  },
  outDir: './dist',
  vite: {
    build: {
      assetsDir: 'static-assets',
    },
  },
});
```

### Reproduction

https://stackblitz.com/edit/vite-rt6z31?file=docs%2F.vitepress%2Fconfig.ts

### Expected behavior

The assets directory should be changed from `/dist/assets` to `/dist/static-assets`

### System Info

```sh
System:
    OS: macOS 12.0.1
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 217.72 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 113.0.5672.126
    Firefox: 113.0
    Safari: 15.1
  npmPackages:
    vitepress: ^1.0.0-alpha.47 => 1.0.0-alpha.75
```

### Additional context

We need to have a vitepress config parameter to allow the user to change the default output directory for `assets`

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

mtermoul:
[...]

PR #2497, opened 2023-06-12
feat(build): support custom `assetsDir`

zonemeen:
close https://github.com/vuejs/vitepress/issues/2404


brc-dd (approved):

dukexie:
The documentation has been updated, looking forward to the release of the code.

src/node/config.ts
```diff
@@ -73,6 +73,9 @@ export async function resolveConfig(
     })
   const site = await resolveSiteData(root, userConfig)
   const srcDir = normalizePath(path.resolve(root, userConfig.srcDir || '.'))
+  const assetsDir = userConfig.assetsDir
+    ? userConfig.assetsDir.replace(/\//g, '')
```
brc-dd, line 77: @zonemeen Do you remember why this replace statement was added?

zonemeen: Sorry for the late reply. I was only thinking about the top-level assetsDir configuration and not about supporting nested assetsDir, and then I added this regular rule to remove any slash at the beginning or end that a user might add to a customized assetsDir name.

PR #2498, opened 2023-06-12
refactor: defining default values for Props using withDefaults.

Alfred-Skyblue:
Define props default values using withDefaults.


Alfred-Skyblue:
TBR

Issue #1656, opened 2022-12-01
Show icon for external links


mrbrianevans:
### Is your feature request related to a problem? Please describe.

It would be nice to easily be able to see whether a link (href) in the documentation is internal/external. Ie whether it takes you to another page of the same website or opens a different website in a new tab. 
The way that many sites achieve this is to have the external link icon next to external links. This indicates to users that it will open a new tab. 

### Describe the solution you'd like

When a link is pointing to a different origin (external link), it should be displayed with an External Link icon.

Like this:

[google.com ![external-link](https://user-images.githubusercontent.com/53117772/205054181-d27cbca5-5d2e-4012-9329-38d0f8e52f34.svg)](https://google.com)

### Describe alternatives you've considered

Could manually make a new `img` tag for every link that points to another website. Very tedius and unnecessary.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

PR #2501, opened 2023-06-13
feat(theme): option to show icon for external links

zonemeen:
closes https://github.com/vuejs/vitepress/issues/1656


brc-dd:
It'll work, but it limits customizability for third party themes. I think we should add the markdown changes behind a flag, and register the component similar to how we do the badge.

Alternatively, I think this can be added directly via CSS. I'll try something.

brc-dd:
@zonemeen Updated. Please take a look.

zonemeen:
Yeah, I didn't take into account the customizability of the third-party theme, I think it's fine after this change.

PR #2528, opened 2023-06-20
fix: simple-git-hooks not working

zonemeen:
It's not clear from which version my local `simple-git-hooks` is not working, I need to update `simple-git-hooks` manually by myself.

So add postinstall script, copied from [vue-core](https://github.com/vuejs/core/blob/d2c3d8b70b2df6e16f053a7ac58e6b04e7b2078f/package.json#LL39C6-L39C17)

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,8 @@
     "docs-debug": "node --inspect-brk ./bin/vitepress dev docs",
     "docs-build": "run-s build docs-build-only",
     "docs-build-only": "node ./bin/vitepress build docs",
-    "docs-preview": "node ./bin/vitepress preview docs"
+    "docs-preview": "node ./bin/vitepress preview docs",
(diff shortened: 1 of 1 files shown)


brc-dd:
Don't add that. At vue/core, it working because they are using a monorepo kind of setup. Here if you add postinstall, it will be run for users trying to install vitepress.

zonemeen:
So now we have to manually execute `npx simple-git-hooks` to make simple-git-hooks take effect. I think it is necessary to remind contributors in the [contributing](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md) file.

brc-dd:
Yeah that would be fine.

zonemeen:
I will submit another PR to remind the rest of the contributors.👀

Issue #1039, opened 2022-07-22
Fail to locate the anchor when id is not a valid css selector


gyhyfj:
### Describe the bug

If the md file’s h2 title start with special character,
page can't scroll to the right place as usual when I click toc in `.VPDocAsideOutline` ,   on the right side of the web,

### Reproduction

For example:
The md file is like this:
```md
# Reading
...
... 
## `- + * /`
...
...
```

If I click toc in `.VPDocAsideOutline` , on the right side of the web,
the page can't scroll to the right place as usual.
Chrome terminal report :
```js
DOMException: Failed to execute 'querySelector' on 'Document': '#-5' is not a valid selector.
    at c (https://www.gyhyfj.com/assets/app.4bd937d4.js:1:124560)
    at $t (https://www.gyhyfj.com/assets/app.4bd937d4.js:1:12734)
    at qe (https://www.gyhyfj.com/assets/app.4bd937d4.js:1:12813)
    at HTMLAnchorElement.n (https://www.gyhyfj.com/assets/app.4bd937d4.js:1:56875)
```

### System Info

```shell
System:
  OS: Windows 10 10.0.19044
  CPU: (8) x64 Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz
  Memory: 1.59 GB / 7.92 GB
Binaries:
  Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
  Yarn: 1.22.18 - ~\AppData\Roaming\npm\yarn.CMD
  npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.62)
  Internet Explorer: 11.0.19041.1566
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.vuejs.org).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
> begin with special char

No, is actually when heading "only" has special characters 😅.

xfq:
One way to fix this is to generate IDs using the hash of the section's content (like [Bikeshed](https://github.com/tabatkins/bikeshed/blob/f1c9d5eb99c9e633f56fffc6cd43143b7533f6eb/bikeshed/unsortedJunk.py#L1012) does), so that the generated ID is relatively stable.

[...]

PR #2560, opened 2023-06-30
fix(client): don't throw on using special chars in element ids

brc-dd:
x-ref: https://github.com/vuejs/vitepress/issues/1039#issuecomment-1614827077

### DIFF
--- a/src/client/theme-default/components/VPDocOutlineItem.vue
+++ b/src/client/theme-default/components/VPDocOutlineItem.vue
@@ -8,9 +8,7 @@ defineProps<{
 
 function onClick({ target: el }: Event) {
   const id = '#' + (el as HTMLAnchorElement).href!.split('#')[1]
-  const heading = document.querySelector<HTMLAnchorElement>(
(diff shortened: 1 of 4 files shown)


meteorlxy (approved):

src/client/theme-default/components/VPDocOutlineItem.vue
```diff
@@ -8,9 +8,7 @@ defineProps<{
 
 function onClick({ target: el }: Event) {
   const id = '#' + (el as HTMLAnchorElement).href!.split('#')[1]
-  const heading = document.querySelector<HTMLAnchorElement>(
-    decodeURIComponent(id)
-  )
+  const heading = document.getElementById(decodeURIComponent(id).slice(1))
```
posva, line 11: What about nod adding the `#` here and passing that to decodeURIComponent instead?

brc-dd: ah yeah, fixed in 00d94e9

PR #2564, opened 2023-07-01
feat(client): add onBeforePageLoad hook for router

buqiyuan:
I am using [fullpage.js](https://github.com/alvarotrigo/fullPage.js), but it doesn't work properly in `vitepress` because it uses `location.href` for anchors internally, which causes the page to be reloaded in `vitepress`. Therefore, I want to add an `onBeforeRoutePageLoad` hook to control the timing of page reloading.

https://github.com/vuejs/vitepress/assets/39730999/ba3a1e52-85e6-4fc3-9475-aefa1adb6ea4

`docs/components/ComponentInHeader.vue`:
```vue
<template>
  <span @click="handleClick">&#x26A1;</span>
</template>

<script setup lang="ts">
import { onMounted } from "vue";
import { useRouter } from 'vitepress'

const router = useRouter()

onMounted(() => {
  console.log('ComponentInHeader onMounted')
})

router.onBeforeRouteChange = ((to) => {
  console.log('onBeforeRouteChange',to) 
})

const handleClick = () => {
  location.href = '#escaping' 
  // const section = document.querySelector('#escaping')!;
  // section.scrollIntoView({ behavior: 'smooth' }); // 平滑滚动到元素位置
}
</script>
```

### DIFF
--- a/docs/reference/runtime-api.md
+++ b/docs/reference/runtime-api.md
@@ -86,8 +86,27 @@ Returns the VitePress router instance so you can programmatically navigate to an
 
 ```ts
 interface Router {
+  /**
+   * Current route.
+   */
   route: Route
-  go: (href?: string) => Promise<void>
+  /**
+   * Navigate to a new URL.
+   */
+  go: (to?: string) => Promise<void>
+  /**
+   * Called before the route changes. Return `false` to cancel the navigation.
+   */
+  onBeforeRouteChange?: (to: string) => Awaitable<void | boolean>
+  /**
+   * Called before the page component is loaded (after the history state is
+   * updated). Return `false` to cancel the navigation.
+   */
+  onBeforePageLoad?: (to: string) => Awaitable<void | boolean>
+  /**
+   * Called after the route changes.
+   */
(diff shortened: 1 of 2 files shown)


brc-dd:
why the early return thing?

buqiyuan:
I don't quite understand what you mean, if I return `Promise.reject()`, `vitepress` will treat it as an error catch, so I hope that the user's active cancellation can be distinguished from the program error
image

brc-dd (approved):

buqiyuan:
Hello! Thank you very much for your work! However, the issue I encountered has not been resolved yet. When using `location.href('#xxx')` for anchor links, it triggers the `popstate` event, but it does not execute the `go` function, so it does not pass through the `onBeforePageLoad` hook.

https://github.com/vuejs/vitepress/blob/fa3780f8ef99b88e998523570f08a4e7f86dfd1b/src/client/app/router.ts#L215-L217

brc-dd:
ah, moving that condition inside loadPage should work then.

brc-dd:
fixed in a3df617

PR #2571, opened 2023-07-03
refactor(theme): restore the external link icon color from previous version

zonemeen:
I think the color of the original external link icon is more appropriate.

image

vs 

image


brc-dd:
What about the links in text?

Current:

image

After this PR:

image

It was looking a bit weird to me, that's why I changed the color. We probably can introduce a color variable and override it only for VPLink (so that c-text-3 is used on sidebar/navbar links). WDYT?

zonemeen:
I think this implementation is feasible. I've submitted new code, take a look.

src/client/theme-default/components/VPDoc.vue
```diff
@@ -205,5 +205,6 @@ const pageName = computed(() =>
 
 .external-link-icon-enabled :is(.vp-doc a[href*='://'], .vp-doc a[target='_blank'])::after {
   content: '';
+  color: currentColor;
```
brc-dd, line 208:
```suggestion
```

src/client/theme-default/styles/components/vp-doc.css
```diff
@@ -534,6 +534,7 @@
   width: 11px;
   height: 11px;
   background: currentColor;
+  color: var(--vp-c-text-3);
```
brc-dd, line 537:
```suggestion
  background: var(--vp-c-text-3);
```

brc-dd (approved):

PR #2572, opened 2023-07-03
chore: official docs config add `externalLinkIcon`

zonemeen:


brc-dd:
We can add this, but I personally don't like clutter in case there are many external links:

image

Not moving forward with this for now. We can enable this if there are any strong reasons to have this.

PR #2575, opened 2023-07-03
fix(theme): --vp-nav-height should be 0 when navbar is hidden

buqiyuan:
before：
image

after：

**If this PR is merged, I will add document for it later.**


src/client/theme-default/components/VPNav.vue
```diff
@@ -1,16 +1,30 @@
...
 
 provide('close-screen', closeScreen)
+
+watchEffect(() => {
+  if (hasNavbar.value) {
+    document.documentElement.style.removeProperty('--vp-nav-height')
+  } else {
+    document.documentElement.style.setProperty('--vp-nav-height', '0')
+  }
+})
```
brc-dd, line 23: this will throw errors during build

buqiyuan: Yeah, you are right

itsWindows11:
In terms of design I'm not particularly a fan of this change, could add some spacing at the top (like 8 or 12px margin as nav height).

kiaking:
yeah it should have some spacing. Can this be fixed if we use `position: sticky` for navbar instead of `position: fixed`? 🤔

Issue #2570, opened 2023-07-02
Allow configuration of minisearch to not split at - Sign (Hyphen, Minus)


gregorwolf:
### Is your feature request related to a problem? Please describe.

As described in [How to search for items with an embedded hyphen? #108](https://github.com/lucaong/minisearch/issues/108) minisearch does split by default at Hyphen/Minus Sign. 

### Describe the solution you'd like

It should be possible to influence this behaviour by defining a [custom split variable](https://github.com/lucaong/minisearch/issues/108#issuecomment-963311944).

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

gregorwolf:
Thank you @zonemeen for the quick implementation and @brc-dd for the review and quick merge.

gregorwolf:
Feature request in the downstream project created: [Configure VitePress search to not split at - Sign (Hyphen, Minus) #270](https://github.com/cap-js/docs/issues/270)

PR #2576, opened 2023-07-03
feat(search): support `minisearch` customization

zonemeen:
closes https://github.com/vuejs/vitepress/issues/2570


brc-dd:
Restricted the options a bit. I don't think other options like `autoVacuum` or `autoSuggestOptions` are relevant to our usage. Removed field related options too as our code relies on them.

Issue #2540, opened 2023-06-22
createContentLoader URL generated from project root instead of src dir


SimonBerens:
### Describe the bug

The `createContentLoader` function generates URLs relative to the project root (https://github.com/vuejs/vitepress/blob/c4909e4298ec706cf1762cb36af03e5fd3637ccc/src/node/contentLoader.ts#L118C48-L118C52). 

When the src directory is set to something else, this causes the URLs to break.

I think the selected code should be `config.srcDir` instead of `config.root`.

### Reproduction

Load https://stackblitz.com/edit/vite-tjipzm?file=docs%2Fsrc%2Findex.md and try clicking the link to the blog post; it will lead to a 404.

If you de-nest the things under the `src` directory and remove `srcDir` in `config.ts`, the link to the blog will work.

### Expected behavior

The link takes me to the blog

### System Info

```sh
stackblitz
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

Issue #2462, opened 2023-06-01
`createContentLoader` resolves urls from project root


shadow-light:
### Describe the bug

`createContentLoader` resolves urls from project root rather than source directory. This results in invalid urls that are relative to the project root, but should be relative to the source directory.

### Reproduction

https://stackblitz.com/edit/vite-vhsxi3?file=docs%2Fsrc%2Findex.md

### Expected behavior

Urls should be relative to the source directory.

### System Info

```sh
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    vitepress: latest => 1.0.0-beta.1
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

zonemeen:
I think it cannot be defined as a bug. You can configure it like this:

```js
import { createContentLoader } from 'vitepress';

export default createContentLoader('src/*.md', {
  transform(raw) {
    return raw.map(({ url }) => ({
      url: url.replace('/src/', ''),
    }));
  },
});
```

PR #2578, opened 2023-07-04
fix: `createContentLoader` generates invalid url when sets `srcDir`

zonemeen:
closes https://github.com/vuejs/vitepress/issues/2540 
closes https://github.com/vuejs/vitepress/issues/2462


brc-dd:
I'll get back on this. Need to check on some projects as this will be a breaking change. Seems right though.

SapneshNaik:
@ brc-dd, when will this be merged? This is causing a weird issue where the logo is no longer showing if I set a custom `srcDir` rather leaving it blank as it comes by default.

brc-dd:
What does this have to do with logo?

SapneshNaik:
@brc-dd the Logo URL (set in the `themeconfig` ) breaks if you set a different source dir using the `srcDir` config option.

brc-dd:
@SapneshNaik That isn't related to this PR. Please create a separate issue for that.

domnantas:
Agree that this change makes sense, I don't see any reason for `createContentLoader` to have access outside `srcDir`. Docs would need to be updated here too https://vitepress.dev/guide/data-loading#createcontentloader

PR #2598, opened 2023-07-10
docs: add genshin map to the use cases

jiazengp:
docs: add genshin map to the use cases
<https://yuanshen.site/docs/en>


kiaking:
Sorry I don't think we can add this since it doesn't really align with other examples 🙇‍♂️

Issue #2344, opened 2023-05-05
Question: Exclude from Local Search and display Frontmatter data


mitchellknight:
The new Local Search is great! I just have a few questions:

1. How do I exclude results from Local Search? For example, let's say I had a `/secret/` folder that I would like to exclude all subpages from search results, how would I do that?
2. How do I get Local Search to use [frontmatter data](https://vitepress.dev/guide/frontmatter#accessing-frontmatter-data) (see screenshot)?

![Frontmatter](https://user-images.githubusercontent.com/1848796/236547876-610465f3-3624-4887-938e-f8779086d058.png)

jeremyriverain:
Hi,

I am currently migrating a doc website from Vuepress to Vitepress. I would like to exclude certain content from the search results but I can't find a way to do it with Vitepress. This ticket seems to indicate that it could be an improvement to Vitepress. Is this a possible development?

I have time to work on this subject. I can contribute.

Is there a particular direction you would like to take on this subject? Which API expose?

I imagine 2 ways to exclude content from search results.

1. from the `.vitepress/config.ts` file, by adding a new parameter, like `isSearchable`:

```typescript
// .vitepress/config
export default defineConfig({
  themeConfig: {
    search: {
      provider: 'local',
      options: {
        isSearchable (pageData) {
          return !pageData.startsWith('/excluded-folder')
        },
      },
    },
})
```

2. at the level of each page, by taking into account a new frontmatter property:

```markdown
---
search: false
---

<!-- excluded-folder/index.md -->
# my content excluded from search results
```


What do you think ?

brc-dd:
Yeah contributions are welcome. Make the API something like this instead:

```ts
interface LocalSearchOptions {
  ignore: string[] | RegExp | (pageData: PageData) => boolean
}
```

PR #2602, opened 2023-07-11
feat(search): exclude content from search results

jeremyriverain:
Hi,

This PR is related to #2344

As requested, I added a new `exclude` property in `LocalSearchOptions` so that the user can exclude content from search results in the vitepress configuration file.

The `pageData` parameter passed to the `exclude` function is partial. It only contains `relativePath` and `frontmatter` properties because I couldn't find a simple and straightforward way to retrieve `PageData` from `localSearchPlugin.ts`. Do you see a better way?


jeremyriverain:
Clicking on the search bar when the server starts for the 1st time and the cache is empty causes the page to reload. I think this is the behavior that made the previous test flaky.

To fix this, I found 2 possible solutions:

The 1st consists in running the tests with the application already compiled (using the variable `VITE_TEST_BUILD`).

The other solution is to visit the home page for the first time and click on the search bar before the actual test so that the page reload does not occur during the test. This is the option I choose because it is the one that seems to me the least intrusive with regard to other tests.

Do you think it is possible to adjust the reload behavior ?

jeremyriverain:
I attach a capture to illustrate this behavior:

Kapture 2023-07-11 at 22 31 30

brc-dd:
> Do you think it is possible to adjust the reload behavior ?

Add those deps to optimizeDeps.include if local search is enabled.

PR #2611, opened 2023-07-12
fix: make pathname protocol respect base url

yue4u:
## Problem:

Currently base url config is not applied when using `pathname://`,  which seems to be the wrong behavior.

Change this would:

1. matches [origin](https://github.com/facebook/docusaurus/blob/26ae4164d6f90c231c6687363a3907b5f9f172b8/packages/docusaurus/src/client/exports/Link.tsx#L81) behavior
2. since the idea of pathname protocol is to escape from spa (within the same site), prepending base url makes sense

## Current

with `base` set to `/docs/`:

`[foo](pathname:///foo)` renders to `/foo`

## Expect

`[foo](pathname:///foo)` renders to `/docs/foo`

### DIFF
--- a/docs/guide/asset-handling.md
+++ b/docs/guide/asset-handling.md
@@ -76,3 +76,27 @@ const { theme } = useData()
   <img :src="withBase(theme.logoPath)" />
 </template>
 ```
+
+Also note `pathname://` will respect `base` config if the rest is an internal non-relative link.
+
+**Input**
+
+With `base` set to `"/docs/"`:
+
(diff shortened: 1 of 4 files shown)


brc-dd:
To me the current behavior seems expected as otherwise how will one write a link to something that's outside the base? (other than writing absolute URLs). However, having pathname protocol itself seems like a bad design. One can always specify target blank as attributes. This will be marked as deprecated in later releases and probably dropped by v1.

Issue #2614, opened 2023-07-13
Cannot handle doc file name containing single quote


f-w:
### Describe the bug

When doc file name has single quote ('), VitePress throws error
```
[plugin:vite:vue] [vue/compiler-sfc] Unexpected token, expected "," (3:32)
...
***/docs/getting-started/whats-new.md:3:32
1  |  <script >
2  |  export const __pageData = JSON.parse("{\"title\":\"What's New\",\"description\":\"\",\"frontmatter\":{\"permalink\":\"/docs/what's-new/\"},\"headers\":[],\"relativePath\":\"docs/what's-new.md\",\"filePath\":\"docs/getting-started/whats-new.md\",\"lastUpdated\":1689116868000}")
3  |  export default {name:'docs/what's-new.md'}</script>
   |                         ^
```
Error indicates single quote is not escaped.

### Reproduction

Just put a single quote in md doc file name.

### Expected behavior

doc page rendered w/o error.

### System Info

```sh
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz
    Memory: 4.63 GB / 15.71 GB
  Binaries:
    Node: 18.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 9.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 114.0.5735.199
    Edge: Spartan (44.19041.1266.0), Chromium (114.0.1823.79)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    vitepress: ^1.0.0-beta.5 => 1.0.0-beta.5
```
```


### Additional context

VuePress doesn't have the issue.

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #2615, opened 2023-07-14
fix(build): cannot handle file name containing single quote

zonemeen:
fix https://github.com/vuejs/vitepress/issues/2614

If the filename contains a single quote. Since the final characters of the `name:'${ data.relativePath }'` contains more than two single quotes, it doesn't compile through.


f-w:
JSON.stringify doesn't escape single quote. Besides, file path is not JSON.

brc-dd (approved):

Issue #2629, opened 2023-07-16
don't process md files in public dir


brc-dd:

Pandapip1:
I would like this to be a toggleable option, as I have a use case in which I *do* want md files in the public dir to be processed.

brc-dd:
@Pandapip1 Any specific use-case? `public` directory is supposed to be copied as-is to dist. Currently it generates a URL like this `example.com/public/foo.html` if you've `foo.md` in your public directory.

Pandapip1:
That is in fact what I want. I have a bunch of articles that each have their own assets (stored in the public directory). However, sometimes the asset is a markdown file. The expected behavior is then that the markdown file gets rendered.

Issue #2526, opened 2023-06-19
`Public folder` outside of `.vitepress` or `docs`


CanadianBeaver:
### Is your feature request related to a problem? Please describe.

Please add on option where the public folder can be specified in build settings: https://vitepress.dev/reference/site-config#build
The public folder can be outside of `.vitepress` or `docs`

### Describe the solution you'd like

Also, please add an option, that public folder can be specified in one, two or more instances and the output has to be combined in result. F.e.: project can have different folders: /imagesfromJohn, /simpleimages, /mybestimages/forsite

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
Adding support for multiple public directories won't be possible as vite doesn't support that. You'll need to write a custom vite plugin to handle that.

PR #2632, opened 2023-07-17
fix(build): don't process md files in public dir

zonemeen:
fix https://github.com/vuejs/vitepress/issues/2629

### DIFF
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -102,7 +102,6 @@ export async function build(
 
       await Promise.all(
         ['404.md', ...siteConfig.pages]
-          .filter((page) => !page.startsWith('public/'))
           .map((page) => siteConfig.rewrites.map[page] || page)
           .map((page) =>
             renderPage(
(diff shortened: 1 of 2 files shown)


brc-dd:
Ah, it will be bit complex 😅 `publicDir` is actually configurable via vite config. Will need to handle that.

zonemeen:
I'll continue to work on it later when I have more time.

brc-dd:
Cool, I don't think it will be exactly possible to support that. Since, vite config can be in a separate file and the way code is laid out resolvePages is called outside plugin contexts too. Using configResolved hook is a no go. What we can probably do is introduce a vitepress level publicDir and discard whatever there is in vite config.

zonemeen:
It is indeed necessary to introduce a vitepress level `publicDir`.🤔Can create a new issue about this new feature.

brc-dd:
I'm reopening #2526. Let's track this there.

zonemeen:
Okay! I'll change the code in this branch when `publicDir` is supported.

Issue #2634, opened 2023-07-17
表格内容换行后页面就报错


sxdPanda:
### Describe the bug

| 属性        |      默认值      |  说明 |  对应组件 |
| ------------- | :-----------: | ----: | ----: |
| col 3 is      | right-aligneright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedright-alignedd | $1600 | $1600 |

### Reproduction

内容换行后就报错：componentStyles.md?import&t=1689578052281:23  DOMException: Failed to execute 'setAttribute' on 'Element': '6' is not a valid attribute name.

### Expected behavior

内容换行后就报错：componentStyles.md?import&t=1689578052281:23  DOMException: Failed to execute 'setAttribute' on 'Element': '6' is not a valid attribute name.

### System Info

```sh
win11 edge浏览器
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

zonemeen:
你好，能方便提供一下复现的仓库或者在线地址吗

CosmoLau:
> 

用反斜杠转义一下能临时解决
```txt
|test|test|
|:-|-|
|{test}|\{test\}|
```
大括号里的内容应该是被 vue 当成了文本差值吧

zonemeen:
解决方案在这个issue里边有了，原因里边也说明了 https://github.com/vuejs/vitepress/issues/2046

Issue #2440, opened 2023-05-29
Uncaught (in promise) DOMException: String contains an invalid character


Elringus:
### Describe the bug

Exception occurs when navigating to page with a specific content: https://naninovel.github.io/guide/configuration (it renders fine when loaded directly via URL; navigate somewhere else and back to the page to observe the issue)

After page reload, however, the content is rendered fine.

Can't figure what exactly causes the error, as the error log neither specify location of the invalid char nor the char itself.

### Reproduction

Page sources: https://github.com/Naninovel/naninovel.github.io/blob/main/docs/guide/configuration.md

1. Build an preview the project
2. Navigate to `/guide/configuration`
3. Notice the page is empty and error is logged
4. Reload the page and notice the page is rendered fine

Note: page reload only works in preview or published build; in dev reloading the page doesn't fix the issue.

### Expected behavior

Page should be rendered fine when first loaded.

### System Info

```sh
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 Intel(R) Core(TM) i7-6900K CPU @ 3.20GHz
    Memory: 45.66 GB / 63.92 GB
  Binaries:
    Node: 18.2.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 9.6.6 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 113.0.5672.127
    Edge: Spartan (44.19041.1266.0), Chromium (113.0.1774.50)
  npmPackages:
    vitepress: ^1.0.0-beta.1 => 1.0.0-beta.1
```

### Additional context

Full error message:

```
Uncaught (in promise) DOMException: String contains an invalid character
    patchAttr runtime-dom.esm-bundler.js:179
    patchProp runtime-dom.esm-bundler.js:328
    mountElement runtime-core.esm-bundler.js:5209
    processElement runtime-core.esm-bundler.js:5156
    patch runtime-core.esm-bundler.js:5028
    mountChildren runtime-core.esm-bundler.js:5284
    mountElement runtime-core.esm-bundler.js:5191
    processElement runtime-core.esm-bundler.js:5156
    patch runtime-core.esm-bundler.js:5028
    mountChildren runtime-core.esm-bundler.js:5284
    mountElement runtime-core.esm-bundler.js:5191
    processElement runtime-core.esm-bundler.js:5156
[...]

Issue #2046, opened 2023-03-07
An error occurred when rendering a markdown file when using vitepress


newstart0514:
### Vue version

3.2.45

### Link to minimal reproduction

https://github.com/newstart0514/vueProblem

### Steps to reproduce

Use vitepress to set the file route normally. The file is routed to the md file (needs to be imported to the local). This error occurs when opening the route

### What is expected?

The md file will be converted to html normally

### What is actually happening?

An error occurred and rendering cannot be performed normally. The error message is: Unhandled error during execution of scheduler flush This is likely a Vue internals bug.

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 AMD Ryzen 7 4800U with Radeon Graphics
    Memory: 4.01 GB / 15.37 GB
  Binaries:
    Node: 18.12.1 - D:\nodejs\node.EXE
    Yarn: 1.22.18 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.19.2 - D:\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (110.0.1587.63)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    vue: ^3.2.45 => 3.2.45
```

kiaking:
Could you provide a full reproduction since that repo contains only 1 file and that wouldn't work with VitePress 🤔

brc-dd:
<img width="1041" alt="image" src="https://user-images.githubusercontent.com/40380293/223950907-7ce2bda1-5192-426d-876d-c02e63cdeedc.png">

Unable to reproduce it.

newstart0514:
> Could you provide a full reproduction since that repo contains only 1 file and that wouldn't work with VitePress 🤔

Of course, you can visit my latest uploaded minimal replication environment. Thank you very much! Address: https://github.com/newstart0514/vitepressProblemMinDemo

newstart0514:
> <img alt="image" width="1041" src="https://user-images.githubusercontent.com/40380293/223950907-7ce2bda1-5192-426d-876d-c02e63cdeedc.png">
> 
> Unable to reproduce it.

Thank you very much for trying to reproduce it, but because of my mistake, you can't reproduce it very well. I have uploaded the smallest reproduction environment. You can try it. I'm very sorry. Address: https://github.com/newstart0514/vitepressProblemMinDemo

zonemeen:
[...]

PR #2641, opened 2023-07-18
feat: set markdown-it-attrs plugin default delimiter options

zonemeen:
https://github.com/vuejs/vitepress/issues/2634
https://github.com/vuejs/vitepress/issues/2440
https://github.com/vuejs/vitepress/issues/2046

These issues are experiencing similar problem, and I think it's necessary to set the markdown-it-attrs plugin's options regarding delimiters to a specific one character, rather than simply `{` and `}`, otherwise the compiled code can easily break vue syntax and crash the page.

For example, if the text is `{ minRows: 6, maxRows: 6 }`, will cause the page to crash.

### DIFF
--- a/docs/reference/site-config.md
+++ b/docs/reference/site-config.md
@@ -433,8 +433,8 @@ interface MarkdownOptions extends MarkdownIt.Options {
   // markdown-it-attrs plugin options.
   // See: https://github.com/arve0/markdown-it-attrs
   attrs?: {
-    leftDelimiter?: string
-    rightDelimiter?: string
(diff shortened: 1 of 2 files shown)


brc-dd:
This will be a breaking change, and it will impact a large number of projects 👀 We probably should just document this in FAQs or something.

zonemeen:
I agree with this one, I'll rework the submission later on.

brc-dd:
I'll create a troubleshooting doc, stating this and other common stuff like hydration errors, invalid esm modules, etc.

Issue #2623, opened 2023-07-16
[style] external links icon is displayed on the links in <VPTeamMembers>


aiktb:
### Describe the bug

![Snipaste_2023-07-16_18-29-21](https://github.com/vuejs/vitepress/assets/104723527/f98d03ac-5812-448b-a5c2-8931931a0abb)

### Reproduction

https://stackblitz.com/edit/vite-vcqcff?file=docs%2Fexample.md

### Expected behavior

![Snipaste_2023-07-16_18-33-34](https://github.com/vuejs/vitepress/assets/104723527/1639e43a-ae16-4d52-9622-c75cdfcb82e7)

### System Info

```sh
System:
    OS: Linux 5.15 Ubuntu 22.04.2 LTS 22.04.2 LTS (Jammy Jellyfish)
    CPU: (12) x64 AMD Ryzen 5 5600U with Radeon Graphics
    Memory: 6.10 GB / 6.70 GB
    Container: Yes
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    Yarn: 1.22.19 - /mnt/c/Users/aikotoba/AppData/Roaming/npm/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
    pnpm: 8.6.7 - /mnt/c/Users/aikotoba/AppData/Local/pnpm/pnpm
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #2650, opened 2023-07-19
fix(theme): external link icon remove in VPTeamMembers

zonemeen:
fix https://github.com/vuejs/vitepress/issues/2623


brc-dd:
fixed in f3a4597

Issue #2665, opened 2023-07-21
🐞Bug: `aside`'s outline marker is flicking when navigating towards above


xsjcTony:
### Describe the bug

The aside's outline marker will have an annoying flick when navigating towards above. All `vitepress` sites seems have this problem.

The video should be enough to demo the issue. (Sorry for the bad quality)

### Reproduction

https://github.com/vuejs/vitepress/assets/45784210/d21095b6-a769-45c5-9119-0fc88f041c3f

### Expected behavior

No flick

### System Info

```sh
Chrome 114
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
This was supposed to look cool 😅 It's hard to explain in words but this was kind of expected. It creates a sense of animation like https://easings.net/#easeOutBack (hover over the graph)

PR #2676, opened 2023-07-25
fix(theme): outline marker flicks when navigating towards above

zonemeen:
fix https://github.com/vuejs/vitepress/issues/2665

### DIFF
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -2,7 +2,6 @@ import type { DefaultTheme } from 'vitepress/theme'
 import { onMounted, onUnmounted, onUpdated, type Ref } from 'vue'
 import type { Header } from '../../shared'
 import { useAside } from './aside'
-import { throttleAndDebounce } from '../support/utils'
 
 // magic number to avoid repeated retrieval
 const PAGE_OFFSET = 71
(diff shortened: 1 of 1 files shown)


xsjcTony:
👀

brc-dd:
I don't think this fixes the flick thing. Sure reduces it by some factor 👀 

https://github.com/vuejs/vitepress/assets/40380293/8b7f8972-a3e3-4115-833b-de09496031c5

(at 0.5x slowdown)

Also, removing debouncing completely is not a good solution. We should pass some third parameter like this:

```ts
export function throttleAndDebounce(
  fn: () => void,
  delay: number,
  throttle = true
): () => void {
  let timeoutId: NodeJS.Timeout
  let called = false

  return () => {
    if (timeoutId) {
      clearTimeout(timeoutId)
    }

    if (!called) {
      fn()
      called = true
      setTimeout(
        () => {
          called = false
        },
        throttle ? delay : 0
      )
    } else {
      timeoutId = setTimeout(fn, delay)
    }
  }
}
```

zonemeen:
If the time between clicks is too short, the flick is obvious. This problem was introduced by https://github.com/vuejs/vitepress/pull/2527. Since the two requestAnimationFrames are triggered almost simultaneously, and then one of them was delayed, resulting in a flick. 

I guess need to modify the implementation of https://github.com/vuejs/vitepress/pull/2527 to avoid using two requestAnimationFrames.🤔

PR #2685, opened 2023-07-26
refactor(theme): remove redundant computed properties

zonemeen:
https://github.com/vuejs/vitepress/blob/51f28bfac96bbb14ea0175c796e0d18fff3b2cc5/src/client/theme-default/components/VPFeature.vue#L17

Since the `tag` attribute will definitely pass a value to the `VPLink` component, there is no need to use the computed attribute again.


brc-dd:
Ah VPLink is being used from other places too (without explicitly specifying tag).

zonemeen:
I misread it. Sorry.

Issue #2616, opened 2023-07-14
Incorrect header anchor icon position with multline headers


Elringus:
### Describe the bug

When the header is long enough to span multiple lines, the anchor icon is position on the last line.

![5797a63f949d72105f2ee13c11197dfa](https://github.com/vuejs/vitepress/assets/2056864/fb1b5eff-7bc5-45a4-b5b6-5a8873422581)

### Reproduction

1. Open https://stackblitz.com/edit/vite-badrqn?file=docs%2Findex.md
2. Hover over header
3. Notice the anchor icon is position over the last line

### Expected behavior

Icon should be position on the first line.

### System Info

```sh
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 Intel(R) Core(TM) i7-6900K CPU @ 3.20GHz
    Memory: 47.18 GB / 63.92 GB
  Binaries:
    Node: 18.2.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.6.6 - C:\Program Files\nodejs\npm.CMD
    pnpm: 7.29.3 - C:\ProgramData\chocolatey\bin\pnpm.EXE
  Browsers:
    Chrome: 114.0.5735.199
    Edge: Spartan (44.19041.1266.0), Chromium (114.0.1823.79)
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #2694, opened 2023-07-28
fix(style): incorrect header anchor icon position with multline headers

zonemeen:
closes https://github.com/vuejs/vitepress/issues/2616


brc-dd:
Messes up the badges 👀 

image

PR #2766, opened 2023-08-09
feat(types): Add Frontmatter type to PageData. This feature is added …

louiss0:
…so that frontmatter is type safe. When using frontmatter the user can access frontmatter in a type safe way.


brc-dd:
useData is not limited to just the default theme. One can have whatever they want in the frontmatter which may or may not conform to the types. If you want types in your project, you can probably use module augmentation and define frontmatter as whatever you plan on using in your pages.

louiss0:
> useData is not limited to just the default theme. One can have whatever they want in the frontmatter which may or may not conform to the types. If you want types in your project, you can probably use module augmentation and define frontmatter as whatever you plan on using in your pages.

The type I have written does acknowledge this. The OptionalFrontmatter type is written to provide autocomplete for the ones that are a part of the theme.

brc-dd:
I don't think it would provide anything other than intellisense, contrary to the PR description that it'd be "type safe". Accessing anything won't give any error. We can probably allow specifying the generic parameter in useData. So in your theme, you could create one file like this: https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/composables/data.ts and use it everywhere.

louiss0:
> I don't think it would provide anything other than intellisense, contrary to the PR description that it'd be "type safe". Accessing anything won't give any error. We can probably allow specifying the generic parameter in useData. So in your theme, you could create one file like this: https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/composables/data.ts and use it everywhere.
> 
> Also with the changes everything seems to fail: https://app.netlify.com/sites/vitepress-docs/deploys/64d3d75592c9d100072c2079 Have you tested them before?

I think It's better for the user to specify a generic that extend the type I could have taken the time to fix the errors but since you don't want the autocomplete it's better for the frontmatter to extend the type I created for it.

Here is the frontmatter type.
```ts 
//!  Leave the curly braces it makes this type work.
type FrontMatter = {
  title: string
  titleTemplate: string| boolean
  description:string
  prev:string | false | { text?: string; link?: string }
  next:string | false | { text?: string; link?: string }
  head: Array<HeadConfig>
  navbar: boolean
  sidebar: boolean
  footer: boolean
  lastUpdated: boolean
  editLink: boolean
  aside: boolean | 'left'
  outline: number | [number, number] | 'deep' | false
  pageClass: string
} & (
  | {
      layout: 'home'
      hero: Hero
      features: Array<Feature>
    }
  | { layout: 'doc' }
  | { layout: 'page' }
)
```
Here is the feature type. **This one has to be in shared for the frontmatter type to work**

```ts
interface Feature {
  // Show icon on each feature box.
  icon?: FeatureIcon

  // Title of the feature.
  title: string

  // Details of the feature.
  details: string

  // Link when clicked on feature component. The link can
  // be both internal or external.
  //
  // e.g. `guid/reference/default-theme-home-page` or `htttps://example.com`
  link?: string

  // Link text to be shown inside feature component. Best
  // used with `link` option.
  //
  // e.g. `Learn more`, `Visit page`, etc.
  linkText?: string

  // Link rel attribute for the `link` option.
  //
  // e.g. `external`
  rel?: string
}

type FeatureIcon =
  | string
  | { src: string; alt?: string; width?: string; height: string }
  | {
      light: string
      dark: string
      alt?: string
      width?: string
      height: string
    }

```

PR #2786, opened 2023-08-13
fix: the terminal print message is not clear

btea:
image


brc-dd:
White too hasn't enough contrast:

image

btea:
Maybe we can use `cyan` without background? 🤔

brc-dd:
Yeah, that looks fine for me.

PR #2809, opened 2023-08-18
refactor: drop body-scroll-lock in favor of vueuse useScrollLock

brc-dd:


brc-dd:
Scrollbar gap is not preserved. Everything else seem to work fine. iOS/Android devices don't have scrollbar enabled by default, so that should be fine in most cases (for collapsible sidebar and mobile nav menu). Desktop users having small viewport sizes might see the issue though. Local search will still have the issues, but it's there for algolia one too (#1054). Easiest fix would be to just set `html { scrollbar-gutter: stable }` in css.

PR #2912, opened 2023-09-06
Hide JS-required features if JS disabled

Pandapip1:
Things like light/dark mode and searching don't work properly when scripts are disabled. This hides those broken features when they are, well, broken.


brc-dd (dismissed):

brc-dd:
There are actually many more things that will be broken. For example all of these need js-required:

image

copy code, code groups, sidebar groups, vplocalnav, vpflyout, last updated, ... all of these will need js-required.

Pandapip1:
It looks like code groups are rendered server-side, so this should be fine.

Pandapip1:
CC @brc-dd for re-review

kiaking:
To be honest... wouldn't it make more sense if we create a new JS free theme if we need one 🤔 ? If we add this, anytime we add new feature to the theme, we have to be very careful. And probably end up having give up slightly complicated feature because of this limitation.

I don't think majority of users uses VitePress site without JS enabled 👀

kiaking:
> he point of this is pull is just a one-off improvement that ensures that most of the stuff that is broken when JavaScript is disabled is hidden.

It wouldn't be one-off, since we must keep this in mind forever in the future whenever we add new feature to the theme. And also "most of the stuff" would be an issue to because if we miss 1 part, we have to fix it once this change is introduced.

> If someone later notices that something doesn't work without JS, they can create pull request in that adds the js-required class themself.

This would be a bit unrealistic. Once any feature is added to the core, we can't ignore those features just because it wasn't added by the team.

I still believe creating a new theme that is carefully tailored in a way that works smoothly without JS would be much simpler 👀

Pandapip1:
> I still believe creating a new theme that is carefully tailored in a way that works smoothly without JS would be much simpler 👀

I disagree; I can't see a way to create an entirely new theme from scratch that is somehow *more* effort than this. I would love to be proven wrong, but I don't feel that I have enough time to learn how to do that properly. Also, it still wouldn't change the fact that there are broken features with the default theme with javascript off.

> This would be a bit unrealistic. Once any feature is added to the core, we can't ignore those features just because it wasn't added by the team.

Yes, you absolutely can ignore them. You're a volunteer maintaining a FOSS project. Don't feel that just because you've helped make something, that you have to maintain it for all of eternity :)

kiaking:
> I disagree; I can't see a way to create an entirely new theme from scratch that is somehow more effort than this.

Fork this repo then apply this change. Done.

Pandapip1:
Sorry, brain fart. Meant less, not more.

PR #2913, opened 2023-09-06
Hide some UI elements when printing

Pandapip1:
When I want to print an article, some UI elements still show up:
image
This PR hides them when trying to print.


itsWindows11:
Might even want to reduce the UI spacing/make the colors more readable but kind of outside the scope of this PR.

Pandapip1:
> Should also hide "Menu" and "Search" if you don't already do that in this PR
> Maybe also the "Help edit this page" and "Next/Prev" buttons

I think those are already hidden somehow. I can definitely add styling for those, though!

brc-dd:
> I think those are already hidden somehow. I can definitely add styling for those, though!

they aren't, check on the deploy preview: https://deploy-preview-2913--vitepress-docs.netlify.app/

also why hide navbar and footer?

Soitora:
> > I think those are already hidden somehow. I can definitely add styling for those, though!
> 
> they aren't, check on the deploy preview: https://deploy-preview-2913--vitepress-docs.netlify.app/
> 
> also why hide navbar and footer?

Doesn't seem necessary for a printout, but the Site Icon+Name should be kept

PR #2924, opened 2023-09-08
Change system info render for issue template

Soitora:
Not a meaningful change more than making the terminal output on issues easier on the eyes

### SH
image

### PWSH
image


brc-dd:
It isn't needed. That's not a shell / powershell script.

PR #2956, opened 2023-09-13
fix: make algolia search work with indices that don't return absolute urls

frederic-schwarz:
this enables the algolia search plugin to work also with indices crawled by the [netlify algolia plugin](https://www.algolia.com/doc/tools/crawler/netlify-plugin/quick-start/) which, depending on the configuration, can crawl pages with relative paths
| relative paths | absolute urls |
| --- | --- |
| image | image |

without the fix, using the algolia search plugin leads to
```
VPAlgoliaSearchBox.vue:88 Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
   at getRelativePath (VPAlgoliaSearchBox.vue:88:30)
```

when absolute URLs are provided the fix doesn't change the outcome because
```js
new URL('https://vitepress.dev/guide/what-is-vitepress#what-is-vitepress', 'https://github.com').origin
// 'https://vitepress.dev'
```

### DIFF
--- a/src/client/theme-default/components/VPAlgoliaSearchBox.vue
+++ b/src/client/theme-default/components/VPAlgoliaSearchBox.vue
@@ -84,8 +84,8 @@ function initialize(userOptions: DefaultTheme.AlgoliaSearchOptions) {
   docsearch(options)
 }
 
-function getRelativePath(absoluteUrl: string) {
-  const { pathname, hash } = new URL(absoluteUrl)
+function getRelativePath(url: string) {
+  const { pathname, hash } = new URL(url, location.origin)
   return (
     pathname.replace(
       /\.html$/,


brc-dd:
LGTM. Quick question though, does the docsearch package works with netlify crawler as well? i.e. does simply providing appId, apiKey, indexName from algolia for netlify works with VitePress?

frederic-schwarz:
> LGTM. Quick question though, does the docsearch package works with netlify crawler as well? i.e. does simply providing appId, apiKey, indexName from algolia for netlify works with VitePress?

yes exactly, providing appId, apiKey and indexName is enough to make it work. It also works fine with vuepress-next docsearch plugin because they parse the hit.url [differently](https://github.com/vuepress/vuepress-next/blob/main/packages/shared/src/utils/resolveRoutePathFromUrl.ts#L1)

PR #2963, opened 2023-09-14
fix(build): respect srcExclude in content loader

huluobotx:
add srcExclude to createContentLoader


brc-dd:
On second thought, people may want to have data being extracted from excluded directories 👀 Gonna revert this. You should use negated patterns, example:

```ts
['**/*.md', '!**/excluded']
```

Refer https://github.com/mrmlnc/fast-glob#how-to-exclude-directory-from-reading

PR #2964, opened 2023-09-14
Update site-config.md

SharePointing:
Added reference for the outline default to be changed site wide


brc-dd:
`outline` is a part of the default theme, not siteConfig and it's already documented - https://vitepress.dev/reference/default-theme-config#outline

PR #2974, opened 2023-09-15
feat(theme): allow forcing dark mode

### LINKED ISSUE #2960: How to force dark mode on

brc-dd:
closes #2960

### DIFF
--- a/docs/reference/site-config.md
+++ b/docs/reference/site-config.md
@@ -426,7 +426,7 @@ When set to `true`, the production app will be built in [MPA Mode](../guide/mpa-
 
 ### appearance
 
-- Type: `boolean | 'dark' | import('@vueuse/core').UseDarkOptions`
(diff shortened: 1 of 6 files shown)


docs/reference/site-config.md
```diff
@@ -426,7 +426,7 @@ When set to `true`, the production app will be built in [MPA Mode](../guide/mpa-
 
 ### appearance
 
-- Type: `boolean | 'dark' | import('@vueuse/core').UseDarkOptions`
+- Type: `boolean | 'dark' | 'force-dark' | import('@vueuse/core').UseDarkOptions`
```
Borange, line 429: Shouldn't this be in description below as well? Or is it intended not to show there.

brc-dd: I don't think anyone should have a confusion with what `force-something` means, but yeah we can add that in the docs.

PR #3014, opened 2023-09-25
style(theme): include configurable title attribute to custom social link icons

rin-w:


brc-dd:
> I just realized the built in icons have their titles embedded in the svg

this is the recommended way - refer https://github.com/vuejs/vitepress/pull/1325#issuecomment-1243585393

Issue #3041, opened 2023-10-02
Global Custom Container Labels


djdjz7:
### Is your feature request related to a problem? Please describe.

Since I’m writing in Chinese, I have to set the default container titles such as INFO, WARNING, TIP, DANGER, to their corresponding Chinese translations every time.

### Describe the solution you'd like

Will it be possible to add a config entry in config.ts to customize these labels globally?

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

PR #3044, opened 2023-10-03
feat: global custom container header

djdjz7:
As mentioned in #3041 , this comment adds global custom container header entry.
Usage:
```ts
// config.ts
export default defineConfig({
// ...
  markdown: {
    math: true,
    container: {
      tipLabel: "提示",
      warningLabel: "警告",
      dangerLabel: "危险",
      infoLabel: "信息",
      detailsLabel: "详细信息"
    }
  },
// ...
})
```

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -20,7 +20,7 @@ import attrsPlugin from 'markdown-it-attrs'
 import emojiPlugin from 'markdown-it-emoji'
 import type { ILanguageRegistration, IThemeRegistration } from 'shiki'
 import type { Logger } from 'vite'
-import { containerPlugin } from './plugins/containers'
+import { containerPlugin, type ContainerOptions } from './plugins/containers'
 import { highlight } from './plugins/highlight'
 import { highlightLinePlugin } from './plugins/highlightLines'
 import { imagePlugin } from './plugins/image'
@@ -57,6 +57,7 @@ export interface MarkdownOptions extends MarkdownIt.Options {
   cache?: boolean
   component?: ComponentPluginOptions
   math?: boolean | any
+  container?: ContainerOptions
 }
 
 export type MarkdownRenderer = MarkdownIt
@@ -95,7 +96,7 @@ export const createMarkdownRenderer = async (
     .use(highlightLinePlugin)
     .use(preWrapperPlugin, { hasSingleTheme })
     .use(snippetPlugin, srcDir)
-    .use(containerPlugin, { hasSingleTheme })
+    .use(containerPlugin, { hasSingleTheme }, options.container)
     .use(imagePlugin)
     .use(
       linkPlugin,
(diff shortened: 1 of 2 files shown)


brc-dd (approved):

pberto:
Hi @brc-dd,  @djdjz7 

I'd like to suggest some improvements here.

The default behavior of custom boxes is that if you type:

```
:::info
blah blah
:::
```

A custom box with the default label `INFO` (or with the "custom default label" provided by this very new feature in config.ts) will be rendered.

Now, I just tried to use this feature and instead of a custom label text use an icon, this works too:

image

```
  markdown: {
    math: true,
    container: {
      infoLabel: '<iconify-icon icon="jam:info" />'
    }
  }
```

But what I believe we should really have is both behaviors, that is being able to use a custom label icon but still be able to append words for the title in markdown to obtain:

image

pseudo md code:

```
:::info info
blah blah
:::

:::danger Windows Threading
blah blah
:::
```

Currently the only way to obtain this is doing the following (which is ugly):

```
:::info <iconify-icon icon="jam:info" /> Info 
blah blah
:::

:::danger <iconify-icon icon="jam:triangle-danger" /> Windows Threading 
blah blah
:::
```

So to recap this feature should be designed this way:

```
// ...
  markdown: {
    math: true,
    container: {
      infoLabel: "信息", // default is INFO when not specified
      infoIcon": "ⓘ" // default is empty string when not specified
    }
  },
// ...
```
This will allow to produce:

`:::info` --> `INFO` (where the default INFO can be overridden in config)
`:::info MyInfo` --> `MyInfo` (where the default INFO is overridden with MyInfo inline)
`:::info` --> `ⓘ` (where ⓘ was defined in config)
`:::info Myinfo` --> `ⓘ MyInfo` (where ⓘ was defined in config and the default INFO is overridden with MyInfo inline)

Last but not least I believe when an icon is specified text should be indented this way (I am trying to find where this could be defined btw...)

image

brc-dd:
Visual indicators should be added via CSS not markdown.

pberto:
That would be fine but not sure how to achieve it, let's say I want to have a (different) icon at the left side of the second paragraph, according to the type of block (note: the first paragraph is the title):

```
.info .custom-block> p:nth-child(2):before{
  content: "";
  display: block;
  height: 1rem;
  left: 1rem;
  -webkit-mask-image: url(/infoicon.png);
  mask-image: url(/infoicon.png);
  -webkit-mask-position: center;
  mask-position: center;
  -webkit-mask-repeat: no-repeat;
  mask-repeat: no-repeat;
  -webkit-mask-size: contain;
  mask-size: contain;
  position: absolute;
  top: 1.5rem;
  width: 1rem;
}
```

brc-dd:
Refer https://github.com/vuejs/theme/blob/HEAD/src/core/styles/vt-doc-custom-blocks.css (change vt-doc to vp-doc)

pberto:
Ok , I solved it with css.

PR #3046, opened 2023-10-05
fix: textual code copy/paste feature

bojanrajh:
This PR fixes the copy/paste functionality for `text`ual codeblocks.

See [broken example](https://stackblitz.com/edit/vite-tcb2bk?file=docs%2Findex.md) - this PR fixes it.

### DIFF
--- a/src/client/app/composables/copyCode.ts
+++ b/src/client/app/composables/copyCode.ts
@@ -23,6 +23,10 @@ export function useCopyCode() {
           .forEach((node) => (text += (node.textContent || '') + '\n'))
         text = text.slice(0, -1)
 
+        if (text.length === 0) {
+          text = sibling.textContent
+        }
+
         if (isShell) {
           text = text.replace(/^ *(\$|>) /gm, '').trim()
         }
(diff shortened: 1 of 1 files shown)


bojanrajh:
@brc-dd Please, take a look at this one.

brc-dd:
Hi, your example seems to be working fine for me. I'm able to copy paste both the code blocks there fine. Can you update that to show what exactly is breaking?

bojanrajh:
I swear that was broken last week and I was able to fix the issue locally. See production [example](https://developer.shopware.com/docs/guides/plugins/apps/app-base-guide.html#file-structure) ([source](https://github.com/shopware/docs/blob/main/guides/plugins/apps/app-base-guide.md#file-structure)), [example 2](https://developer.shopware.com/docs/guides/hosting/infrastructure/reverse-http-cache.html#hard-purge), [example 3](https://developer.shopware.com/docs/guides/integrations-api/general-concepts/generated-reference.html#swagger-ui), [example 4](https://developer.shopware.com/docs/guides/plugins/plugins/plugin-fundamentals/using-composer-dependencies.html#adding-private-composer-dependencies) - all using `txt` or `text` language and not being transformed to `span` elements.

image

I'll try to find a better reproduction, maybe there's an issue with `shiki`.

brc-dd:
Something seems wrong with the blocks on your prod site. For some reason the adaptive theme blocks are being generated there when clearly you have single theme for code blocks. Is the config of your docs public?

PR #3055, opened 2023-10-08
docs: update routing.md,append the origin of “pure.html”

edgexie:
append the origin of pure.html


bluwy:
I don’t think we need this as it doesn’t necessarily need to come from a public directory

PR #3060, opened 2023-10-09
fix: add declaration type for exclude option

jeremyriverain:
Hi,

I propose this work to fix the type declaration of the `exclude` property.

Capture d’écran du 2023-10-09 14-27-10

Here is the result after adding the type declaration:

Capture d’écran du 2023-10-09 14-27-45


brc-dd:
Hi, this was removed in favor of `_render` - https://vitepress.dev/reference/default-theme-search#example-excluding-pages-from-search -- it now allows better customization and you can use node-only APIs there as well.

PR #3066, opened 2023-10-11
fix: correct vitepress default output path and add pnpm yml

manyuemeiquqi:
First, vite press defualt output path is 
image
`.vitepress` , which isn't is `docs/.vitepress `.
Secondly, if use pnpm delopy.yml, document shuould add `      #   with:
      #     version: latest `, without this , action will cause build error.
image

### DIFF
--- a/docs/guide/deploy.md
+++ b/docs/guide/deploy.md
@@ -110,7 +110,7 @@ Note: the `vercel.json` file should be placed at the root of your **repository**
 Set up a new project and change these settings using your dashboard:
 
 - **Build Command:** `npm run docs:build`
-- **Output Directory:** `docs/.vitepress/dist`
+- **Output Directory:** `.vitepress/dist`
 - **Node Version:** `18` (or above)
 
 ::: warning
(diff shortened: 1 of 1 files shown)


brc-dd:
Regarding first, that page explicitly lists:

The following guides are based on some shared assumptions:
- The VitePress site is inside the `docs` directory of your project.

Since you've that at root instead of `docs`, it's your responsibility to update the paths.

Regarding second, you should specify `"packageManager": "pnpm@8.9.0"` in your `package.json`. `version` in workflow file is not required.

manyuemeiquqi:
Firstly, as a documentation guide, I don't think it's a good idea to have a discrepancy between the declaration in the document header and the actual output of the internal script. The assumptions based on historical legacy can lead to the maintenance of the documentation revolving around those assumptions. Secondly, when developers create a project using pnpm add -D vitepress and npx vitepress init, the resulting project's package.json does not have the "packageManager": "pnpm@8.9.0" field, and there is no relevant explanation in develop.yml. I find this a bit frustrating.

PR #3111, opened 2023-10-21
docs: modify the example usage of postcssIsolateStyles

luohuidong:


brc-dd:
Can you explain why this is needed? It should working fine for me without spreading.

luohuidong:
If we don't use the spreading, maybe the `plugins` should be an array not a object: 

```js
import { postcssIsolateStyles } from "vitepress";

export default {
  plugins: [
     postcssIsolateStyles()
  ],
};
```

luohuidong:
If the plugins is an object, I think it is a syntax error：

image

brc-dd:
Object syntax won't work. It needs key to be "a Node.js module name, a path to a JavaScript file that is relative to the directory of the PostCSS config file, or an absolute path to a JavaScript file". postcssIsolateStyles is a named export from vitepress. Object format cannot handle that.

luohuidong:
I agree it should be an array instead, so i close this pr.

brc-dd:
I updated that here itself. No need to create separate PR. Thanks

PR #3123, opened 2023-10-22
feat(theme): don't print navbar when disabled in frontmatter (narrow screens)

### LINKED ISSUE #3086: Navbar visible on print when navbar param equals false
### Describe the bug

When I hide the navbar (parameter "navbar: false") it works correctly on normal browser navigation, but when I print a page something is wrong.

On the printed page the navbar is not hide and is shown with the text "Return to top" inside.

### Reproduction

Create a new vitepress project.

In the index.md file, write the following code:
```md
---
navbar: false
---

# Hello
```

Run vitepress dev, brose to the site, call print (Ctrl + P) and print the page or save as pdf.

### Expected behavior

The navbar must be hide also on pdf.

### System Info

```Text
System:
    OS: macOS 12.6.9
    CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
    Memory: 34.23 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.18.1 - ~/.nvm/versions/node/v18.18.1/bin/node
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.1/bin/npm
  Browsers:
    Chrome: 118.0.5993.70
    Safari: 16.6
  npmPackages:
    vitepress: ^1.0.0-rc.21 => 1.0.0-rc.21
```

### Additional context

_No response_

### Validations

userquin:
This PR includes a global style for `screen-only`: we cannot hide `VPLocalNav`  on narrow screens (we lost the navigation), `screen-only` class added when disabled in frontmatter `navbar: false`.

**missing navbar**
imagen

**navbar: false**
imagen

closes #3086

### DIFF
--- a/src/client/theme-default/styles/base.css
+++ b/src/client/theme-default/styles/base.css
@@ -250,9 +250,3 @@ mjx-container {
 mjx-container > svg {
   margin: auto;
 }
-
-@media print {
-  .screen-only {
-    display: none !important;
-  }
-}
--- a/src/client/theme-default/styles/base.css
+++ b/src/client/theme-default/styles/base.css
@@ -250,3 +250,9 @@ mjx-container {
 mjx-container > svg {
   margin: auto;
 }
+
+@media print {
+  .screen-only {
+    display: none !important;
+  }
+}
--- a/src/client/theme-default/components/VPDocOutlineDropdown.vue
+++ b/src/client/theme-default/components/VPDocOutlineDropdown.vue
@@ -23,7 +23,7 @@ onContentUpdated(() => {
 </script>
 
 <template>
-  <div class="VPDocOutlineDropdown" v-if="headers.length > 0">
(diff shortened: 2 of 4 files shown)


userquin:
I need to test this PR on large screen, on my laptop the print preview using narrow screen: maybe we need to also include the logic in the `VPNavBar`

src/client/theme-default/styles/base.css
```diff
@@ -250,3 +250,9 @@ mjx-container {
 mjx-container > svg {
   margin: auto;
 }
+
+@media print {
+  .screen-only {
```
brc-dd, line 255: maybe move this to src/client/theme-default/styles/utils.css ?

userquin: done

userquin:
@brc-dd I can prepare a new PR to include print options, it will go to the branch of this PR (title will be `feat(theme): add print options` and will respect fm sidebar and navbar options when disabled), what do you think?

Issue #3136, opened 2023-10-24
allow siteConfig.extends to be an array


jcbhmr:
### Is your feature request related to a problem? Please describe.

problem: extending from _two_ premade "plugin-like" presets: blog & jsdoc

```js
// config.ts
export default defineConfig({
  // for blog preset
  extends: blog({ rss: true }),
  // OR
  // for jsdoc preset
  extends: jsdoc({ conf: "jsdoc.json" }),
  // how do i combine these?
})
```

of course, i know i _could_ do it manually, but it'd be nice to support an array like below

### Describe the solution you'd like

```js
// config.ts
export default defineConfig({
  extends: [
    blog({ rss: true }),
    jsdoc({ conf: "jsdoc.json" })
  ]
})
```

### Describe alternatives you've considered

using a vite plugin like

```js
// config.ts
export default defineConfig({
  vite: {
    plugins: [
      blog({ rss: true }),
      jsdoc({ conf: "jsdoc.json" })
    ]
  }
})
```

problem is this doesn't get you access to the _pre-normalized_ config and only lets you mutate the post-normalized config.vitepress.* config stuffs in the vite plugin hooks (if that's even supported by vitepress officially 😬)

```js
// myplugin.js
return {
  name: "my-vitepress-blog-plugin",
  config(config) {
    console.log(config.vitepress.sidebar)
    // hacky mutations to the vitepress site config object here
  }
}
```

### Additional context

i think extending a premade config is a pretty good way to go one step higher than just a vite plugin in order to configure vitepress-specific things like sidebars and such. being able to mix multiple of these base configs would be nice imo

if theres another good way to go about the pattern demo'ed above 👆 then id be happy to hear it 😊

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

jcbhmr:
https://github.com/vuejs/vitepress/blob/c343d938fccd878134c8b9dbf2b12dce226a1c19/src/node/config.ts#L178-L187

could be something like this idk

```ts
async function resolveConfigExtends(
  config: RawConfigExports
): Promise<UserConfig> {
  const resolved = await (typeof config === 'function' ? config() : config)
  if (resolved.extends) {
[...]

PR #3143, opened 2023-10-26
export `mergeConfig()`

jcbhmr:
related to #3136


brc-dd:
you'll need to re-export if from node/index.ts too

PR #3148, opened 2023-10-27
fix: "VPNavScreenMenuGroup" component HTML not supported

akirarika:
For the visitor VitePress on the desktop, the TEXT of NAV written by NAV supports any HTML tags, which can make us achieve some simple effects, such as bold and corner markers

Snipaste_2023-10-28_04-22-07

But when accessing on the mobile terminal, HTML was showed out

Snipaste_2023-10-28_04-21-54

This PR keeps them consistent


akirarika:
@brc-dd I apologize for disturbing you, I would like to know if my pull request can be merged.

English is not my native language, so I may not have expressed myself clearly enough. If you need further explanation about this PR, please let me know.

Please !! ><

PR #3170, opened 2023-11-02
perf: overall prevention of duplicate rendering in localSearch

zonemeen:
The cache should be defined outside of `debouncedWatch`, otherwise a new cache will be redefined every time when `debouncedWatch` is retriggered. It's a bit of a waste of performance.

### DIFF
--- a/src/client/theme-default/components/VPLocalSearchBox.vue
+++ b/src/client/theme-default/components/VPLocalSearchBox.vue
@@ -28,6 +28,7 @@ import {
   watchEffect,
   type Ref
 } from 'vue'
+import { LRUCache } from 'lru-cache'
 import type { ModalTranslations } from '../../../../types/local-search'
 import { pathToFile } from '../../app/utils'
 import { useData } from '../composables/data'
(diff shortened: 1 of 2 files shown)


brc-dd:
This will work, but will also cause memory leaks and that cache will keep on growing. We probably should LRU cache here with some max limit like 256 anchors. Instead of using lru-cache (which adds up extra 15kB), IG we simply can do something like https://stackoverflow.com/a/46432113/11613622

zonemeen:
LRU cache is indeed more suitable for this scene🤔. I'll change it later when I have time.

zonemeen:
@brc-dd Nice work on the next step! There's also a bit of a question here about whether 16 cache entries would be a bit too few.

brc-dd:
Ah, 16 here actually means 16 files in memory. So there will be like 16 maps having anchor to rendered html matching which will already be quite big. And I tested this on our docs, the cache misses were very few.

zonemeen:
Ah, got it! 🤗

PR #3181, opened 2023-11-04
docs: always use 'extends' to extend themes

skirtles-code:
These 3 examples were still using `...DefaultTheme` to extend the default theme, rather than `extends: DefaultTheme`.

These examples aren't using `enhanceApp`, so they do work in isolation, but the use of `...DefaultTheme` makes it less clear how to combine the various examples into a single configuration.

I'm not aware of any reason why we'd want to show `...DefaultTheme` in the docs.

### DIFF
--- a/docs/guide/extending-default-theme.md
+++ b/docs/guide/extending-default-theme.md
@@ -135,7 +135,7 @@ import DefaultTheme from 'vitepress/theme'
 import MyLayout from './MyLayout.vue'
 
 export default {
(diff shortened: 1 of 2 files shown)


brc-dd:
> I'm not aware of any reason why we'd want to show ...DefaultTheme in the docs.

Ah, that's old code. `extends` wasn't there earlier.

brc-dd (approved):

PR #3215, opened 2023-11-17
fix(ally): clear up confusion with the title of theme switch

### LINKED ISSUE #3214: Theme: the title of theme switch
### Describe the bug

Currently, the title of the theme switch is `toggle dark mode` whether in light mode or dark mode.
image

image

### Reproduction

https://vitepress.dev/guide/getting-started

### Expected behavior

In light mode, the title should be `toggle dark mode`; in dark mode, the title should be `toggle light mode`

### System Info

```Text
Both in Mac or Windows
```

### Additional context

_No response_

### Validations

zh-lx:
fix #3214

### DIFF
--- a/src/client/theme-default/components/VPSwitchAppearance.vue
+++ b/src/client/theme-default/components/VPSwitchAppearance.vue
@@ -12,7 +12,7 @@ const toggleAppearance = inject('toggle-appearance', () => {
 })
 
 const switchTitle = computed(() => {
-  return isDark.value ? 'Switch to light mode' : 'Switch to dark mode'
+  return isDark.value ? 'Switch to light theme' : 'Switch to dark theme'
 })
 </script>
 
--- a/src/client/theme-default/components/VPSwitchAppearance.vue
+++ b/src/client/theme-default/components/VPSwitchAppearance.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { inject } from 'vue'
+import { inject, computed } from 'vue'
 import { useData } from '../composables/data'
 import VPSwitch from './VPSwitch.vue'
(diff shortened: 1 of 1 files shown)


brc-dd:
Ah, this is bit subjective, but when we write "toggle dark mode", we mean toggle the state of dark mode to its opposite. It's different from "toggle to dark mode". In that case you suggestion would've made sense. We can make it something like "Switch to light theme"/"Switch to dark theme" instead if the current title is bit confusing.

src/client/theme-default/components/VPSwitchAppearance.vue
```diff
@@ -10,11 +10,15 @@ const { isDark } = useData()
 const toggleAppearance = inject('toggle-appearance', () => {
   isDark.value = !isDark.value
 })
+
+const switchTitle = computed(() => {
+  return isDark.value ? 'toggle light mode' : 'toggle dark mode'
```
brc-dd, line 15:
```suggestion
  return isDark.value ? 'Switch to light mode' : 'Switch to dark mode'
```

src/client/theme-default/components/VPSwitchAppearance.vue
```diff
@@ -10,11 +10,15 @@ const { isDark } = useData()
 const toggleAppearance = inject('toggle-appearance', () => {
   isDark.value = !isDark.value
 })
+
+const switchTitle = computed(() => {
+  return isDark.value ? 'Switch to light mode' : 'Switch to dark mode'
```
brc-dd, line 15:
```suggestion
  return isDark.value ? 'Switch to light theme' : 'Switch to dark theme'
```

Issue #3226, opened 2023-11-19
Browser forword&backword not trigger onAfterRouteChanged


maicss:
### Describe the bug

Browser forword&backword not trigger `router.onAfterRouteChanged`

### Reproduction

https://github.com/maicss/website/blob/main/.vitepress/theme/GiscusLayout.vue#L161-L170

When click forword or backward of browser, keeping print 111 but not 222

### Expected behavior

Should trigger `router.onAfterRouteChanged`

### System Info

```Text
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
    Memory: 14.63 GB / 23.94 GB
  Binaries:
    Node: 20.9.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.2.3 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.10.5 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Spartan (44.19041.3636.0), Chromium (119.0.2151.72)
  npmPackages:
    vitepress: 1.0.0-rc.28 => 1.0.0-rc.28
```
```


### Additional context

_No response_

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #3227, opened 2023-11-20
fix(client): No onAfterRouteChanged called after popstate

dy-xiaodong2022:
This Pull Request Fix issue #3226: No onAfterRouteChanged called after popstate

In `src/client/app/router.ts` popstate eventListener, it should call `onAfterRouteChanged` hook after `loadPage`,
but it is missing.

```diff
...
    window.addEventListener('popstate', (e) => {
      loadPage(
        normalizeHref(location.href),
        (e.state && e.state.scrollPosition) || 0
-      )
+      ).then(() => {
+        router.onAfterRouteChanged?.(location.href)
+      })
    })
...
```

Before Fix:
Screenshot 2023-11-20 at 9 26 52 PM

After Fix:
Screenshot 2023-11-20 at 9 29 02 PM

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -226,8 +226,9 @@ export function createRouter(
       loadPage(
         normalizeHref(location.href),
         (e.state && e.state.scrollPosition) || 0
-      )
-      router.onAfterRouteChanged?.(location.href)
+      ).then(() => {
+        router.onAfterRouteChanged?.(location.href)
+      })
     })
 
     window.addEventListener('hashchange', (e) => {
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -227,6 +227,7 @@ export function createRouter(
         normalizeHref(location.href),
         (e.state && e.state.scrollPosition) || 0
       )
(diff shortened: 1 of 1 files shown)


src/client/app/router.ts
```diff
@@ -226,7 +226,9 @@ export function createRouter(
       loadPage(
         normalizeHref(location.href),
         (e.state && e.state.scrollPosition) || 0
-      )
+      ).then(() => {
+        router.onAfterRouteChanged?.(location.href)
```
sxzz, line 230: Use await expression instead
```suggestion
      await router.onAfterRouteChanged?.(location.href)
```

dy-xiaodong2022: Change:
```diff
-    window.addEventListener('popstate', (e) => {
-      loadPage(
+    window.addEventListener('popstate', async (e) => {
+      await loadPage(
        normalizeHref(location.href),
        (e.state && e.state.scrollPosition) || 0
-      ).then(() => {
-        router.onAfterRouteChanged?.(location.href)
       )
+      router.onAfterRouteChanged?.(location.href)
    })
```

PR #3233, opened 2023-11-22
Update markdown.md

dannysmc95:
Removed double spaces in some of the `[!code` tags, as when copying from the docs, it doesn't work, as there are two spaces.


brc-dd:
read the note just above that example 🤦‍♂️

dannysmc95:
@brc-dd My bad, we copied from there and was like "why is it not working"

Issue #3264, opened 2023-11-27
Regression: Router breaks navigation to custom home URL


elringus:
### Describe the bug

When using custom home page and `themeConfig.logoLink` is set to that paget, after any navigation happens in the docs, clicking the logo loads 404 page. See video below for a demo.

https://github.com/vuejs/vitepress/assets/2056864/69241914-58ab-4400-9013-7d9379cb4c14

### Reproduction

Not sure how reproduce this in a sandbox (the docs have to be deployed), but here are the sources of the demo above: https://github.com/naninovel/Documentation/tree/master/docs

Notice, that the bug only kicks in after some navigation happens in the docs; eg, it won't reproduce if you open the docs and immediately click the logo (in this case the home page will load correctly).

### Expected behavior

Clicking on logo should load custom home page.

### System Info

```Text
System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 Intel(R) Core(TM) i7-6900K CPU @ 3.20GHz
    Memory: 52.30 GB / 63.92 GB
  Binaries:
    Node: 20.5.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.6.6 - C:\Program Files\nodejs\npm.CMD
    pnpm: 7.29.3 - C:\ProgramData\chocolatey\bin\pnpm.EXE
  Browsers:
    Chrome: 119.0.6045.160
    Edge: Spartan (44.19041.3636.0), Chromium (119.0.2151.72)
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
should be fixed in vitepress 1.0.0-rc.29

elringus:
I've actually tested this on the latest rc.31 (in a [branch](https://github.com/naninovel/Documentation/tree/feat/optimize)). 

Just to make sure I've updated to 31 on the main branch and deployed — still the same issue.

zonemeen:
Try changing the `logoLink: "https://naninovel.com"` in your configuration to `logoLink: "//naninovel.com"` or `logoLink: "naninovel.com"`.

elringus:
[...]

PR #3271, opened 2023-11-29
fix: custom home page with setting `logoLink` leads to 404 page

zonemeen:
temporary fix for #3264

We can circumvent this case in the `router`, which is that when using a custom homepage and `themeConfig.logoLink` is set to that custom homepage, clicking on the logo does not try to load the vitepress-generated homepage, but instead jumps directly to the custom homepage.

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -183,7 +183,10 @@ export function createRouter(
           )
 
           // temporary fix for #3264
-          if (siteDataRef.value.themeConfig?.logoLink === origin) return
+          if (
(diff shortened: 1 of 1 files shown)


elringus:
I've tried this change and (while it fixes the logo navigation) page now reloads on every navigation (including inside the app), but weirdly only on Firefox (Chrome is fine). Any idea why Firefox may behave like that?

elringus:
Actually, it happens on all browsers, just was more noticeable on Firefox. Hopefully, there is a better solution to this issue, as reloading the page on local navigation significantly affects UX.

elringus:
Nevermind, I've just messed the regex. The fix is working. Sorry for the noise.

Issue #1047, opened 2022-07-22
[Docs] Explain how to use TypeScript in a VitePress project


staghouse:
### Is your feature request related to a problem? Please describe.

Im currently migrating a VuePress 1 project over to VitePress and I would like to integrate TypeScript for my .vue files. The documentation does not explain how to enable TypeScript in a VitePress project. I've looked over other documentation to enable TS in a Vue 3 project but that was based on vue-cli and seems irrelevant (I could be wrong)

My IDE, VS Code, has `Vue Language Features (Volar)` and `TypeScript Vue Plugin (Volar)` installed.

Here is my eslintrc.js that was migrated over from VuePress (if relevant?)

```js
module.exports = {
  parser: "vue-eslint-parser",
  parserOptions: {
    ecmaVersion: 2018
  },
  env: {
    node: true,
    browser: true,
    commonjs: true,
    jest: true,
    es6: true
  },
  extends: [
    "eslint:recommended",
    "plugin:vue/recommended"
  ],
  plugins: [
    "vue"
  ],
  ignorePatterns: ["dist/", "node_modules/", "*.json", ".temp/"],
  rules: {
    // Rules to ignore
    "vue/multi-word-component-names": "off",
    "no-use-before-define": "off",
    "no-case-declarations": "off",
    "no-useless-escape": "off",
    "no-control-regex": "off",
    "no-unused-vars": "off",
    "no-console": "off",
    "no-empty": "off",
    "no-prototype-builtins": "off",
    "guard-for-in": "off",
    // Rules to warn
    indent: ["warn", 2, { SwitchCase: 1 }],
    // Rules to error
    curly: "error",
    "semi-spacing": "error",
    "no-unneeded-ternary": "error",
    "no-useless-concat": "error",
    "no-useless-return": "error",
    "no-multi-spaces": "error",
    "no-alert": "error",
    "no-eq-null": "error",
    "no-eval": "error"
  }
};
```

Simple, when assigning a type like:

`const str: string = ''` nets an error: `error  Parsing error: Unexpected token :`...

### Describe the solution you'd like

A section to describe what needs to be done to get TypeScript working in VitePress

### Describe alternatives you've considered

Looked at other documentation about Vue 3, TypeScript implementation, etc. But could not find anything relating to VitePress. I've also looked at other projects that use VitePress and I am just not understanding what is making their files get statically type checked

### Additional context

[...]

PR #3291, opened 2023-12-05
feat(config): add `themeDir` option

Waleed-KH:
Together with #3290 I want to achieve project structure like the following:
```
.
├─ docs                    # project root
│  ├─ src
│  │  └─ theme
│  │     └─ index.js       # theme entry
│  ├─ vitepress.config.ts  # config file
│  └─ index.md
└─ package.json
```

The main motivation: working with private folder (`.vitepress`) cause a-lot of issues with TypeScript and eslint (like motioned in #1047), even when you get the typechecker and the linter to work with `.vitepress`, the code editor (VSCode) fails to evaluate the the files correctly.

Also, the above structure is more familiar to Vite, I hope it would be the default for VitePress

### DIFF
--- a/docs/reference/site-config.md
+++ b/docs/reference/site-config.md
@@ -437,6 +437,19 @@ export default {
 }
 ```
 
+### themeDir
+
+- Type: `string`
+- Default: `./.vitepress/theme` if exists, otherwise [Default Theme](../guide/extending-default-theme) path.
+
+The directory of the [theme entry file](../guide/custom-theme#theme-resolving), relative to [project root](../guide/routing#root-and-source-directory).
+
+```ts
+export default {
+  themeDir: './awesome-theme'
+}
+```
+
 ### ignoreDeadLinks
 
 - Type: `boolean | 'localhostLinks' | (string | RegExp | ((link: string) => boolean))[]`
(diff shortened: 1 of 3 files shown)


zhangyx1998:
Would it be better to include this as a commandl ine argument? e.g. `npx vitepress dev docs --entry docs/src`

In this way you can even specify a directory outside docs: `npx vitepress dev docs --entry src`

zhangyx1998:
Not really. I meant the alternative path to `.vitepress`. That's where vitepress should look for user provided `config.mjs`. This is impossible to live inside user config.

Currently vitepress always assumes user config entry to be `[docRoot]/.vitepress/config.(js|mjs|cjs)`. It's hard coded [here](https://github.com/vuejs/vitepress/blob/d279e63cb4d417420cdc3fb3e6e03c96b777289f/src/node/config.ts#L29) and [here](https://github.com/vuejs/vitepress/blob/d279e63cb4d417420cdc3fb3e6e03c96b777289f/src/node/config.ts#L164-L169)

zhangyx1998:
It's just some random thought, please feel free to discard it.

Waleed-KH:
That might be a good option, there are already option to configure the `outDir` and `cacheDir` which uses `.vitepress` by default, only the `theme` dir and the config file path are not configurable, I've created [another PR](https://github.com/vuejs/vitepress/pull/3290) to specify the config file location, but it's been a month with no response from the maintainers.

PR #3312, opened 2023-12-09
feat: fall back if translation is missing

babakfp:
This PR helps to fall back to a different locale if a page in the current locale is missing. This feature can be enabled by setting the `localesFallback` option to `true`. All locales will fall back to the default locale (root), which can be changed by using the `localesDefaultFallback` option (and setting the value to the key of a locale). The `fallback` option is available for each locale to have its custom locale to fall back to.

## Example 1

```js
{
    locales: {
        root: { lang: 'en' },
        fa:   { lang: 'fa' },
    },
    localesFallback: true,
}
```

1. If `/fa/a` was not found, fall back to `/a`.
2. If `/a` was not found, show `404`.

- `root` is the default locale to fall back to.
- If `localesFallback` option was set to `false`, things would continue to work like this PR never happened.

## Example 2

You can set a custom locale to fall back to for each locale.

```js
{
    locales: {
        en: { lang: 'en', },
        fa: { lang: 'fa', fallback: 'en' },
    },
    localesFallback: true,
}
```

1. if `/fa/a` was not found, fall back to `/en/a`.
2. if `/en/a` was not found, show `404`.

- `root` can have its `fallback` option too.
- A locale can't fall back to itself.

## Example 3

Change the default locale that all locales can fall back to.

```js
{
    locales: {
        en: { lang: 'en' },
        fa: { lang: 'fa' },
    },
    localesDefaultFallback: 'en',
    localesFallback: true,
}
```

1. if `/fa/a` was not found, fall back to `/en/a`.
2. if `/en/a` was not found, show `404`.

- A locale can't fall back to itself.

## Example 4

```js
{
    locales: {
        en: { lang: 'en' },
        fa: { lang: 'fa', fallback: 'tr' },
        tr: { lang: 'tr', fallback: 'en' },
    },
    localesFallback: true,
}
```

1. if `/fa/a` was not found, fall back to `/tr/a`.
4. if `/tr/a` was not found, fall back to `/en/a`.
5. if `/en/a` was not found, show `404`.

---

## Helpful errors

- If the `fallback` option was set to `'root'`, you will get an error, because it's the default value and there is no reason to do it.
    - The error message is: "Invalid VitePress Config: A locale (the-locale-key), cannot fall back to (root).".
- If the `fallback` option was set to its locale key, you will get an error, because this will create a loop.
    - The error message is: "Invalid VitePress Config: A locale (the-locale-key), cannot have a fallback to itself."
- If the `fallback` option was set to a locale key that didn't exist, you will get an error, because, I guess if you have a locale, it should be configured in VitePress Config too.
    - The error message is: "Invalid VitePress Config: A locale (the-locale-key), cannot have a fallback to a non-existing locale."

---

I fixed the issue reported [here](https://github.com/vuejs/vitepress/pull/3312#issuecomment-1951426608).


babakfp:
Ops, let me add the description...

hamishwillee:
The only thing that occurs to me is whether search needs to consider the fact that any "not found" topics should be taken from the fallback trees.

altrusl:
I support this PR
Often in multilanguage websites there are some on purpose missing section translations but Google Search Console and other SEO tools downgrade rating for 404 pages
Language selector should be disabled on those pages or it should fall back to existing translations

liana-p:
I'm also very interested in this being merged. I was just in the process of starting to localise a docs website and realised there's no good way to work around the language picker 404ing if the page doesn't exist in the other language.

It could also help reuse sidebars/navbars.

IMB11:
It seems that index pages are not working.

When going to `es_es/players/index.md`, it redirects to the english page. You can test this in the closed PR above this comment.

babakfp:
> It seems that index pages in sub-folders are not working.

Thank you for letting me know. I fixed it in the recent commits. Feel free to check it out, it's ready for review.

tyzero:
I support this PR

PR #3319, opened 2023-12-11
feat: add Issue Labeled

wChenonly:
In order to facilitate the management of issues, I think we need a workflows about Issue Labeled.

### DIFF
--- a/.github/workflows/issue-labeled.yml
+++ b/.github/workflows/issue-labeled.yml
@@ -1,0 +1,41 @@
+name: Issue Labeled
+
+on:
+  issues:
+    types: [labeled]
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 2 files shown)


kiaking:
@brc-dd 
Do you think this might help you? Personally, I think having so many labels and actions just make things cluttered, so personally I'm against adding this. We already have too many labels and bot messing around with `stale` label and such 😅

brc-dd:
With the current volume I don't think these are needed. Mostly we need to post a custom comment with need-more-info anyway, and this will just adds more noise. Other peoples from the team are using this for some of their projects, but as of now, I don't think this is needed in VP. If, in future, I see myself repeating such responses, I'll add this.

PR #3321, opened 2023-12-12
docs: fix pnpm init command

yuyinws:
Use the right [pnpm command](https://pnpm.io/cli/dlx).


brc-dd:
If you actually read that guide, it's expected that you first install vitepress in your project manually, then run that command as the init command doesn't add vitepress to deps or generate a package.json for you.

PR #3330, opened 2023-12-13
fix: GET https://cdn.usefathom.com/script.js net::ERR_SSL_PROTOCOL_ERROR

babakfp:
Screenshot 2023-12-13 153516


brc-dd:
inBrowser in config file has no effect as the config is only evaluated on server. It will always be false.

PR #3343, opened 2023-12-15
chore: use transformer to clean-up highlight result

antfu:
This would be more efficient and reliable

### DIFF
--- a/src/node/markdown/plugins/highlight.ts
+++ b/src/node/markdown/plugins/highlight.ts
@@ -79,7 +79,7 @@ export async function highlight(
     transformerNotationHighlight(),
     transformerNotationErrorLevel(),
     {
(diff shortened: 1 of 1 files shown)


src/node/markdown/plugins/highlight.ts
```diff
@@ -116,16 +122,6 @@ export async function highlight(
     }
 
     const lineOptions = attrsToLines(attrs)
-    const cleanup = (str: string) => {
-      return str
-        .replace(
-          preRE,
-          (_, attributes) =>
-            `<pre ${vPre}${attributes.replace(' tabindex="0"', '')}>`
```
antfu, line 124: Clean-up was doing two things: adding `v-pre` as well as removing styles. Split into two transformers to make it better scoped.

PR #3353, opened 2023-12-18
feat: support dir in frontmatter

Jinjiang:
Support dir for each page

Ref: https://github.com/vuejs/router/pull/2085


src/client/theme-default/Layout.vue
```diff
@@ -22,12 +22,22 @@ watch(() => route.path, closeSidebar)
...
+
+watch(() => frontmatter.value.dir, (metaDir) => {
+  if (!inBrowser) { return }
+  const pageDir = metaDir || dir.value || 'ltr'
+  // <html dir>
+  document.documentElement.dir = pageDir
+  // <body dir> and <body style="direction">
+  document.body.dir = pageDir
+  document.body.style.direction = pageDir
+}, { immediate: true })
```
posva, line 40: Shouldn't this be set at build-time? Wouldn't this create Hydration mismatches? I wonder if there is a way to only set `<body dir="auto">` and let the browser choose.

Jinjiang: Not sure neither. I feel it's hard because these elements are out of the app. Seems [transformHtml](https://vitepress.dev/reference/site-config#transformhtml)
 is another way to help? But I feel it's too tricky.

Jinjiang: @posva I tried more `dir="auto"` and found it's really tricky. It only detects the direction by its first descendant element which contains text. Unless you set `dir="auto"` in every single element, the value would probably be determined by the logo label or the top ads if you set `dir="auto"` on the `<body>` which doesn't make sense.

And based on the current implementation, I also added the same logic into the build process.

Does it look good? :-)

posva: It should be good. I only wanted to drop the ideas I had, just in case. I can't test much at the moment. so please ignore my comments if you have already thought of it 😄

Jinjiang:
@brc-dd seems this new commit doesn't work 👀 
image

brc-dd:
Ah weird. I checked this earlier on router docs. I'll see what's wrong. The dir should be added to html root itself and if body doesn't have style direction, it's automatically inferred as same of html.

brc-dd:
@Jinjiang fixed in https://github.com/vuejs/vitepress/commit/c2b4c66e79fde7479f5f43841e1921a5c220c9a5

Actually, `useData` returns `lang` and `dir` but we were accessing it directly using `site.value.dir` which doesn't take the frontmatter into account.

PR #3355, opened 2023-12-18
fix(theme): fix ineffective activeMatch configuration

Red-Asuka:
In the settings of the default theme's Navbar, we can set the `activeMatch` for Navbar items with children. However, when the regular expression does not match and one of the links in the children matches the current page, it will be highlighted incorrectly.

This is the definition of NavItemWithChildren in the default theme:

```ts
  export interface NavItemWithChildren {
    text?: string
    items: (NavItemChildren | NavItemWithLink)[]

    /**
     * `activeMatch` is expected to be a regex string. We can't use actual
     * RegExp object here because it isn't serializable
     */
    activeMatch?: string
  }
```

### DIFF
--- a/src/client/theme-default/components/VPNavBarMenuGroup.vue
+++ b/src/client/theme-default/components/VPNavBarMenuGroup.vue
@@ -24,17 +24,21 @@ const isChildActive = (navItem: DefaultTheme.NavItem) => {
 }
 
 const childrenActive = computed(() => isChildActive(props.item))
+
+const menuActive = computed(() => {
+  const { activeMatch } = props.item
+  if (activeMatch)
+    return isActive(page.value.relativePath, activeMatch, true)
+  else
+    return childrenActive.value
+})
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


Red-Asuka:
@brc-dd , could you please review the following PR? This PR fixes a bug where the theme navbar configuration does not respect the NavItem's activeMatch. I'll illustrate through a real-life example.

Let's say I have a dropdown list for versioning - I wanted to highlight only the specific version number and not the "Versions" tab. So I set an activeMatch that wouldn't practically be matched, but to no avail - "Versions" was still highlighted. I perceive this as a bug; The correct logic should foremost respect the parent menu's activeMatch.

```ts
const nav = {
  text: 'Versions',
  activeMatch: `^/no-active/`,
  items: [
    { text: 'v1.0', link: '/docs/v1' },
    { text: 'v2.0', link: '/docs/v2' },
    { text: 'v3.0', link: '/docs/v3' },
  ]
}
```

image

PR #3374, opened 2023-12-23
perf(localSearch): add concurrency pooling, cleanup logic, improve performance

### LINKED ISSUE #3377: [local search] indexing too slow to be usable for large sites
### Describe the bug

I am trying to convert an ancient documentation site into vitepress. The volume is very large (a lot of of pages, and many of them has very long content).

I waited for __*more than 4 hours*__ to build my site with local search enabled. I would have to wait for the same amount of time to launch a dev server.

Unfortunately, although the build finished without any error, the rendered search index was __*empty*__. This is caused by another problem: the local search plugin use [one _regex_](https://github.com/vuejs/vitepress/blob/main/src/node/plugins/localSearchPlugin.ts#L242) to find headers, and [another _regex_](https://github.com/vuejs/vitepress/blob/main/src/node/plugins/localSearchPlugin.ts#L243) to remove tags. Which works only if the html is directly parsed from markdown, and is very likely to fail upon embedded html (or vue template).

In addition, if you've looked at the outputs from PR #3374, you might have already noticed that each file was indexed _**twice**_ for each build. That doubles the wait time.

### Reproduction

```diff
  vitepress v1.0.0-rc.32

⠹ building client + server bundles...
(!) Some chunks are larger than 500 kB after minification. Consider:
 - Using dynamic import() to code-split the application
 - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
 - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
/Users/Yuxuan/Lab/xorg-doc/docs/index.md
✓ building client + server bundles...
✓ rendering pages...
- build complete in 16008.00s.
```

> 16,008 seconds = 4.45 hours

### Expected behavior

1. Index faster.

2. No double indexing for same source file.

3. When splitting page into sections, match headers more robustly.

### System Info

```Text
Not relevant
```

#### Also, I can provide a fix to the following issue after some discussion with maintainers:

1. On `dev server`, only provide the search-index when the search box is focused (expanded), and show a loading indicator in the search dropdown before indexing completes. In this way the indexing process will not block dev preview.

4. Fix the double indexing problem so large sites will build faster.

### Validations

zhangyx1998:
fixes #3377 

## Highlights

This PR enables the local search plugin to work in _**true**_ parallel. While keeping its API backward compatible,  it now allows asynchronous user-provided callbacks for both html generation and section splitting (which is very expensive).

In addition, by refactoring the logic of task dispatching, the plugin will no longer accumulate content for all pages and locales before indexing. Therefore, this change will benefit all _local-search_ users  **out-of-box**.

## What's next

I've implemented multiprocessing that worked with this PR. That implementation is not yet included here because (1) I am not sure whether the inclusion of a new package (JSDOM) is acceptable, and (2) I do not know if people would agree with the idea of multiprocessing.

As you can see in this [commit](https://github.com/zhangyx-lab/x-doc/commit/cd2bfb7cb80b1c2fb422764f39253186bf127fe4), the new `miniSearch.splitIntoSections()` config option allows an AsyncGenerator as its return value. And the AsyncGenerator can be used to proxy a worker thread using JS Event model.

One important reason I chose _local-search_ plugin to start experimenting is data isolation. The splitting task only needs html string, so it's easy to be stripped out of the context. In my project, the overhead of IPC is reduced to the minimum since the main process only sends a path, not the entire content.

However, if done correctly, many other processes can also be refactored to run in parallel. For example, the `renderPage()` looks very promising, which currently takes relatively long to complete, especially for large volume of content. If we can make that part parallel, I believe almost everyone would receive a noticeable speedup.

## The speedup is crazy

Both task are performed on Apple M1 Max, building [this project](//github.com/zhangyx-lab/x-doc)

> ### Before: 16008.00 seconds (4.4 hours)
> ```plaintext
>   vitepress v1.0.0-rc.32
> 
> ⠹ building client + server bundles...
> (!) Some chunks are larger than 500 kB after minification. Consider:
> - Using dynamic import() to code-split the application
> - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
> - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
> /Users/Yuxuan/Lab/xorg-doc/docs/index.md
> ✓ building client + server bundles...
> ✓ rendering pages...
> build complete in 16008.00s.
> ```

> ### After: 240 seconds (4 minutes)
> 
> ```plaintext
>   vitepress v1.0.0-rc.32
> 
> 🔍️ Indexing files for search...
> ✅ Indexing finished...
> ⠹ building client + server bundles...
> (!) Some chunks are larger than 500 kB after minification. Consider:
> - Using dynamic import() to code-split the application
> - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
> - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
> 🔍️ Indexing files for search...
> ✅ Indexing finished...
> ✓ building client + server bundles...
> ✓ rendering pages...
> build complete in 248.53s.
> ```

## Problem to be fixed

The default section splitting function `splitPageIntoSections()` uses two regular expressions to detect and stip headers from html sources. However, it assumes that each `<h*>` element has an `<a>` element as it first child. This is only true when generated by pure markdown sources. However, markdown allows html element embedding (vitepress even makes it reactive), and users would expect their own headers to be detected normally, which may not follow the RegExp patterns.

In addition, the original splitter used a sparse array to track current header hierarchies. However, it fails at some very simple cases. For example:

```markdown
.       <!-- expected | actual   -->
#   A1  <!-- A1       | A1       -->
##  A2  <!-- A1 A2    | A1 A2    -->
#   B1  <!-- B1       | B1 A2    -->
### B3  <!-- B1 B3    | B1 A2 B3 -->
```

All these problems have been fixed in the [`worker` implementation](https://github.com/zhangyx-lab/x-doc/commit/cd2bfb7cb80b1c2fb422764f39253186bf127fe4#diff-c2a81544e9e46223053faa70670178d4e45d8b590252ba092c99bedd3751f69f) from my project. I can spend some time to port them back (with or without multiprocessing).


zhangyx1998:
fixes #3377

zhangyx1998:
Force pushed to keep up with main branch (use pMap instead).

types/default-theme.d.ts
```diff
@@ -393,13 +396,34 @@ export namespace DefaultTheme {
...
+       * Overrides the default regex based page splitter.
+       * Supports async generator, making it possible to run in true parallel
+       * (when used along with `node:child_process` or `worker_threads`)
+       * ---
+       * This should be especially useful for scalability reasons.
+       * ---
+       * @param {string} path - absolute path to the markdown source file
+       * @param {string} html - document page rendered as html
+       */
+      splitIntoSections?: (
```
brc-dd, line 410: prefix this with underscore, otherwise it will be included in client side code

zhangyx1998: done

brc-dd (approved):
LGTM, I'll merge this after testing once with unocss docs or something.

zhangyx1998:
> LGTM, I'll merge this after testing once with unocss docs or something.

Sure, I will create another PR to port [the JSDOM based section splitter](https://github.com/zhangyx-lab/x-doc/commit/cd2bfb7cb80b1c2fb422764f39253186bf127fe4#diff-c2a81544e9e46223053faa70670178d4e45d8b590252ba092c99bedd3751f69f) back after this one gets merged. After that all users should receive performance benefit as well as enhanced indexing robustness.

PR #3383, opened 2023-12-28
Fix transparent nav bar

Justin3go:
## Background

When I upgraded from version beta.1 to rc.33, I found that the transparent navigation bar on the custom homepage (`layout: page`) was missing. I then looked in the changelog and found that [this PR -- dont show transparent navbar other than home](https://github.com/vuejs/vitepress/pull/2742/files) affected my display effect.

## Feature Implementation

Before: Users could use the transparent navigation bar when `layout: home`, but not on other pages.

After: Users can configure the `transparentNavBar` option in the page `frontmatter`. If set to `true`, the transparent navigation bar will be used.

## Other

The description of this option has been added to the documentation.

### DIFF
--- a/docs/reference/frontmatter-config.md
+++ b/docs/reference/frontmatter-config.md
@@ -124,6 +124,20 @@ navbar: false
 ---
 ```
 
+
+### transparentNavBar
+
+- Type: `boolean`
+- Default: `false`
+
+Whether to display the navigation bar as transparent when the scroll distance is 0. It is used by default when [`layout: home`](./default-theme-layout) is set. It can be added to other pages if needed.
+
+```yaml
+---
+transparentNavBar: true
+---
+```
+
 ### sidebar
 
 - Type: `boolean`
--- a/src/client/theme-default/components/VPNavBar.vue
+++ b/src/client/theme-default/components/VPNavBar.vue
@@ -27,9 +27,10 @@ const { frontmatter } = useData()
 const classes = ref<Record<string, boolean>>({})
 
 watchPostEffect(() => {
+  const isTransparentPage = frontmatter.value.transparentNavBar || frontmatter.value.layout === 'home' 
   classes.value = {
     'has-sidebar': hasSidebar.value,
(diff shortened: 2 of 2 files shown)


kiaking:
I don't think it is good idea to have option like this one. Should be configurable by css variable customization 👀

I think you already can by customizing `--vp-nav-bg-color`.

Justin3go:
> I don't think it is good idea to have option like this one. Should be configurable by css variable customization 👀
> 
> I think you already can by customizing `--vp-nav-bg-color`.

Translate to English: Yes, it can certainly be achieved through this method. However, I believe that the function of the transparent navigation bar has already been implemented in the project. Making it configurable and exposing it for user use, rather than having the user re-implement it, would be a better approach in my opinion. On the other hand, if users do not use this configuration, they will also use the default effect (transparent navigation only in layout-home).

Justin3go:
> I don't think it is good idea to have option like this one. Should be configurable by css variable customization 👀
> 
> I think you already can by customizing `--vp-nav-bg-color`.

Yes, it can certainly be achieved through this method. However, I believe that the function of the transparent navigation bar has already been implemented in the project. Making it configurable and exposing it for user use, rather than having the user re-implement it, would be a better approach in my opinion. On the other hand, if users do not use this configuration, they will also use the default effect (transparent navigation only in layout-home).

kiaking:
I don't think we have any other part we do this. For example, how about font weight on nav bar items? Should we have `thickNavBarFontWeight: boolean` option...?

I still think we should have these kind of style related thing in a config.

However, I think we should make navbard bg transparent by default on `page`, when there is no `sidebar` and no `local-nav` 👀 (because in this case, the page layout is same as `home`).

Justin3go:
I think you have a point, but I maintain my opinion. For example, the previous prev/next could not be globally disabled, so I implemented it through display: none, but now it has also become configurable.

Justin3go:
Of course, currently I do achieve the previous effect by changing layout:page => layout:home.

PR #3387, opened 2023-12-29
Fix: Make the outline follow the page scroll

huyikai:
This can be considered a user experience issue. 

When the page is exceptionally long with numerous anchor points, the height of the right outline may exceed the screen height, causing the activated anchor points to become invisible.

 This issue is evident on this page https://vitepress.dev/reference/site-config. 

After adjustments, the activated anchor within the outline will scroll along with the page, without affecting the separate scrolling of the outline

### DIFF
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -183,6 +183,10 @@ export function useActiveAnchor(
       activeLink.classList.add('active')
       marker.value.style.top = activeLink.offsetTop + 33 + 'px'
       marker.value.style.opacity = '1'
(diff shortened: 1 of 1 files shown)


brc-dd (dismissed):

brc-dd:
Doesn't seem to be working on the deploy preview - https://deploy-preview-3387--vitepress-docs.netlify.app/reference/site-config

huyikai:
> Doesn't seem to be working on the deploy preview - https://deploy-preview-3387--vitepress-docs.netlify.app/reference/site-config

emmm Through this link, I see that it is working.  According to the information on "can I use", there should be no problem with compatibility.

PR #3388, opened 2023-12-30
refactor: switch to local custom LRUCache

zonemeen:


zhangyx1998:
Will it help if LRUCache works like a function proxy? Just like python's `@cache` decorator provided by `functools`.

e.g.

```ts
export function cached<K, V>(fn: (...args: K) => V) {
  const cache = LRUCache<K, V>();
  return (...args: K) => {
    if (!cache.has(args)) cache.set(k, fn(...args))
    return cache.get(args)
  }
}
```

PR #3392, opened 2023-12-30
feat(theme): add 'inert' attribute to prevent unnecessary traversal of hidden content

userquin:
Same as the original https://github.com/vuejs/vitepress/pull/2932 but providing global inert controls and state (both exported in default theme).

Also included `focusTrap` in outline (`On this page`) popup.


src/client/app/inert.ts
```diff
@@ -0,0 +1,60 @@
...
+export function useInertState() {
+  return inject<UnwrapNestedRefs<InertState>>(inertStateSymbol)
+}
+
+export function provideInert(app: App) {
+  const inert = reactive({
+    isSidebarOpen: false,
+    isScreenOpen: false,
+    isSidebarEnabled: false,
+    onAfterRouteChanged() {
```
userquin, line 41: use `onAfterRouteChanged` on `nav.ts`

src/client/theme-default/components/VPLocalNavOutlineDropdown.vue
```diff
@@ -1,37 +1,50 @@
...
+  escapeDeactivates: true,
+  delayInitialFocus: false,
+  onDeactivate: () => {
+    open.value = false
+  },
 })
 
+const vh = ref(0)
+
 onKeyStroke('Escape', () => {
```
userquin, line 35: maybe we don't need this

src/client/app/inert.ts
userquin: @brc-dd I'm going to expose directly the logic since we are exposing both composables

userquin: we cannot expose it, it is being used `@click="closeScreen"`

userquin:
Sidebar not woking...

src/client/theme-default/components/VPLocalNavOutlineDropdown.vue
```diff
@@ -155,11 +161,12 @@ function scrollToTop() {
...
 .header {
   background-color: var(--vp-c-bg-soft);
+  padding: 2px 16px;
 }
 
 .top-link {
   display: block;
-  padding: 0 16px;
+  padding: 0;
   line-height: 48px;
```
brc-dd, line 170:
```suggestion
  line-height: 44px;
```

userquin: I just added the div container to allow inert work...

userquin: upps, it is about the outline

brc-dd: Yeah that div thing on skip link is fine.

PR #3405, opened 2024-01-03
fix: avoid history pushState triggering multiple times when click on same internal link

zonemeen:
Clicking on the `a` tag of the same internal link, such as clicking on the same sidebar item or navBar title multiple times, will cause `history.pushState` to trigger multiple times, which normally shouldn't be triggered.

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -329,7 +329,7 @@ function shouldHotReload(payload: PageDataPayload): boolean {
 }
 
 function updateHistory(href: string) {
-  if (inBrowser && href !== normalizeHref(location.href)) {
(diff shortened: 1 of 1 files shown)


src/client/app/router.ts
```diff
@@ -329,7 +329,7 @@ function shouldHotReload(payload: PageDataPayload): boolean {
 }
 
 function updateHistory(href: string) {
-  if (inBrowser && href !== normalizeHref(location.href)) {
+  if (inBrowser && href !== location.href) {
```
brc-dd, line 332:
```suggestion
  if (inBrowser && normalizeHref(href) !== normalizeHref(location.href)) {
```

PR #3468, opened 2024-01-17
docs: GitHub Pages supports clean URLs

skirtles-code:
See also #3020.

The docs claim that GitHub Pages doesn't support clean URLs. That may have been true at some point, but it doesn't seem to be the case now.

I'm using GitHub Pages myself and I haven't noticed any problems using `cleanUrls: true`.


brc-dd:
It works, but there is an issue there - GH pages don't redirect `/foo/` to `/foo` if `/foo/index.html` is not there. But if someone visits that, vitepress changes the route to correct one. So if one is seeing missing content / hydration mismatch and they hit reload it would work fine.

PR #3482, opened 2024-01-20
feat: support GitHub-flavored alerts

### LINKED ISSUE #3278: Support GitHub-style Alert?

GitHub introduced a syntax to have alerts: https://github.com/orgs/community/discussions/16925

Basically, it's:

```
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

I think it would be nice for VitePress to support that built-in. So markdown files authored that way can have a closer experience when reading on GitHub.

I wrote a markdown-it plugin for that: https://github.com/antfu/markdown-it-github-alerts that VitePress might be able to use directly. (and also a short intro and preview on my site: https://antfu.me/posts/github-alerts)

I guess we don't need to align exactly the same as GitHub's styling. We could just render our existing alerts.

_No response_

### Additional context

_No response_

### Validations

antfu:
close #3278

Added three variants:
- `NOTE` (Same as `INFO`)
- `CAUTION` (Same as `DANGER`)
- `IMPORTANT` (with a new purple color).

Screenshot 2024-01-20 at 13 40 07

### DIFF
--- a/src/node/markdown/plugins/githubAlerts.ts
+++ b/src/node/markdown/plugins/githubAlerts.ts
@@ -1,0 +1,63 @@
+import type MarkdownIt from 'markdown-it'
+import type { ContainerOptions } from './containers'
+
+export const gitHubAlertsPlugin = (
+  md: MarkdownIt, 
+  options?: ContainerOptions
+) => {
+  const markers = ['TIP', 'NOTE', 'INFO', 'IMPORTANT', 'WARNING', 'CAUTION', 'DANGER']
+  const matchCaseSensitive = true
+  const titleMark = {
+    tip: options?.tipLabel || 'TIP',
+    note: options?.noteLabel || 'NOTE',
+    info: options?.infoLabel || 'INFO',
+    important: options?.importantLabel || 'IMPORTANT',
+    warning: options?.warningLabel || 'WARNING',
+    caution: options?.cautionLabel || 'CAUTION',
+    danger: options?.dangerLabel || 'DANGER',
+  } as Record<string, string>
+
+  const markerNameRE =  markers.join('|')
+  const RE = new RegExp(`^\\[\\!(${markerNameRE})\\]([^\\n\\r]*)`, matchCaseSensitive ? '' : 'i')
+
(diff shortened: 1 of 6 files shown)


src/node/markdown/plugins/githubAlerts.ts
```diff
@@ -0,0 +1,75 @@
...
+  const markers = [
+    'TIP',
+    'NOTE',
+    'INFO',
+    'IMPORTANT',
+    'WARNING',
+    'CAUTION',
+    'DANGER'
+  ]
+  const matchCaseSensitive = true
```
brc-dd, line 17: Any specific reason for having this true? GitHub's implementation seems to be case insensitive.

antfu: I think it was case-sensitive, was copied from `markdown-it-github-alerts` https://github.com/antfu/markdown-it-github-alerts/commit/34f264da25042a40e50e0475bf1f5c9d392f339e

Let me update this

antfu:
It can be another PR, but do you think we should add icons to those callouts and make it configurable like `infoLabel`?

brc-dd (approved):

brc-dd:
> It can be another PR, but do you think we should add icons to those callouts and make it configurable like infoLabel?

In one PR Guillaume attempted adding it via CSS icons. We can do that probably and instead of making it configurable via config, better to just allow overriding the icon variables in custom CSS?

PR #3488, opened 2024-01-22
chore: upgrade shikiji to v0.10

antfu:
This release improves a lot by the accuracy and TwoSlash integrations: https://github.com/antfu/shikiji/releases/tag/v0.10.0

It should be a non-breaking change for end-users.


brc-dd:
released rc.40

PR #3502, opened 2024-01-26
feat: handle `<img>` tag in the markdown file

yonecdeng:
There is currently an issue with writing `` tags directly in a markdown file, where the tag is not appended with `loading='lazy'`. The added code solves this problem.


brc-dd:
I think we shouldn't be forcing this as someone might knowingly haven't specified loading=lazy or have specified some other value for loading, for example, because it's an above fold image and they want to improve lcp metrics. If someone is using html img tag and want to lazy load an image, they should manually specify loading=lazy attribute.

If they don't want to manually specify and lazy load all their images, it might be better to just make this into a third party plugin, or they can just copy paste this and add it to their config. I don't think many people want this though considering this doesn't have any upvote.

PR #3506, opened 2024-01-26
feat: rename shiki packages

antfu:
A full round circle, and now we are back to Shiki :P

The v1.0.0-beta.0 is the exact same implementation as the latest Shikiji, it shouldn't make any difference to end users.

The only breaking change is that `shikijiSetup` is renamed to `shikiSetup` (it's easy to provide soft deprecation, but since we are in RC that breaking change is still acceptable + I did a quick search on the entire GitHub and it seems no many ppl is using that hook, should be easy to migration as well).

As I don't expect more breaking changes from Shiki, the beta version is used to make sure everything works fine and the migration path is reasonable - It's ok to either have this now, or wait a few days for stable 1.0 release - either way should do the same.

### DIFF
--- a/docs/guide/markdown.md
+++ b/docs/guide/markdown.md
@@ -301,7 +301,7 @@ VitePress also supports [GitHub-flavored alerts](https://docs.github.com/en/get-
 
 ## Syntax Highlighting in Code Blocks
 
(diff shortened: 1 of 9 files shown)


brc-dd:
Let’s cut another rc and wait for few days before releasing stable.

PR #3522, opened 2024-01-31
feat: wrap app title in span to get targeted by CSS

jd-solanki:
Hi 👋🏻 

I want to remove site title only in home page and as there's no direct option I'm using CSS but unfortunetly current DOM structure doesn't allow targeting the text only so I wrapped title in span tag which I good as well.


jd-solanki:
Sorry @brc-dd I was in hurry and forgot the second one 😓

PR #3524, opened 2024-01-31
✨ feat: add hero slots that are inside container

jd-solanki:
Hi 👋🏻 

When trying to add some content before app title & after actions **but inside container** there's no easy way, we have to create custom container class and VitePress' container classes are also scoped which we can't leverage.

Basically, What I'm trying to create is something like this:
image

> _P.S. There's no need for `home-hero-actions-before` because `home-hero-info-after` can do the job_

### DIFF
--- a/docs/guide/extending-default-theme.md
+++ b/docs/guide/extending-default-theme.md
@@ -195,7 +195,9 @@ Full list of slots available in the default theme layout:
   - `aside-ads-after`
 - When `layout: 'home'` is enabled via frontmatter:
   - `home-hero-before`
+  - `home-hero-info-before`
   - `home-hero-info`
+  - `home-hero-actions-after`
   - `home-hero-image`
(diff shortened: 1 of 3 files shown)


jd-solanki:
If possible can you please let me know when you are planning for new release because I'm going to release Vue Cheatsheet based on VitePress today so I can decide if I should have to wait or not.

brc-dd (approved):

brc-dd:
> you are planning for new release

Most likely by this weekend. I'm waiting on some other libraries to release patches. Hopefully that will be the last/second to last RC.

pirog:
@jd-solanki @brc-dd wouldn't it make sense to also bubble these slots up to at least VPHomeHero.vue since that is the component that is actually exported with the theme?

brc-dd:
Yeah right. I'll update that.

PR #3537, opened 2024-02-02
perf(theme): move svg icons to css

### LINKED ISSUE #3399: Switch to CSS icons or SVG sprites
Currently there are lot of repeated icons in generated dist/**/*.html files. Those can be avoided by these.

userquin:
I need to move also local search svg icons and  `src/client/theme-default/support/socialIcons.ts`, we cannot use title inside the svg.

CSS increase ~2KB (brotli).

There is ~5KB less per page (~2KB per page with brotli), tested with VP docs and the markdown page (https://vitepress.dev/guide/markdown):
- current page 24257B (brotli)
- on my local 19574B (brotli)

VP docs only using Github icon, sites with more social icons will reduce the size more.

Right now the icons cannot be changed, if we want to allow change the icons this PR requires some changes.

closes #3399


userquin:
This PR will not remove unused icons and will have same problem described here https://github.com/vuejs/vitepress/pull/3539#issuecomment-1925901068 (svg sprites).

PR #3564, opened 2024-02-10
docs: pnpx instead of pnpm for the setup wizard

onmax:


brc-dd:
1. pnpx is deprecated.
2. pnpm vitepress will behave the same as pnpx vitepress if vitepress is installed. And if vitepress is not installed, you’ll need to install it manually.

PR #3588, opened 2024-02-21
fix: import `treatAsHtml` path error

btea:


brc-dd:
`shared/shared.ts` is copied to `node/shared.ts` and `client/shared.ts` during build / dev. You can also run `pnpm dev:shared` or `pnpm build:prepare`

btea:
Oh, I see.

PR #3592, opened 2024-02-22
feat: config file hot update error prints error message

btea:
When modifying the configuration file causes an error, the error message currently output is `failed to load config from xxx.config.ts`. 

I think the information is not clear enough. Printing out the error stack information can help users better locate the problem.


btea:
@brc-dd Do you think this modification is necessary?

src/node/plugin.ts
```diff
@@ -384,6 +384,7 @@ export async function createVitePressPlugin(
         try {
           await resolveUserConfig(siteConfig.root, 'serve', 'development')
         } catch (err: any) {
+          console.error(err)
```
brc-dd, line 387: can you refactor it to use `siteConfig.logger.error` instead

btea: Ok.

PR #3605, opened 2024-02-25
docs: fix wrong command

jsparkdev:
To run the wizard, you must run the `pnpm dlx` command, not `pnpm`.


brc-dd:
It won’t work. You’ll need to install vitepress manually and then run the init command as recommended in the docs. Before opening PRs please do a search and avoid creating duplicates.

PR #3169, opened 2023-11-02
docs: fix Github Pages Action example error

maicss:
1, Set default node version to 22
2, Update `pnpm` block, previous not working
3, Default projet root is current dir not `docs`, change `build` and `project path` blocks. Previous config throw error `No such file or directory`

PR #3626, opened 2024-03-02
docs: set pnpm's version in deploy.yml

xxhls:


AyushSehrawat:
Better changes done in https://github.com/vuejs/vitepress/pull/3169 pull request. This pr also aims to solve the same issue as that one

PR #3633, opened 2024-03-04
fix(theme): judgment conditions error with `markdownStyles`

kovsu:
When I do not declare markdownStyles in frontmatter but the `<Content />` still be wrapped by `<VPHomeContent />`. So there is a problem with the judgment conditions here.


brc-dd:
It's enabled by default now. You need to manually disable it you want the older behavior. Refer the note here - https://vitepress.dev/reference/default-theme-home-page#markdown-content

PR #3634, opened 2024-03-06
feat: add `window.__VITEPRESS__`

LoTwT:
## description

This PR is to add `window.__VITEPRESS__`. 

It helps to detect if the current site is powered by Vitepress, referring to [here](https://github.com/vuejs/devtools-next/issues/202#issuecomment-1977781974).


src/client/app/index.ts
```diff
@@ -64,6 +65,9 @@ const VitePressApp = defineComponent({
 })
 
 export async function createApp() {
+  const target = getGlobalThis()
+  target.__VITEPRESS__ = true
```
brc-dd, line 69: Try simply with `globalThis.__VITEPRESS__ = true` once. I don't think we need to polyfill it as we don't target very old browser.

Issue #2881, opened 2023-08-30
an option to auto-collapse sidebar + scroll item into view


brc-dd:
refer #2879

PR #3654, opened 2024-03-13
feat: scroll active sidebar link into view on page load

### LINKED ISSUE #3351: Keep aside marker visible

Hi

https://github.com/vuejs/vitepress/assets/44144724/4eb6c290-e381-48e8-9cd3-3ab0173b347d

When aside is tall, when scrolled down, the marker goes out of view.

I don't know. Just keep it visible.

_No response_

### Additional context

_No response_

### Validations

BobbieGoede:
* Related #2881
* Resolves #3351 

I find myself searching in the sidebar in the VueUse docs for related composables, the current page link is often out of view in the sidebar. 

Not too happy with my solution in `VPSidebar.vue`, I guess this won't scroll the relevant link into view when navigating using in page links, also not sure if `querySelector` is the appropriate way to find the relevant link. Should a router hook be used instead?

I can add tests to confirm the changes work as expected, let me know if changes should be made!

### DIFF
--- a/src/client/theme-default/components/VPSidebar.vue
+++ b/src/client/theme-default/components/VPSidebar.vue
@@ -1,7 +1,7 @@
 <script lang="ts" setup>
 import { useScrollLock } from '@vueuse/core'
 import { inBrowser } from 'vitepress'
-import { ref, watch } from 'vue'
+import { nextTick, onMounted, ref, watch } from 'vue'
 import { useSidebar } from '../composables/sidebar'
 import VPSidebarGroup from './VPSidebarGroup.vue'
 
@@ -35,6 +35,11 @@ watch(
   },
   { deep: true }
 )
+
+// scroll active link into view
+onMounted(() => {
+  nextTick(() => navEl.value?.querySelector('.is-active')?.scrollIntoView({ block: 'center'}))
+})
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


src/client/theme-default/components/VPSidebar.vue
```diff
@@ -25,6 +25,11 @@ watch(
   },
   { immediate: true, flush: 'post' }
 )
+
+// scroll active link into view
+onMounted(() => {
+  nextTick(() => navEl.value?.querySelector('.is-active')?.scrollIntoView({ block: 'center'}))
```
brenoepics, line 31: And if the same route or pattern is placed 2+ times in the sidebar, is it working fine? also I'd calculate to scroll only when it is out of the viewport

src/client/theme-default/composables/outline.ts
```diff
@@ -197,6 +197,7 @@ export function useActiveAnchor(
 
     if (activeLink) {
       activeLink.classList.add('active')
+      activeLink.scrollIntoView({ behavior: 'smooth', block: 'center' })
```
brenoepics, line 200: PR preview
image

Fixed
image

we also need to fix some edge cases like scroll to 0 when there is no active link, I'd also scroll only when it is out of viewport

brenoepics (changes requested):

Issue #3660, opened 2024-03-16
Feature: export `<VPSwitch />` for better custom theme


peterroe:
### Is your feature request related to a problem? Please describe.

I noticed a closed pr: https://github.com/vuejs/vitepress/pull/2347/files

If someone want to implement it by himself, it can be helpful to use the exported component (VPSwitch)

### Describe the solution you'd like

```ts
import { VPSwitch } from 'vitepress/theme'
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
To implement that particular PR one doesn't need the VPSwitch component. Refer - https://vitepress.dev/guide/extending-default-theme#on-appearance-toggle

peterroe:
I missed the documentation on this part, thanks for the reminder 🙏

PR #3664, opened 2024-03-17
docs:  add `ts-expect-error` flag for Transition API

peterroe:
https://github.com/vuejs/vitepress/issues/3660


brc-dd:
Nope, you need to install `@types/dom-view-transitions` and reference it.

PR #3741, opened 2024-04-03
fix: using file inclusion, if included file contains relative path, t…

windsonR:
When using markdown file inclusion

The included markdown file(`a.md`) wll directly copy file content to replace the comment in a file(`b.md`)

But if the included markdown file contains relative path (such as ``), when access `b.html`, it will report error(`./a.png can not resolve by vite`)

reproduce repo

In reproduce repo, i've add a patch to fix this.

This PR replace the relative path to absolute path to fix this problem.


brc-dd:
This will be a breaking change and not sure if we want that. Including markdown should behave similar to importing components. And if in a vue component you have a path relative to that file, that won't be transformed to something else.

windsonR:
It's not like vue component for now. The `file inclusion` is just copy the markdown file's content to replace the `@include` comment. 

Maybe should re-implement `File inclusion`? 

To make sure that it is not simply copying the content and replacing the file include comment with it, but injecting the compiled markdown file to the comment?

li-zhixin:
I'm having the same problem, is there any recommended solution?

PR #3764, opened 2024-04-07
docs(getting-started.md):Add a space

conglinyizhi:
This is a space prompted by the instructions when the bun is runtime, and after the system installs the bun, there is no instruction called bunx but needs 'bun x' to execute normally, so a space is added.

### DIFF
--- a/docs/zh/guide/getting-started.md
+++ b/docs/zh/guide/getting-started.md
@@ -72,7 +72,7 @@ $ pnpm vitepress init
 ```
 
 ```sh [bun]
-$ bunx vitepress init
+$ bun x vitepress init
 ```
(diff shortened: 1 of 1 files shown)


brc-dd:
You probably have a broken installation of bun.

conglinyizhi:
Okay, it looks like the way I installed the bun wasn't very formal, maybe something weird went wrong with the installation, I realized that there might be a bug on my side, so I'll close this PR

conglinyizhi:
For more information for latecomers, I installed the bun globally via pnpm, which doesn't seem to be on the official installation instructions, as I'm used to using pnpm instead of npm.

Issue #3756, opened 2024-04-05
Broken sidebar when loading page


florian-sattler:
### Describe the bug

First of all: Thank you all very much for your work on this amazing software! 👍 

When navigating to page using a direkt link the sidebar overlap the content of the page:

![image](https://github.com/vuejs/vitepress/assets/65666934/726ef0a4-950f-4f87-bb57-f5b304b07416)

However when accessing the page from the main page, everything looks normal:

![image](https://github.com/vuejs/vitepress/assets/65666934/d9fce76e-ed50-4f93-ba97-a9d2440c6df6)

The problem seems to be caused by a missing class on two elements. Take a look at the developer console:

Broken:
![image](https://github.com/vuejs/vitepress/assets/65666934/02333a4d-2617-4911-ac00-a138d3d102c5)

Working:
![image](https://github.com/vuejs/vitepress/assets/65666934/70f7ad5a-2f94-4705-aaaa-ffd41cac2794)

Notice how the class `.has-sidebar` is missing on the first (`.VPLocalNav`) and third element (`#VPContent`) below the header (`.VPNav`)

When manually adding this class in the browers dev tools the site looks normal again 😄 

### Reproduction

See above.

### Expected behavior

The sidebar should be located on the side not above the page content.

### System Info

```Text
System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
    Memory: 12.23 GB / 15.61 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
  Browsers:
    Chrome: 123.0.6312.105
  npmPackages:
    vitepress: ^1.0.2 => 1.0.2
```

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

brc-dd:
[...]

PR #3770, opened 2024-04-09
docs: added nginx deployment documentation

florian-sattler:
Added nginx deployment documentation as discussed in #3756 with some links for further reading.

### DIFF
--- a/docs/guide/deploy.md
+++ b/docs/guide/deploy.md
@@ -291,3 +291,47 @@ You can deploy your Vitepress website on [Kinsta](https://kinsta.com/static-site
 ### Stormkit
 
 You can deploy your VitePress project to [Stormkit](https://www.stormkit.io) by following these [instructions](https://stormkit.io/blog/how-to-deploy-vitepress).
+
+### Nginx
+
+Here is a example of an Nginx server block configuration. This setup includes gzip compression for common text-based assets, rules for serving your VitePress site's static files with proper caching headers as well as handling `cleanUrls: true`.
+
+```nginx
+server {
+    gzip on;
+    gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
+
+    listen 80;
+    server_name _;
+    index index.html;
+
+    location / {
+        # content location
+        root /app;
+
+        # exact matches -> reverse clean urls -> folders -> not found
+        try_files $uri $uri.html $uri/ =404;
+
+        # non existent pages
+        error_page 404 /404.html;
+
+        # a folder without index.html raises 403 in this setup
+        error_page 403 /404.html;
+
+        # adjust caching headers
+        # files in the assets folder have hashes filenames
+        location ~* ^/assets/ {
+            expires 1y;
+            add_header Cache-Control "public, immutable";
+        }
+    }
+}
+```
+
+This configuration assumes that your built VitePress site is located in the `/app` directory on your server. Adjust the `root` directive accordingly if your site's files are located elsewhere.
+
+::: warning Do not default to index.html
+The try_files resolution must not default to index.html like in other Vue applications. This would result in an invalid page state.
+:::
+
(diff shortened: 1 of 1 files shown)


kiaking (approved):

brc-dd:
This doesn't handle the case of removing trailing slash. `/foo/bar/` should also check for `/foo/bar.html`

florian-sattler:
@brc-dd Yes indeed. Removing the trailing slash by redirecting also has downsides: https://blog.mehdi.cc/articles/vitepress-cleanurls-on-nginx-environment#the-trade-off-for-index-pages

Using the currently described setup:
- When accessing a folder which has an index.html file this will get delivered whether or not using a trailing "/"
- When trying to access a document (e.g. example.html) this will break if being accessed using a trailing "/" (the page gets shown however filled from the error page which results in a broken sidebar as discussed in #3756)

I'm currently on vacation, so I couldn't test the behavior when there is a folder (example/) along a file with the same name (example.html). This might indeed break.

Did I miss a (easy) way to handle this?

brc-dd:
I don't think that blog post is doing things correctly. They are doing 301 redirects while vitepress explicitly states you need rewrites for cleanUrls to work (rewrites here are different than nginx rewrite directive). Rewrite basically means serving the page without doing 301/302. I'll try and update the configuration.

meduzen:
Hi! Author of the [article](https://blog.mehdi.cc/articles/vitepress-cleanurls-on-nginx-environment) here.

## What I do

> I don't think that blog post is doing things correctly. They are doing 301 redirects while vitepress explicitly states you need rewrites for cleanUrls to work (rewrites here are different than nginx rewrite directive)

Actually I do both. My goal is that the user is always on the canonical (no trailing `/` nor `.html`), so:
- I **redirect** to make sure the clean URL is actually clean.
- Once redirected, I do **rewrite** to make sure the right `path/to/file.html` gets served.

For example, to serve the file at `https://blog.mehdi.cc/articles.html`:

- https://blog.mehdi.cc/articles is the canonical (default behaviour of Vitepress [`cleanUrls`](https://vitepress.dev/reference/site-config#cleanurls))
- the following URLs get redirected to the canonical:
  - https://blog.mehdi.cc/articles.html
  - https://blog.mehdi.cc/articles/
  - https://blog.mehdi.cc/articles/index.html (2 redirects: goes first to https://blog.mehdi.cc/articles/)

(And I also want all URLs to work without JavaScript in the browser. But in a way or another, static site generators have their limits, and looking at what [hosting service do](https://github.com/slorber/trailing-slash-guide?tab=readme-ov-file#summary), there are redirects and URL inconsistencies (user point of view) in most _clean URLs_ scenarios. So choices have to be made and URL consistency is very important for me.)

### Alternatively

Since I provide a `<meta name="canonical">` I could just skip the redirects, but it seems not easy with Nginx to cover all use cases, like for example the one you describe:

> This doesn't handle the case of removing trailing slash. /foo/bar/ should also check for /foo/bar.html when cleanUrls is enabled.

I can’t have this without first getting rid of the trailing slash with a redirect. And then it creates more issues if. More on this in next section.

## Testing domain and branch

To help with testing purpose, I opened a branch and a new subdomain:
- the only [change on the branch](https://github.com/meduzen/blog/compare/cf4cf86e01a54181ed9415313e3a57aea6f70a0f...test/nginx-cleanurls) is to moves `/notes.md` to `/notes/index.md` (while `/articles.md` remains at the same place).
- the subdomain https://test.blog.mehdi.cc applies the nginx config described in Vitepress doc (so the one from this PR, not the one from my article).

When doing this:
- All variations of https://test.blog.mehdi.cc/notes works (`/notes`, `/notes/`, `/notes.html`, `/notes/index.html`) to serve the file is at https://test.blog.mehdi.cc/notes/index.html, but `/notes.html` and `/notes/index.html` only works if JavaScript is available client-side
- https://test.blog.mehdi.cc/articles works but https://test.blog.mehdi.cc/articles/ fails (the wanted file is at https://test.blog.mehdi.cc/articles.html)

I also tested with `rewrite ^(.+)/+$ $1;` (remove trailing slash) to enable the use case quoted in the previous section (`/foo/bar/`): it works fine in all scenarios, but:
- forward slash fails: for `/notes` and `/notes/` both redirect endlessly to `/notes/`;
- https://test.blog.mehdi.cc/articles/index.html works (redirects to `https://test.blog.mehdi.cc/articles` client-side) but fails without JS

(This is not the config live on the `test.` subdomain but I can enable it if you want to have a look.)

Let me know if you would like to test specific things.

PR #3772, opened 2024-04-10
feat: upgrade markdown-it types to v14

meteorlxy:
Upgrade `@types/markdown-it` and `@mdit-vue/*`


patches/@types__markdown-it@14.0.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/package.json b/package.json
+index 46d7bfb79a36d8090d98d1ea1e36216137e00072..7bc1e8be5f1c5b74f52f7fa08935d2df3d19d3e2 100644
+--- a/package.json
++++ b/package.json
+@@ -27,7 +27,7 @@
+         }
+     ],
+     "main": "",
+-    "types": "index.d.ts",
++    "types": "index.d.mts",
```
meteorlxy, line 10: Without this patch, tsc could pass but rollup-plugin-dts would throw an error.

This patch is to make rollup-plugin-dts work. May need more investigation for the root cause.

package.json
```diff
@@ -126,13 +126,13 @@
...
   "devDependencies": {
     "@clack/prompts": "^0.7.0",
-    "@mdit-vue/plugin-component": "2.0.0",
-    "@mdit-vue/plugin-frontmatter": "2.0.0",
-    "@mdit-vue/plugin-headers": "2.0.0",
-    "@mdit-vue/plugin-sfc": "2.0.0",
-    "@mdit-vue/plugin-title": "2.0.0",
-    "@mdit-vue/plugin-toc": "2.0.0",
-    "@mdit-vue/shared": "2.0.0",
+    "@mdit-vue/plugin-component": "2.1.0",
```
brc-dd, line 129: feel free to use semver ranges, `^2.1.0` . I had locked because it was causing some issues 👀

src/node/markdown/markdown.ts
```diff
@@ -238,7 +240,7 @@ export const createMarkdownRenderer = async (
     slugify,
     permalink: anchorPlugin.permalink.linkInsideHeader({
       symbol: '&ZeroWidthSpace;',
-      renderAttrs: (slug, state) => {
+      renderAttrs: (slug, state: StateCore) => {
```
brc-dd, line 243: if we are going down the patch route, maybe patch this as well

node_modules/markdown-it-anchor/types/index.d.ts

```ts
import MarkdownIt from 'markdown-it';
import Token from 'markdown-it/lib/token.mjs'
import State from 'markdown-it/lib/rules_core/state_core.mjs';
```

this should fix those unresolved imports in dts build too

package.json
meteorlxy: Updated

src/node/markdown/markdown.ts
meteorlxy: Patched `markdown-it-anchor` and reverted the change here

brc-dd (approved):

patches/@types__markdown-it@14.0.0.patch
brc-dd: > Need more investigation for the root cause.

`resolvedCompilerOptions` seem to be always empty here - https://github.com/Swatinem/rollup-plugin-dts/blob/3835bfa27125db7fa5a9d071f0b93ec513d2664b/src/index.ts#L196

If we force moduleResolution bundler there then it resolves to correct files, but then the transform breaks polka by converting `export default function<T` to `declare functionexport_default<T`

PR #3847, opened 2024-04-30
feat: support icon by svg tag

### LINKED ISSUE #3828: add an API to support SVG's dynamic logo

now the vitepress logo use img element , so when I use dynamic svg as logo it not  have any dynamic effects

the logo can use dynamic svg 

_No response_

### Additional context

_No response_

### Validations

gweesin:
VPImage support rendering SVG icon, not wrapped in an  tag.

closes #3828


bluwy:
I don't think we should be fetching the SVG and rendering it this way. Ideally it should be inlined. Either automatically or via an option, it should probably be discussed in the issue first.

PR #3877, opened 2024-05-10
feat: open code block as modal

ToBinio:
This PR makes it possible to add `:modal` to a code block.
This adds a button to the code block which makes it possible to open it in a "fullscreen"
 modal.
This is useful for cases where the code is very wide and you would need to constantly scroll left and right to see the whole content.
example in the [wip fabric wiki](https://docs.fabricmc.net/develop/items/potions#registering-the-potion)

image


brc-dd:
Just a heads up though, `<dialog>` doesn't match our baseline. So it's unlikely this will be accepted any time soon.

ToBinio:
I think it should still be easily doable without `<dialog>` it would just make the code a bit simpler

ToBinio:
@brc-dd 
Hi, i just recently saw this PR again and wanted to quickly ask if I should resolve the merge conflicts, or won't it be merged anyway?

its-miroma:
fyi: This is very similar to `FullscreenCode`, courtesy of @Jimmy474. `FullscreenCode` includes a few more features: toggle wrap code, copy code, support for code groups, etc.

https://github.com/user-attachments/assets/1caaaaa2-f49d-4b81-86ad-69cad8025b06

https://github.com/FabricMC/fabric-docs/blob/main/.vitepress/theme/components/FullscreenCode.vue

The feature is live on the Fabric Docs! (they've grown so much since @ToBinio's screenshot :face_holding_back_tears:)

https://docs.fabricmc.net/develop/blocks/container-menus#creating-the-menu

The way it works is that a `<dialog>` is always mounted in `layout-bottom`, and when the button is clicked the code block it is in is cloned into the `dialog#fullscreen`.

It's probably too big to be upstreamed to VitePress, and the code does not follow this repo's baseline and style requirements. Regardless, I thought I could share.

Also, one interesting thing that @brc-dd might want to look at is the way we did the line numbers margin: it is pure CSS, meaning it has less impact on the DOM, and it works when wrapping lines. Vitepress currently has a `div.line-numbers-wrapper` on the left.

Jimmy474:
as this also supports code groups tabs appear in fullscreen mode too.

PR #3918, opened 2024-05-26
From `pnpm vitepress init` to `pnpx vitepress init`

berzanorg:
The command `pnpm vitepress init` doesn't work as `vitepress` isn't a subcommand of pnpm.
But the command `pnpx vitepress init` works as `pnpx` is the equivalent of `npx` for pnpm.

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -68,7 +68,7 @@ $ npx vitepress init
 ```
 
 ```sh [pnpm]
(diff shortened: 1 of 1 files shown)


brc-dd:
It's not supported currently. First adding deps needs to be implemented in the init part. You currently need to manually install vitepress and run that command. Also, `pnpx` is deprecated in favor of `pnpm dlx`.

PR #3953, opened 2024-06-11
Update default-theme-sidebar.md

abdul-alhasany:
Update folder tree. The current tree confuses reader because it seems that config is a subfolder while it is on the same level as guide.


brc-dd:
it is consistent with what `tree` gives as output

Issue #3954, opened 2024-06-11
[New feature] export `VPHomeContent` to provide content-adaptive layout


peterroe:
### Is your feature request related to a problem? Please describe.

Conveniently use content-adaptive layout in components on the home page

### Describe the solution you'd like

`.vitepress/theme/index.ts`:
```ts
import HomePage from './component/HomePage'

export default {
  Layout: () => {
    return h(Theme.Layout, null, {
      'home-features-after': () => h(HomePage),
    })
  },
```

---

`HomePage.vue`:

```html
<script setup lang="ts">
import{ VPHomeContent } from 'vitepress/theme'
</script>

<template>
	<VPHomeContent>
		<!-- my custom content -->
	</VPHomeContent>
</template>
```

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

PR #3960, opened 2024-06-14
feat(theme): export VPHomeContent

xiaodong2008:
Add #3954


src/client/theme-default/index.ts
```diff
@@ -1,4 +1,8 @@
 import './styles/fonts.css'
 
+import VPHomeContent from './components/VPHomeContent.vue'
+
 export * from './without-fonts'
 export { default as default } from './without-fonts'
+
+export { VPHomeContent }
```
brc-dd, line 8: export it directly in from without-fonts file, see how others are exported there

brc-dd (approved):

PR #3981, opened 2024-06-23
refactor: replace `npm-run-all` with `npm-run-all2`

sxzz:
A maintenance fork of `npm-run-all`, with less deps.

Already replaced in vue-core https://github.com/vuejs/core/pull/10228

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "micromatch": "^4.0.7",
     "minimist": "^1.2.8",
     "nanoid": "^5.0.7",
-    "npm-run-all": "^4.1.5",
+    "npm-run-all2": "^6.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -166,7 +166,7 @@
     "micromatch": "^4.0.7",
     "minimist": "^1.2.8",
     "nanoid": "^5.0.7",
-    "npm-run-all": "^4.1.5",
```
sheremet-va, line 169: Just a note, but pnpm supports this out of the box. We are using this in Vitest for example to run both several scripts at the same time:

```ts
pnpm run --stream '/^dev:/'
```

brc-dd: Thanks, updated!

PR #3996, opened 2024-06-28
Separate and rewrites directory for en

F-loat:


brc-dd:
move docs/components to docs/en/components

brc-dd:
ah wait, I'll update those

PR #4000, opened 2024-06-29
feat: allow using components in navigation bar

### LINKED ISSUE #3998: Allow use components inside navbar

Add a new option to `DefaultTheme.NavItem` to allow use a component.

I'll try to send a PR tmr.

/cc @brc-dd 

Replace the content via `component::is`

None

### Additional context

Check https://github.com/unocss/unocss/pull/3931

### Validations

userquin:
✔️  I  need to add an example/test...

closes #3998


userquin:
we should also add fly, maybe includng `mode` type: `type Mode = 'screen' | 'fly' | undefined`  or `type Mode = 'screen' | 'fly' | 'navbar'`

userquin:
@brc-dd same component in multiples places, no makes sense, do we need the new mode prop?  maybe not, since we should always have screen in screen menu (we'll need to add screen prop when custom component in root)

fly-screen-navbar

fly-screen-navbar-2

brc-dd:
Do we need some option to move component to flyout in medium viewports? Inside this thing:

image

userquin:
I guess there is no need, `fly` will be used when the custom component is not a root child in the navbar entry. If the custom component is in the root navbar, VP will not use `fly` (will use media query bp to collapse or expand the full navbar)

brc-dd (approved):

PR #4014, opened 2024-07-03
fix: normalize path with suffix slash

F-loat:
normalize `guide/routing/` -> `guide/routing`, with the following directory structure:

```
.
├─ routing
│  └─ index.md
└─ index.md
```

### DIFF
--- a/src/shared/shared.ts
+++ b/src/shared/shared.ts
@@ -20,6 +20,7 @@ export const APPEARANCE_KEY = 'vitepress-theme-appearance'
 const HASH_RE = /#.*$/
 const HASH_OR_QUERY_RE = /[?#].*$/
 const INDEX_OR_EXT_RE = /(?:(^|\/)index)?\.(?:md|html)$/
+const SUFFIX_SLASH_RE = /\/$/
 
 export const inBrowser = typeof document !== 'undefined'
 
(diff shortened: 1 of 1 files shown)


brc-dd:
It's not the expected behavior as relative urls will break. The slash should be kept there. If you want to remove the trailing slash you'll need to use rewrites.

F-loat:
Currently, if I rename `routing.md` to `routing/index.md` while keeping the previous navigation configurations intact, everything works fine except for the prev/next components. Is it necessary to modify the prev/next components to accommodate this scenario?

``` json
{
  "nav": [
     {
        "text": "Routing",
        "link": "/guide/routing"
     }
  ]
}
```

* routing.md

routing.md

* routing/index.md

routing/index.md

brc-dd:
It's strict in that regard. You need trailing slash in sidebar if your url has trailing slash. If you don't want trailing slash in your URL you can add some config like this:

```ts

export default defineConfig({
  rewrites: {
    '^(?<rest>.*(?=\\/))?\\/?(?<dir>.+?)\\/index\\.md$': ':rest*/:dir.md'
  }
})
```

F-loat:
Yes, you are right. In this case, it makes sense to either use rewrites or modify the nav config.

PR #4028, opened 2024-07-08
fix: bun scripts start w bunx

rithulkamesh:
The getting started document uses `bun`, an npx-like functionality can be achieved by using bunx


brc-dd:
Read that guide. You need to first install vitepress. The init command won't install vitepress. While `bunx vitepress` would work, in the end you'll need to manually install vitepress anyway. So better install it first then run `bun vitepress`. For same reason we haven't listed commands like `pnpm dlx` which download and execute scripts without installing.

PR #4041, opened 2024-07-14
feat: improve printing experience.

7086cmd:
In some cases, we may need to print documents as PDFs, using [mdbook](https://github.com/rust-lang/mdBook) as an example. While certain components can be omitted during printing, there are situations where we might want to retain them. To address this, I am introducing an option to manage these components.

In the current implementation, I use the `important` key to emphasize hiding certain elements, which may lead to unintended side effects. I plan to remove these commands to avoid such issues.

#### Before
Before Image

#### After
After Image

With the `removeExtraElementsInPrint` option in `DefaultTheme.Config`, users can decide whether to keep or omit specific components during printing. Additionally, I have modified the dark mode switch function to automatically switch to light mode when the `beforeprint` event is triggered.

---

Let me know if any further changes are needed!


7086cmd:
So strange. Local code are formatted, but online repo reported errors.

brc-dd:
You seem to be using different prettier version. Try re-running pnpm install. Then pnpm format.

7086cmd:
@brc-dd It works. Thanks!

PR #4046, opened 2024-07-15
fix: change the setup wizard command for `pnpm`

amir78729:
The `pnpm vitepress init` command will throw this error

```
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "vitepress" not found
```

The alternative of `npx` in pnpm is `pnpm dlx` (`pnpx`)


brc-dd:
init doesn't add vitepress to your deps. You'll have to manually install vitepress either way. Read the section before that, it mentions that you first need to install.

PR #4062, opened 2024-07-18
ci: setup pkg.pr.new

brc-dd:
WIP

Some credits to https://github.com/radashi-org/radashi and https://github.com/antfu-collective/.github/blob/main/.github/PULL_REQUEST_TEMPLATE.md 😅

TODO:
- [x] adjust permissions
- [x] replace echo publish with actual command
- [ ] ignore changes in docs and markdown files (maybe later)
- [ ] use --compact with pkg-pr-new when fix is released upstream

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+### Description
+
+<!-- Please insert your description here and provide especially info about the "what" this PR is solving. -->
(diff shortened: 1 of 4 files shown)


Aslemammad:
Thank you so much y'all for the opportunity, it means a lot!

PR #4082, opened 2024-07-28
fix(markdown/table): use `Token.attr[Get|Join]` insteed the hardcode `tabindex` attr

TheNorthMemory:
`markdown-it-attrs` supported extra attributes for eg: `{tabindex=1}` after `table` block, fix it.

### Description

https://github.com/vuejs/vitepress/blob/8f31a4c0780657af9137762c9394c56912f6e383/src/node/markdown/markdown.ts#L235

### Additional Context

refer: https://markdown-it.github.io/markdown-it/#Token.attrGet and https://markdown-it.github.io/markdown-it/#Token.attrJoin


TheNorthMemory:
Hi there, this site https://wechatpay.im, is strictly rely on this function, because there were more than 1000+ markdown tables rended. Hopes this fix shall be merged.

image

PR #4125, opened 2024-08-11
Put all styles in `@layer` for lowered specificity

felix-bohlin:
**Problem I'm trying to solve:** 
I want to use Vitepress to document my own component library. The issue I have is that Vitepress' own styling sometimes override my own components. By putting all of Vitepress' own styles inside `@layer` we lower the specificity of those styles and make sure they are easily overridden.


brc-dd:
It doesn't meet our current baseline (Chrome/Edge >= 88, Firefox >= 78, Safari >= 14).

felix-bohlin:
@brc-dd Ah, too bad. Thanks for your reply! :)

PR #4154, opened 2024-08-21
chore: use pnpx init project

btea:
### Additional Context

---

> [!TIP]

### DIFF
--- a/docs/en/guide/getting-started.md
+++ b/docs/en/guide/getting-started.md
@@ -72,7 +72,7 @@ $ npx vitepress init
 ```
 
 ```sh [pnpm]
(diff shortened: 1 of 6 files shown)


brc-dd:
You need to install vitepress first. Even if you use pnpx, you'll need to manually install vitepress. The init script doesn't add it in deps for you. The prerequisite part tells to install vitepress first.

btea:
According to the documentation, using `pnpx` does not require any dependencies to be installed locally. Am I misunderstanding this description?

image

brc-dd:
vitepress needed you to install vitepress first

btea:
I see, thanks for the quick response.

PR #4196, opened 2024-09-12
docs: Update deploy.md

HoronLee:
### Description

An output directory of `docs/.vitepress/dist` will cause vercel to be unable to find the dist directory, so you need to change the output directory to `.vitepress/dist`.

### Linked Issues

I didn't create an issue

### Additional Context

I tried to deploy vitepress on vercel, but when filling in the project output path, I found that the path in the docs didn't exist, so I deleted `docs/` before `.vitepress/dist`


brc-dd:
Ah actually that part of documentation assumes that you’ve `docs` as vitepress root. But if you’ve run vitepress init with default options you probably have content outside docs directory.

PR #4202, opened 2024-09-15
Fix md pre style

fanhefeng:
### Additional Context

---

> [!TIP]


brc-dd:
It isn't needed. Things are kept simple in examples. Also, it's not recommended to directly use `language-`. Better to use something like this:

````md
```json-vue
{{ theme }}
```
````

PR #4266, opened 2024-10-10
Update VPSidebar.vue

erwinheldy:
Reduce excessive padding-bottom which causes scrollbars to appear

### Description

Scrollbar appears in FHD resolution with bottom panel

### Comparison

Before:
before

After:
after


brc-dd:
It's intentional from UX perspective. One shouldn't have to look at the very bottom or have to move their mouse to bottom to click something. It's a common practice, for example, vs and vscode, zed, notepad++, all implement this. You can override this from CSS if you don't want it.

PR #4283, opened 2024-10-14
fix(deadlinks): Check if public file exists

veryard:
### Description

Currently if you have a file inside the `public/` folder like such as `public/files/example.xlsx` it will be flagged a dead link. So this just updates the `markdownToVue` file to check if it's a public file and it exists so it doesn't flag as dead link. 
Ie.
```md
[here](/files/example.xlsx)
```
The above would result in a dead link.

This was causing some build issues with a team of 10 people writing docs. Means they don't have to specify the full path to the file.

Happy to hear any feedback!

Thanks.

### DIFF
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -176,9 +176,15 @@ export async function createMarkdownToVueRenderFn(
         )
         resolved =
           siteConfig?.rewrites.inv[resolved + '.md']?.slice(0, -3) || resolved
+
+        const htmlFileExists = fs.existsSync(path.resolve(dir, publicDir, `${resolved}.html`))
+        const checkIfPublicFileExists = fs.existsSync(path.resolve(dir, publicDir, resolved))
+
         if (
           !pages.includes(resolved) &&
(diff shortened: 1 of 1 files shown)


brc-dd:
add this to top of your config file

```ts
process.env.VITE_EXTRA_EXTENSIONS = ['xlsx'].join(',') // add any other extensions you want
```

veryard:
> add this to top of your config file
> 
> ```ts
> process.env.VITE_EXTRA_EXTENSIONS = ['xlsx'].join(',') // add any other extensions you want
> ```

Yeah I understand, just thought it would be a bit more user friendly for non technical people writing docs to just have it work verses having them go and amend this every time a file type is added.

brc-dd:
I don't think checking if the file exists in public directory is the right way to deal with this. Non-html files in public dir should simply be ignored (and I think html files in public directory should also be ignored, not sure why we aren't doing that already).

People have different workflows and even have different apps combined with vitepress. Many people are manually copying this stuff or keeping static files in different app's public directory. It works in runtime because all of those files end up in the same place. It will be a breaking change for them as they will need to custom logic to ignoreDeadLinks check.

Issue #4300, opened 2024-10-21
Unescaped code blocks syntax highlight doesn't work


salazarr-js:
### Describe the bug

When using [Unescaped code blocks](https://vitepress.dev/guide/using-vue#unescape-in-code-blocks) with the content coming from a string variable or a [data loader](https://vitepress.dev/guide/data-loading#data-from-local-files), the code blocks aren't highlighted 

````html
```ts-vue
{{ scriptString }}
```

```ts-vue
{{ scriptData }}
```

<script setup lang="ts">
import { data } from './examples.data'

const scriptData = data

const scriptString = `
/** Generate a greet message with a provided text */
function greet(text) {
  return \`Hello, \${text}\`;
}

console.log(greet('world'));
`
</script>
````

### Reproduction

https://stackblitz.com/edit/vite-ucyboc?file=docs%2Fexamples%2Ftest-example.md

### Expected behavior

![image](https://github.com/user-attachments/assets/769189a0-a274-478c-827c-80057a982011)

### System Info

```Text
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 16.46 GB / 31.91 GB
  Binaries:
    Node: 20.17.0 - ~\AppData\Local\fnm_multishells\29352_1729538469442\node.EXE
    npm: 10.8.2 - ~\AppData\Local\fnm_multishells\29352_1729538469442\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.86)
    Internet Explorer: 11.0.22621.3527
```
```


### Additional context

_No response_

### Validations

- [X] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

**Comment by brc-dd** [comment]
See #2315

return html from your loader and use `<div v-html="data" />`

https://github.com/vuejs/vitepress/discussions/2921#discussioncomment-7023589

**Comment by brc-dd** [comment]
it should be:

```ts
  return md.render(
    `
\`\`\`ts
${scriptText}
\`\`\`
    `
  )
```

(remove leading spaces from the code block, otherwise it will be considered an [indented code block](https://spec.commonmark.org/0.31.2/#indented-code-blocks))

PR #4321, opened 2024-10-29
perf: cache shiki

Barbapapazes:
### Description

Hello 👋,

I'm building my website, https://soubiran.dev, with VitePress.

At first, it was fast but as I added more content and more data loaders, it became slower and slower to build. Today, it takes around 4 minutes to build the website and each time I add a new page, it adds 10 seconds to the build time.

So I decided to investigate to find the annoying issue. At the same time, I had this warning: `Shiki is supposed to be used as a singleton, consider refactoring your code to cache your highlighter instance;`.

I moved all my content to a fresh VitePress project and I found that the speed build was back to normal. In my project, soubiran.dev, I started to play with `DEBUG=vitepress:md` and I discover that some pages took 27s to build while some was under 100ms. This led me to the conclusion that the issue was related to the markdown content and potentially the parser: MarkdownIt.

With the warning in mind, I quickly understand the highlighter creation was the issue: https://github.com/vuejs/vitepress/blob/main/src/node/markdown/markdown.ts#L208

I also discover that every time a `createContentLoader` is used, a new instance of MarkdownIt is created: https://github.com/vuejs/vitepress/blob/main/src/node/contentLoader.ts#L129

So I try to patch VitePress in my project by caching the highlighter instance.

Before, in the Cloudflare Pages CI: more than 4 minutes to build the website.
After, in the Cloudflare Pages CI: 47 seconds to build the website.

related to https://x.com/soubiran_/status/1851189188634288614

This also improve significantly the dev startup of the project.

**Unfortunately, this project is private but I can happily invite  maintainers if needed.**

### Additional Context

---

> [!TIP]


Barbapapazes:
_While I was writing this PR, I was wondering if we couldn't go further and cache the MardownIt instance?_

brc-dd:
Actually the issue is we need to reload the highlighter instance and destroy the old one too in certain cases like reloading config file (because a user might change shiki's theme and expect it to reload). But yeah I'm seeing perf degradation with both cold start (and config reload).

If possible, can you add me temporarily with read-only access on the repo? You can dm me on discord if you've any concerns.

But from what you've told, I believe the issue in your case is because of using createMarkdownRenderer in data loader (similar to note in https://github.com/vuejs/vitepress/issues/4300#issuecomment-2439675326). We probably should provide better API instead of createMarkdownRenderer, or maybe an option to re-use the instance (enabled by default, but disabled inside our vite plugin so that on config reload it still creates and caches a new instance but inside data loaders, etc. it can re-use the cached one).

Barbapapazes:
> /publish

I've updated soubiran.dev and this PR is in production! (build time: 30 seconds on CF, 13 seconds on my computer and a way better dev start time)

Barbapapazes:
I just update my dependencies, and now, it's under 10 seconds 🥳

Screenshot 2024-11-02 at 19 34 24

PR #4326, opened 2024-10-31
perf: lazy load shiki languages

brc-dd:


KnorpelSenf:
We are seeing ~75 % faster builds since we updated from vitepress 1.4.1 to 1.5.0 and this is the only pull request related to performance. Is that possible?

brc-dd:
@KnorpelSenf There is #4321 too.

KnorpelSenf:
Thanks for explaining! We're using very few languages but a large number of code examples, so that matches up. Very cool changes!

PR #4379, opened 2024-11-23
typo: wrong translation in markdown.md

MC-XiaoHei:
### Description

in Chinese ver of markdown.md, there is an translation error

### Linked Issues

no

### Additional Context

nothing

### DIFF
--- a/docs/zh/guide/markdown.md
+++ b/docs/zh/guide/markdown.md
@@ -891,7 +891,7 @@ $$ x = {-b \pm \sqrt{b^2-4ac} \over 2a} $$
 export default {
   markdown: {
     image: {
-      // 默认启用图片懒加载
+      // 默认禁用；设置为 true 可为所有图片启用懒加载。
       lazyLoading: true
     }
   }
--- a/docs/zh/guide/markdown.md
+++ b/docs/zh/guide/markdown.md
@@ -891,7 +891,7 @@ $$ x = {-b \pm \sqrt{b^2-4ac} \over 2a} $$
 export default {
   markdown: {
     image: {
-      // 默认禁用图片懒加载
+      // 默认启用图片懒加载
       lazyLoading: true
     }
   }


brc-dd:
I'm not a Chinese speaker, but the English version of that comment is "image lazy loading is disabled by default", and the Chinese should be something like "默认情况下，图片懒加载功能是禁用的" 👀

MC-XiaoHei:
> I'm not a Chinese speaker, but the English version of that comment is "image lazy loading is disabled by default", and the Chinese should be something like "默认情况下，图片懒加载功能是禁用的" 

i'm not very sure, but in my opinion, in context, this comment should be "image lazy loading is enabled by default"
cause you just set `lazyLoading` to true, then in markdown, lazyLoading should be enabled by default

MC-XiaoHei:
Maybe it would be better to change it to"now enable lazy loading in markdown by default"?

brc-dd:
Ah, we want to tell that lazyLoading is disabled by default but you can enable it for all images by setting markdown. lazyLoading to true. How would you write that in Chinese?

MC-XiaoHei:
```js
export default {
  markdown: {
    image: {
      // 默认情况下在 markdown 中默认禁用图片懒加载（即该配置默认为 false）
      // 若想要在 markdown 中将默认启用图片懒加载，应将其设置为 true 。
      lazyLoading: true
    }
  }
}
```
maybe better?

awxiaoxian2020:
Hi, @brc-dd He means if the comment is "image lazy loading is disabled by default", the code should be

```js
export default {
  markdown: {
    image: {
      // image lazy loading is disabled by default
      lazyLoading: false
    }
  }
```


@MC-XiaoHei this comment I think it doesn't describe the code while just tell us why we need `lazyLoading: true`.

MC-XiaoHei:
For me personally, this is a bit ambiguous, and if you think this is correct, you can close this PR

brc-dd:
Would something like this work - "默认禁用；设置为 true 可为所有图片启用懒加载。"

PR #4382, opened 2024-11-24
feat: markdown file inclusion support specify title

### LINKED ISSUE #4375: 包含 md 文件的功能中，被包含的文件可能会经常变动。如果可以按照title 进行包含就好了
包含 md 文件的功能中，被包含的文件可能会经常变动。如果可以按照title 进行包含就好了；

just  like:
``` markdown

<!--@include: ./parts/basics.md{#titleA}-->
<!--@include: ./parts/basics.md{#titleB}-->
<!--@include: ./parts/basics.md{#titleC}-->

```

那就太棒啦

btea:
### Linked Issues
close #4375

### Additional Context

---

> [!TIP]


btea:
Could you please help review this PR? Thanks! @brc-dd

brc-dd:
It's bit tricky than that.
- Need to handle nesting. 
- lso, what should be done when same title is there many times?
- `/[^#]/g` matches `^` or `#`, you probably meant to write `/^#+/`

btea:
> * Need to handle nesting.
> * Also, what should be done when same title is there many times?

Indeed, I thought too simply.
> * `/[^#]/g` matches `^` or `#`, you probably meant to write `/^#+/`

I reviewed it and there is no mistake here. 

`/[^#]/g`matches all characters except `#`. I intentionally remove all characters here to get the length of the title `#` and find the end of the content under the matching title.

brc-dd:
Ah right, sorry. So, you're essentially counting number of # in a heading? But this will give prefix length for `# hi # there` = 2. Might be better to do `str.match(/^#+/)?.[0]?.length`

btea:
Yeah,  you wrote it is better for the situation you mentioned. 👍

btea:
> * Need to handle nesting.

The interior looks to have handled the situation.
https://github.com/vuejs/vitepress/blob/d1f2afdf0fbb022f12cc12295723b3b7c7ef5cb1/src/node/utils/processIncludes.ts#L63
> * Also, what should be done when same title is there many times?

Generally speaking, a page should not have multiple identical titles. If this is the case, perhaps we can keep the current logic and give a warning message.

brc-dd:
@btea Can you review the implementation in #4608? The main change is it using heading ids (the one you see in url fragments) instead of whole title. It solves various cases where there are things like badges or markdown syntax in headings, or even duplicate heading.

PR #4389, opened 2024-11-27
i18n.md : add missing variable

nicolas-bastien:
### Description

inBrowser was missing in the imported data.


nicolas-bastien:
Thanks @brc-dd, I noticed I forgot the translated file so here the PR to finish the correction https://github.com/vuejs/vitepress/pull/4391

PR #4412, opened 2024-12-08
docs: fix vitepress execution via bunx not bun

hadihammurabi:
### Description
When I am trying to start new project using bun, this problem happen.
The docs said using this command to init new vitepress project.
```bash
bun vitepress init
```

I got error:
`error: Script not found "vitepress"`

As i know what should i do, it's no problem for me.
Immediately i run bunx command like this.
```bash
bunx vitepress init
```
Everything ok, right now.

But, i think something need to do at the docs.
Then, i make this pull request.

### Additional Context

---

> [!TIP]

### DIFF
--- a/docs/en/guide/getting-started.md
+++ b/docs/en/guide/getting-started.md
@@ -80,7 +80,7 @@ $ yarn vitepress init
 ```
 
 ```sh [bun]
-$ bun vitepress init
+$ bunx vitepress init
 ```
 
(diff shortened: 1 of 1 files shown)


brc-dd:
Read that whole page. You should install vitepress first. Even if you run `bunx vitepress init` it won't add vitepress to your deps and you'll need to do that manually after init.

PR #4467, opened 2025-01-06
feat!: upgrade vite to v6

sxzz:
### Description

Upgrade Vite to v6

### Additional Context

It's a breaking change in Vite, but it seems no any further changes are needed in VitePress. Please consider bumping the major or minor (at least) version on VP side (Vitest did a major).

---

> [!TIP]

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "mark.js": "8.11.1",
     "minisearch": "^7.1.1",
     "shiki": "^1.24.4",
-    "vite": "^5.4.11",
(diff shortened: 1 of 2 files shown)


brc-dd:
https://github.com/vuejs/vitepress/blob/c61182ab278350699b5d50461788478a340790aa/__tests__/e2e/data-loading/data.test.ts#L47

This test will break. It is skipped but it should work. Need to migrate to new hotUpdate hook for it to work.

PR #4493, opened 2025-01-14
fix(theme): rendering of sponsor and team components inside vp-doc containers

OrbisK:
### Description

I am not sure why we need js to calculate the window width. This leads to a hydration mismatch:

example vueuse docs:
image

By replacing the js calculation with css vw this should be solved.

### Additional Context

Maybe I missed something.

---

> [!TIP]


bluwy (approved):
dcb7a75532c5472060ec379d25a5fafbc7932637 added the `data-allow-mismatch="style"` to suppress the error, but I agree that there's no need to compute the vw in js at all.

brc-dd:
It was there to adjust for scrollbar width I think. I don’t remember completely, though.

bluwy:
Ok I see what it's trying to do. It wants to expand the inner sponsor/team component content to full width like below:

image

So it adds a negative margin to the parent, so the inner container's default padding offsets it back to align to the container content area.

I think there's much simpler ways to fix this, like disabling the padding would probably be the easiest. I'll try some ways to fix this.

Also, the current intent also doesn't seem to work because `useWindowSize` only returns a value during the first resize. On page load, it's zero and scrollbar issue and causes a horizontal scrollbar to appear.

bluwy:
This should be good now. I tested this by adding this markup in `docs/en/index.md`:

<details><summary>Details</summary>

```html
<script setup>
import {
  VPHomeSponsors,
  VPTeamPage,
  VPTeamPageTitle,
  VPTeamPageSection,
  VPTeamMembers
} from 'vitepress/theme'
</script>

<VPHomeSponsors
  actionLink="https://github.com/sponsors/vuejs"
  message="Help"
  :data="[]"
/>

<VPTeamPage>
  <VPTeamPageTitle>
    <template #title>Meet the Team</template>
    <template #lead>
      Test
    </template>
  </VPTeamPageTitle>
  <VPTeamMembers :members="[]" />
  <VPTeamPageSection>
    <template #title>Advisors</template>
    <template #lead>
      Test
    </template>
    <template #members>
      <VPTeamMembers size="small" :members="[]" />
    </template>
  </VPTeamPageSection>
</VPTeamPage>
```
</details> 

And checking the alignment is right from devtools.

brc-dd:
Hmm, yeah it was for full page bleed. There is margin outside that container on larger widths. `padding: 0` I think won't extend that divider? Also, seems like other styles are also broken. I'll push a fix.

PR #4502, opened 2025-01-21
fix(VPNav): Allow clicks on custom navbar

michaelcozzolino:
### Description

in my vitepress project I wanted to use a custom navbar instead of the default vitepress one

so i made an alias in `vite.config.ts`

```
resolve: {
        alias: [
            {
                find:        /^.*\/VPNavBar\.vue$/,
                replacement: resolve(__dirname, '.vitepress/theme/Components/MyCustomNavbar.vue'),
            },
        ],
    },
```
the navbar got replaced but I couldn't be able to click any link on it. this PR fixes it by setting the pointer-events to auto

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/components/VPNav.vue
+++ b/src/client/theme-default/components/VPNav.vue
@@ -45,7 +45,7 @@ watchEffect(() => {
   left: 0;
   z-index: var(--vp-z-index-nav);
   width: 100%;
-  pointer-events: none;
+  pointer-events: auto;
   transition: background-color 0.5s;
 }
 


brc-dd:
It is intentional for the default theme. If you are patching components, you should override it using CSS.

michaelcozzolino:
Shouldn't vitepress be a plug and play framework? I mean it feels weird to override something to make something else work, if it really provides a smooth replacement for the default components there should not be any override at all, furthermore the default navbar already has the same property in itself and my fix seems to not change anything regarding the whole default navigation but maybe I'm wrong.

brc-dd:
It prevents clicking on top of scrollbar. You can try resizing the window and moving the scrollbar of sidebar. It was added in https://github.com/vuejs/vitepress/commit/44953dcd1e0224bae95666c87e368b9d3fdf17ae

PR #4508, opened 2025-01-22
fix: fix static content removal for lean chunks due to Vue 3.5 changes

yyx990803:
@brc-dd we need to do some manual testing to make sure this works with Vue 3.4 and below (could be explicitly installed by user)

Also let's test this for Vue and Vite docs before releasing.

### DIFF
--- a/src/client/app/index.ts
+++ b/src/client/app/index.ts
@@ -119,21 +119,14 @@ function newApp(): App {
 
 function newRouter(): Router {
   let isInitialPageLoad = inBrowser
-  let initialPath: string
 
   return createRouter((path) => {
     let pageFilePath = pathToFile(path)
     let pageModule = null
 
     if (pageFilePath) {
+      // use lean build if this is the initial page load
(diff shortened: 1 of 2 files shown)


src/client/app/index.ts
```diff
@@ -119,21 +119,14 @@ function newApp(): App {
...
 function newRouter(): Router {
   let isInitialPageLoad = inBrowser
-  let initialPath: string
 
   return createRouter((path) => {
     let pageFilePath = pathToFile(path)
     let pageModule = null
 
     if (pageFilePath) {
+      // use lean build if this is the initial page load
```
yyx990803, line 128: Because in 3.5 the static vnodes are cached per-instance instead of globally hoisted, we can no longer rely on adopted content during hydration when navigating back to the initial page. We have to load the full chunk here in that case.

brc-dd (approved):
Tested on vite and vue docs. Seems to work fine with vue 3.4 too.

PR #4517, opened 2025-01-24
Using Environment Variables In VitePress - issue#4516

### LINKED ISSUE #4516: Using Environment Variables In VitePress
Hey guys, thank you for such an incredible project.

Today I was trying to integrate Algolia Search but found it hard how to use environment variables until I stumbled upon [the General Workflow for Connecting to a CMS](https://vitepress.dev/guide/cms#general-workflow)

Let's add a section about how to use such environment variables.
I will try opening a PR to add a section for environment variables configuration on the VitePress reference page and wait for your suggestions

Thanks,

KareimGazer:
### Description

I've created a page in VitePress reference for how to use environment variables in VitePress

### Linked Issues

closes #4516

### DIFF
--- a/docs/en/reference/environment-variables.md
+++ b/docs/en/reference/environment-variables.md
@@ -1,0 +1,69 @@
+---
+outline: deep
+---
+
+# Environment Variables
+
+## Basic Workflow
+
+Using environment variables in VitePress is essential for managing sensitive data such as API keys and service URLs.
+VitePress allows you to load environment variables from `.env` files using the `loadEnv` function to load these variables into your configuration (e.g. `.vitepress/config.ts`).
+
+### Step 1: Create the .env File
+
+The first step in setting up environment variables is to create a `.env` file in the root directory of your VitePress project.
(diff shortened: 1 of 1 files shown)


docs/en/reference/environment-variables.md
```diff
@@ -0,0 +1,69 @@
+---
+outline: deep
+---
+
+# Environment Variables
+
+## Basic Workflow
+
+Using environment variables in VitePress is essential for managing sensitive data such as API keys and service URLs.
```
skirtles-code, line 9: I think this could be misleading.

Using environment variables won't protect sensitive data. The values of the variables end up in the built JS files, where anyone can access them.

KareimGazer: Yes, that's true
I used the API keys as GitHub secrets in my CI/CD pipeline but still these values need to be read from the environment even if they are not written in the `.env` file.

Could you think of a better, more specific example?

1357310795:
very useful guide!

PR #4608, opened 2025-03-08
feat: support using header anchors in markdown file inclusion

### LINKED ISSUE #4375: 包含 md 文件的功能中，被包含的文件可能会经常变动。如果可以按照title 进行包含就好了

brc-dd:
closes #4375
closes #4382

### DIFF
--- a/src/node/utils/processIncludes.ts
+++ b/src/node/utils/processIncludes.ts
@@ -1,18 +1,20 @@
 import fs from 'fs-extra'
 import matter from 'gray-matter'
+import type { MarkdownItAsync } from 'markdown-it-async'
 import path from 'node:path'
 import c from 'picocolors'
 import { findRegion } from '../markdown/plugins/snippet'
 import { slash } from '../shared'
 
 export function processIncludes(
+  md: MarkdownItAsync,
   srcDir: string,
   src: string,
   file: string,
   includes: string[]
 ): string {
   const includesRE = /<!--\s*@include:\s*(.*?)\s*-->/g
-  const regionRE = /(#[\w-]+)/
+  const regionRE = /(#\S+)/
   const rangeRE = /\{(\d*),(\d*)\}$/
 
   return src.replace(includesRE, (m: string, m1: string) => {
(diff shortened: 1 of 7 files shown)


src/node/utils/processIncludes.ts
```diff
@@ -1,18 +1,21 @@
...
   srcDir: string,
   src: string,
   file: string,
-  includes: string[]
+  includes: string[],
+  cleanUrls: boolean
 ): string {
   const includesRE = /<!--\s*@include:\s*(.*?)\s*-->/g
-  const regionRE = /(#[\w-]+)/
+  const regionRE = /(#[^\s\{]+)/
```
btea, line 18: Spaces should be allowed between the title and #. Generally, spaces are retained when writing md.

brc-dd: So the proposed usage is like this:

```
<!--@include: ./parts/basics.md#heading-id-->
```

"heading-id" here is basically what you write in inks like `[some section](#heading-id)`. These don't contain space. In your PR, you were keeping whole title there which can contain spaces.

btea: Yes, because of usage habits, people may be used to adding a space between # and the title. Perhaps, we can add bold to emphasize it in the document.

btea: There seems to be no warning message when there are multiple copies of the same title in the same document.

brc-dd: > There seems to be no warning message when there are multiple copies of the same title in the same document.

There won't be multiple headings with same id (unless user is manually doing so).

```md
# Hi

## Section - 1

### Sub Section

## Section - 2

### Sub Section
```

this is rendered like this:

image

\
so using `<!--@include: ./foo.md#sub-section-->` will point to the first one, `<!--@include: ./foo.md#sub-section-1-->` will point to the second one.

btea: OK, got it. But this seems to match the result after conversion. Maybe we should explain this in the documentation to reduce user confusion.

brc-dd: updated: 👀 

image

btea: Looks great.

PR #4620, opened 2025-03-13
Update Shiki to 3.2.1

ouillie:
### Description

Just bumping Shiki up from `3.1.0` -> `3.2.1`.

### Linked Issues

I'm interested in this because it adds new syntax highlighting languages.

### Additional Context

Tested with `pnpm run docs`, tried out the new languages and they work.


brc-dd:
Just delete your node_modules and lockfile. shiki’s version isn’t locked.

ouillie:
I'm far from a Node expert, but simply doing that does not working for me. Seems like the transitive dependency resolution is pretty complicated.

brc-dd:
Which vitepress version are you on? vitepress v1 don’t use shiki v3.

ouillie:
I see. I was on 1.6.3 and switching to 2.0.0-alpha.4 seems to have brought in the newer version of Shiki. Thanks for the tip!

PR #4623, opened 2025-03-15
chore: use SVG logo on main website page

### LINKED ISSUE #4198: Make the main icon on the site in svg format

The current picture is not clear

Make the same logo in SVG format

Increase the clarity of the main icon

### Additional context

In my opinion, it would be much better if the main icon was in SVG format, because the current one is not very clear (at least it is visible on a 2.8K display)

### Validations

okineadev:
Closes #4198


okineadev:
https://mokacoding.com/blog/symliks-in-git/

I recommend reading this article as it prevents problems with symbolic links in Git

brc-dd (commented):
LGTM but I think there was a shadow in the webp version to create some sense of depth.

\cc @kiaking how was that generated? I didn't find it in figma 👀

kiaking:
It was created using Photoshop, and yes was using PNG (-> webp) in order to add drop shadows 👀 

Is there a way to add subtle shadows using SVG? 🤔

brc-dd:
Maybe https://stackoverflow.com/a/13624469/11613622 👀 What were the shadow params in photoshop? We can probably just add drop shadow on whole image using css.

kiaking:
Was looking for but the original file is gone 😅  Anyway Photoshop shadow params are totally different from CSS anyway so... I think something like this should work well 🤔 

```css
drop-shadow: 0 12px 32px rgba(0, 0, 0, .1), 0 2px 6px rgba(0, 0, 0, .08);
```

But last time I tries (few years ago) applying drop shadow with css in SVG file added rectangle shadow, so I assume we need to add this shadow inside SVG file, I guess 👀

brc-dd:
seems to work (checked with red for more contrast):

image

\
needs this on ``: 👀 

```css
filter: drop-shadow(0 12px 32px rgba(0, 0, 0, .1)) drop-shadow(0 2px 6px rgba(0, 0, 0, .08))
```

kiaking:
Uhhhh nice! Yeah this should be much better 👍

PR #4630, opened 2025-03-18
feat(customization): sidenav components / skip title update

Tynopia:
### Description  

Hi, we've added the following features to VitePress:  

- Custom side navigation components, allowing for greater flexibility in modifying passed content. This is achieved by overriding the component within `enhanceApp`.  

- A boolean flag in `themeConfig`, enabling users to bypass the original title update logic. This provides more control over how titles are handled within the theme.


Tynopia:
Ready for review

brc-dd:
> A boolean flag in themeConfig, enabling users to bypass the original title update logic. This provides more control over how titles are handled within the theme.

It should be outside themeConfig. DefaultTheme.Config is only for default-theme stuff. This option will impact people using other themes too.

Also, how will you use this flag? I see that you want to render dynamic things in document.title. How exactly though?

Tynopia:
So in the sidenav and local search components, we can use `compile` with the current `this` instance without any issues. However, to get it running, I had to use `compile` from `vue/dist/vue.esm-bundler.js`, which may not be possible for other runtimes.  

That’s also why developers should have the opportunity to use it based on their specific needs.

Tynopia:
@brc-dd Any new infos here?

PR #4634, opened 2025-03-19
feat: active outline link scroll to page center

btea:
### Description
When I clicked a link to jump to the page, the page content was centered according to the hash, but the active links in the sidebar did not scroll to the middle of the page, resulting in the inability to find the corresponding outline link for a while.

https://vitest.dev/guide/cli.html#testtimeout

### Additional Context

---

> [!TIP]


btea:
@brc-dd Do you think this modification is necessary?

bluwy:
This should be covered in #5377 now which will scroll to nearest into view instead of center.

btea:
Looks great. 👍

PR #4639, opened 2025-03-21
fix(theme-default): ensure proper sizing of SVG hero images

GabrielxDuO:
### Description

When using vector graphics (SVGs) without specified dimensions as hero images, they can render at unexpectedly small sizes in certain viewport widths. This creates inconsistent presentation compared to bitmap images.

### Before/After Comparison

|            | WebP                                                         | SVG                                                          |
| ---------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
| **Before** | before-webp | before-svg |
| **After**  | after-webp | after-svg |

### DIFF
--- a/src/client/theme-default/components/VPHero.vue
+++ b/src/client/theme-default/components/VPHero.vue
@@ -323,7 +323,7 @@ const heroImageSlotExists = inject('hero-image-slot-exists') as Ref<boolean>
   max-width: 192px;
   max-height: 192px;
   width: 100%;
-  height: 100%
+  height: 100%;
   object-fit: contain;
   /*rtl:ignore*/
   transform: translate(-50%, -50%);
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPHero.vue
```diff
@@ -322,6 +322,8 @@ const heroImageSlotExists = inject('hero-image-slot-exists') as Ref<boolean>
   left: 50%;
   max-width: 192px;
   max-height: 192px;
+  height: 100%;
+  aspect-ratio: 1;
```
brc-dd, line 326: aspect-ratio of images might not be 1

```suggestion
  width: 100%;
  height: 100%
  object-fit: contain;
```

GabrielxDuO:
Yes, I also tried using the following approach:

```css
width: 100%;
height: 100%;
object-fit: contain;
```

However, thinking back to when I tested it with `vitepress-logo-large.webp`, the inconsistency might have been caused by an outer glow or shadow (as opposed to the glow in `vitepress-logo-large.svg`, which is achieved using `filter`), making the `.webp` version appear slightly larger and therefore seem to scale down more noticeably. 

That said, considering the broader issue that the hero image may not always be square, this solution (`object-fit: contain`) is indeed better suited. Thank you for making these adjustments!

brc-dd:
Yeah, and the webp was also of older version I think. 

image

\
(left one is svg, right is webp)

PR #4647, opened 2025-03-25
refactor: avoid assigning to bundle object

sapphi-red:
### Description

[Assigning to `bundle` object is discouraged by rollup](https://rollupjs.org/plugin-development/#generatebundle:~:text=DANGER,this.emitFile.) and is not supported by rolldown.

This PR changes the code using assignment to use `this.emitFile` instead. This would make vitepress work with Rolldown powered Vite.

It should not cause any change unless there's a code that expects the lean chunks to be `type: 'chunk'` instead of `type: 'asset'`.

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/node/plugin.ts
+++ b/src/node/plugin.ts
@@ -366,15 +366,12 @@ export async function createVitePressPlugin(
             pageToHashMap![chunk.name.toLowerCase()] = hash
 
             // inject another chunk with the content stripped
-            bundle[name + '-lean'] = {
-              ...chunk,
+            this.emitFile({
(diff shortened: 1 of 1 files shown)


brc-dd (approved):
Thanks!

brc-dd:
It seems like this is causing issues as that asset might contain stuff like `__VITE_PRELOAD__` for example in case of dynamic imports. See #4737

sapphi-red:
I see. I guess adding `order: 'post'` to `generateBundle` would fix that.

PR #4660, opened 2025-03-30
feat(config): Hierarchical site configuration using distributed config files

### LINKED ISSUE #4659: Multi-layer hierarchical configuration overloading
### Description

Currently, a `vitepress` site is configured through a centralized monolithic configuration entry. This makes it difficult to apply fine grained configuration alternation for a specific part of the website.

For example, to achieve multilingual support, the official documentation had to organize it's configuration as follows:

```plaintext
docs/.vitepress/config
├── en.ts
├── es.ts
├── fa.ts
├── index.ts
├── ko.ts
├── pt.ts
├── ru.ts
├── shared.ts
└── zh.ts
```

In addition to the organization of the configuration files, the UI components have to be rewritten to provide locale support. For example, the search box component has to provide `locales` configuration entry and programmatically derive current locale using global state. These extra burdens will likely add up and become unmaintainable in the near future.

We can instead provide a distributed configuration hierarchy based on the same folder structure used to organize markdown contents.

For example, the multilingal support can be achieved as follows:

```plaintext
docs
├── .vitepress
│   ├── config.ts  # Global configuration (root)
│   └── ...
├── en
│   ├── config.ts  # English specific configuration
│   ├── index.md
│   └── ...
├── zh
│   ├── config.ts  # Chinese specific configuration
│   ├── index.md
│   └── ...
└── ...
```

In the example above, `docs/en/config.ts` only applies to pages under `docs/en/` directory, and the same can be said for other language specific configuration files.

### Additional Benefits

Almost any configuration can be fine-adjusted in the suggested architecture without specific adaptation. Features added in the future will also be automatically compatible. This will replace the `locale` entry in the current configuration interface.

As an example, I simply added [6 lines of code](https://github.com/zhangyx1998/vitepress/commit/c3c0d71d1d5ee6244e6b51abef4bbead713aac03#diff-169e70e8b8105432b9949c9ab7ca95265c4ed4476b7056f05f50cf21b50bedbdR44-R49) in `docs/zh/config.ts` and made a Chinese version of the 404 page, effective for ANY page under the `/zh/` path:

| Default 404 Page (English) | Chinese Specific 404 Page |
|:---:|:---:|
|Image|Image|
|[preview link](https://deploy-preview-4660--vitepress-docs.netlify.app/404)|[preview link](https://deploy-preview-4660--vitepress-docs.netlify.app/zh/404)|

This is currently **impossible** because `NotFound.vue` does not provide an locale configuration interface.
After introducing the proposed feature, it **automatically** supports multi-language - _without changing one line of code inside it_!

### Validations

zhangyx1998:
### Description

> Please see #4659 for more context.

#### Summary of changes

1. [‎src/client/app/data.ts](https://github.com/zhangyx1998/vitepress/commit/c3c0d71d1d5ee6244e6b51abef4bbead713aac03#diff-0ea7943937a6fe3ca683f99fd58d5d2fcac661ad25dbd04e39c48d304967e9a7):

    + Added a [wildcard import](https://github.com/zhangyx1998/vitepress/commit/c3c0d71d1d5ee6244e6b51abef4bbead713aac03#diff-0ea7943937a6fe3ca683f99fd58d5d2fcac661ad25dbd04e39c48d304967e9a7R73-R83) and [supporting utilities](https://github.com/zhangyx1998/vitepress/commit/c3c0d71d1d5ee6244e6b51abef4bbead713aac03#diff-0ea7943937a6fe3ca683f99fd58d5d2fcac661ad25dbd04e39c48d304967e9a7R85-R102) to include configuration files from user directory.
    + Modified [`initData()`](https://github.com/zhangyx1998/vitepress/commit/c3c0d71d1d5ee6244e6b51abef4bbead713aac03#diff-0ea7943937a6fe3ca683f99fd58d5d2fcac661ad25dbd04e39c48d304967e9a7R105-R112) to dynamically compute `siteData` and apply extra configuration layers based on `route.relativePath`.

2. [‎src/client/app/utils.ts](https://github.com/zhangyx1998/vitepress/commit/c3c0d71d1d5ee6244e6b51abef4bbead713aac03#diff-bf426b0e56731a77e9ad18f74d584a9235c6754388fb67a317b703da3f3a0bd9):

    + Added `stackView()` and supporting utilities to merge multiple layers of objects into a readonly view, without mutating the original objects - this is important because certain layers may be dropped later when user navigates outside its scope.

3. Configuration Files

    The configuration files are moved from `.vitepress/config` to their corresponding language subdirectories (e.g. `docs/.vitepress/config/en.ts` -> `docs/en/config.ts`) to demonstrate the benefits of the proposed changes. In addition, a language-specific locale is added for 404 pages under `/zh/` directory. It is reflected in the rendered preview.

### Linked Issues

closes #4659 
closes #4654 

### Additional Context

Please refer to the linked issue for highlights of this PR.

---

> [!TIP]

/publish


src/client/theme-default/support/translation.ts
```diff
@@ -15,6 +15,7 @@ export function createSearchTranslate(
     const isObject = themeObject && typeof themeObject === 'object'
     const locales =
       (isObject && themeObject.locales?.[localeIndex.value]?.translations) ||
+      (isObject && themeObject?.translations) ||
```
brc-dd, line 18: why is this needed?

src/node/build/bundle.ts
```diff
@@ -139,6 +139,12 @@ export async function bundle(
                   ) {
                     return 'framework'
                   }
+                  if (/(^|\/)config\.([cm]?js|ts|json)$/.test(id)) {
+                    return 'framework'
+                  }
+                  if (/(^|\/)vitepress\/package.json$/.test(id)) {
```
brc-dd, line 145: this seems like hardcoding it just to support our config?

brc-dd:
An obvious downside is we can't use node stuff in configs. What if we glob and load all the configs at plugin-level instead of doing it via import.meta.glob in client code?

brc-dd:
Also, the site appears to be broken. There is probably a circular dependency somewhere in generated code.

src/client/theme-default/support/translation.ts
zhangyx1998: Because this PR will make it unnecessary to enumerate translations as locales.

#### Previous configuration looks like:

```js
export default {
  themeConfig: {
    search: {
      options: {
        locales: {
          en: { translations: {/* ... */} },
          zh: { translations: {/* ... */} },
          fa: { translations: {/* ... */} },
          ko: { translations: {/* ... */} },
          ru: { translations: {/* ... */} },
          // etc...
        }
      }
    }
  }
}
```

#### New configuration will look like:

```js
export default {
  themeConfig: {
    search: {
      options: {
        translations: {
            // Dynamically overloaded based on route
        }
      }
    }
  }
}
```

In other words, `search.options.locales.[locale].translations` can be phased out in favor of `site.theme.search.translations` after adaption of this PR.

zhangyx1998: It was added by mistake. I was trying to solve a problem that is not caused by this.

I've reverted this change.

zhangyx1998:
> An obvious downside is we can't use node stuff in configs. What if we glob and load all the configs at plugin-level instead of doing it via import.meta.glob in client code? We can create a [virtual module](https://vite.dev/guide/api-plugin#virtual-modules-convention) that exports `extraConfig`. Probably `resolveUserConfig` can be adjusted and reused. That would also avoid hardcoding things in `manualChunks`.

This seems like a better solution. I'll try to adapt this approach.

src/node/build/bundle.ts
zhangyx1998: This has been reverted.

zhangyx1998:
### Updates:

#### 1. Removed `import.meta.glob` from client side. Config collection is now implemented in [`resolveUserConfig()`](https://github.com/vuejs/vitepress/pull/4660/files#diff-183b42f9c3bb38bbeb6c4522590df6e7152643045f0d1ca4bf3887da0a8628ceR236-R245).
#### 2. Instead of creating a virtual module `@additionalConfig`, an optional configuration entry `UserConfig.additionalConfig` is implemented for better flexibility:

+ **When left undefined**, `resolveUserConfig` will automatically collect all eligible config files and organize them into a dictionary as its default value.

+ **When defined as an object**, the client code will auto select config layers to apply. For example:

    ```js
    import en_config from './en'
    import zh_config from './zh'
    export default defineConfig({
        /* ... */
        additionalConfig: {
            '/en/': en_config,
            '/zh/': zh_config,
        }
    })
    ```

+ **When defined as a function** `(path: string) => AdditionalConfig[]`, user can have even higher flexibility defining which configuration layers should be applied, and in what order they will apply:

    ```js
    import en_config from './en'
    import zh_config from './zh'
    export default defineConfig({
        /* ... */
        additionalConfig(path) {
            return [{
                title: `Hello from ${path}`
            }]
        }
    })
    ```

    Produces:

    image

@brc-dd can you please review this again? Appreciate it!

brc-dd:
Thanks a lot! 🙌 Overall looks good. I'll test some stuff locally and merge if nothing pops up.

types/shared.d.ts
```diff
@@ -161,6 +162,27 @@ export interface LocaleSpecificConfig<ThemeConfig = any> {
   themeConfig?: ThemeConfig
 }
 
+export interface AdditionalConfigEntry<ThemeConfig = any>
+  extends LocaleSpecificConfig<ThemeConfig> {
+  /**
+   * Source of current config entry, only available in development mode
+   */
+  src?: string
```
brc-dd, line 170: where is this used?

brc-dd:
```ts
if (mode === 'development') (configExports.config as any).VP_SOURCE = file
```

Might be better to make it something like `__src` or something.

brc-dd: Ah, nevermind. Stuff with `_` is stripped. I'll update the types.

zhangyx1998: My first attempt was `__src__` before I realized that 🤣

zhangyx1998: I forgot to remove the type annotation that you pointed out. Please feel free to clean it up!

zhangyx1998:
### Summary of update:

#### Code Cleanup

+ Removed type annotation of injected `VP_SOURCE` symbol (internal use only, under dev mode).
+ Renamed the injection symbol from `VP_SOURCE` to `[VP_SOURCE]`. It is now impossible to be confused with a JS variable nor a legal URL path element. The new symbol is consistent with the one introduced in #4663.

#### Typing Helper

+ Added type `DeepPartial<...>` to reflect the nature of type `AdditionalConfig`
+ Added type helper `defineAdditionaConfig()` and `defineAdditionalConfigWithTheme()`
+ Updated additional config files under `/docs/` to reflect the above update.

This PR should be ready to merge.

types/shared.d.ts
zhangyx1998: This is removed in the latest commit

brc-dd:
Give me a day or two. I'll check everything and merge it by weekend. Got bit busy with day job.

zhangyx1998:
The latest commit relocates config synthesis from `initData()` (client/data.ts) to `resolveSiteConfigByRoute()` (shared.ts). This ensures the SSR context also receives correct configuration based on its route.

zhangyx1998:
@brc-dd Since commit [27c4151](https://github.com/vuejs/vitepress/commit/27c4151c723054b9a124d7dd72ab488edaec5852), Netlify build constantly fails due to "javascript heap out of memory". My local build and test runs did not have the same problem though.

brc-dd:
ah it's failing on main too. I'll revert it

brc-dd:
Yeah. I've some changes locally, I'll push them.

brc-dd:
Something is wrong with stacking. Try opening http://localhost:5173/zh/guide/what-is-vitepress, then change language to Portuguese from switcher in navbar. It goes to `http://localhost:5173/pt/zh/guide/what-is-vitepress` but it should go to `http://localhost:5173/pt/guide/what-is-vitepress`. Also, the debug logs say:

```
Config Layers for pt/zh/guide/what-is-vitepress.md:
===================================================
1. /pt/config.ts
2. /config.ts
3. locale config (zh)
4. .vitepress/config (root)
```

here the third one shouldn't be there?

zhangyx1998:
> Something is wrong with stacking. Try opening http://localhost:5173/zh/guide/what-is-vitepress, then change language to Portuguese from switcher in navbar. It goes to http://localhost:5173/pt/zh/guide/what-is-vitepress but it should go to http://localhost:5173/pt/guide/what-is-vitepress. Also, the debug logs say:
> 
> ```
> Config Layers for pt/zh/guide/what-is-vitepress.md:
> ===================================================
> 1. /pt/config.ts
> 2. /config.ts
> 3. locale config (zh)
> 4. .vitepress/config (root)
> ```
> 
> Here the third one shouldn't be there? Seems to be broken since the ssr refactor commit - [a2a297e](https://github.com/vuejs/vitepress/commit/a2a297e724d1fc0c1ce6aa7305e0d7bb5ffa794f)

The direct cause is `getLocaleForPath()` always returns `"root"`. And this is in turn caused by `siteData.locales` being empty. However, the property does exist in the raw siteData.

image

zhangyx1998:
@brc-dd this is fixed.

The fix is very simple: [add `localeIndex` property to localeConfig]() so it masks the default localeIndex on resolved siteData.

---

There are actually two issues in the code that makes the problem intertwined:

1. `getLocaleForPath` returns the first locale that matches any segment in the path, even if another locale also exist in the same path.

    ```js
    getLocaleForPath(__VP_SITE_DATA__, 'pt/zh/index.md')
    // "zh"
    getLocaleForPath(__VP_SITE_DATA__, 'zh/pt/index.md')
    // "zh"
    ```

    ```js
    // Search Order
    {
        root: { /* last */ }, // (last)
        zh: { label: "..." }, // 1st
        pt: { label: "..." }, // 2nd
        ru: { label: "..." }, // 3rd
        es: { label: "..." }, // 4th
        ko: { label: "..." }, // 5th
        fa: { label: "..." }, // 6th
    }
    ```

    For example, the `zh` locale will always supersede other locals when `/zh/` presents in any part of the path, regardless of where it appears.

2. According to `getLocaleForPath`, the locale path segment can appear **any where** in the relativePath. For example, paths `zh/foo/bar` and `foo/zh/bar` are both considered `zh` locale. However, the computed localeLinks simply substitutes the **start of a relativePath** with the new locale.

I think this also needs to be fixed. I can do a separate PR for this later.

brc-dd (approved):

PR #4673, opened 2025-04-03
refactor(theme-default): Migrate `useSidebar` to `useLayout`, add `isHome` frontmatter option

### LINKED ISSUE #4671: [default-theme] Allow custom layout used as home layout

Currently the default theme hardcodes a string comparison to determine if current page is home layout:

```js
class = {
  'is-home': frontmatter.layout === 'home'
}
```

There are many occurrences in the default theme where the above expression is used to control special treatments for home pages (e.g. hide sidebar, disable local nav, expand content width in CSS).

In my case, I need to create a custom Home layout in addition to the one provided by the default theme. However, it's almost impossible to make a custom layout page behave like the home layout because of these hardcoded conditions.

Accept an optional config `isHomeLayout` in front-matter. When left undefined, this entry should default to `frontmatter.layout === 'home'`.

Example:

```md
---
layout: my-custom-layout
isHomeLayout: true
---

# My home content

Lorem ipsum dolor sit amet ....
```

_No response_

### Additional context

_No response_

### Validations

zhangyx1998:
### Description

Modified `useData()` (data.ts) to automatically derive `frontmatter.isHomeLayout` if it's not speficied.

Supports:

```md
---
layout: my-custom-home-layout
isHomeLayout: true
---

# My home title

Lorem ipsum dolor sit amet ....
```

Making it possible to use a custom layout as home layout in the default-theme.

### Linked Issues

closes #4671 

### Additional Context

---

> [!TIP]

/publish


src/client/theme-default/composables/data.ts
```diff
@@ -1,4 +1,12 @@
...
-export const useData: typeof useData$<DefaultTheme.Config> = useData$
+export function useData() {
+  const data = useData$<DefaultTheme.Config>()
+
+  // Allow an arbitrary layout to be used as the home layout
+  ;(data.frontmatter as any).isHomeLayout ??=
+    (data.frontmatter as any).layout === 'home'
+
+  return data
+}
```
brc-dd, line 12: Ah, maybe not a good idea. It works without an extra import, but it might be better to move it to separate composable. useData is modified here as a type helper only.

Also, this won't be reactive I guess and frontmatter is a computed ref, assigning values directly to it won't work?

zhangyx1998: This is indeed a bug, thanks for pointing that out! I'll push a fix for this.

zhangyx1998: fixed in latest push

zhangyx1998: BTW I considered 2 different design options:

### 1. `useIsHomeLayout(void)`

```ts
// is-home-layout.ts
export function useIsHomeLayout() {
  const { frontmatter } = useData()

  const isHomeLayout = computed(() => {
    if (frontmatter.value.isHomeLayout !== undefined)
      return Boolean(frontmatter.value.isHomeLayout)
    else return frontmatter.value.layout === 'home'
  })

  return { isHomeLayout }
}

```

### 2. `useDerivedFrontmatter(frontmatter: Ref<string, any>)`

```ts
// derived-frontmatter.ts
export function useDerivedFrontmatter(frontmatter: Ref<string, any>) {
  const isHomeLayout = computed(() => {
    if (frontmatter.value.isHomeLayout !== undefined)
      return Boolean(frontmatter.value.isHomeLayout)
    else return frontmatter.value.layout === 'home'
  })

  return { isHomeLayout }
}

```

Which one do you think looks better design-wise?

brc-dd: First one. But maybe something like `useLayout` and returns `isHome`. We can extend it later if we want 👀

zhangyx1998: > (Assuming `useIsHomeLayout` has been renamed to `useLayout`)

I found `useLayout` and `useSidebar` are always used together, and the functions inside `useSidebar` actually summarizes better if it's called `useLayout`. For example, property `hasAside` is not related to sidebar, but is part of the layout.

Maybe we should just move `useSidebar` out of `sidebar.ts` and put it into `composables/layout.ts` and rename it as `useLayout()` ?

brc-dd: Yeah we can do that. It will be a breaking change since useSidebar is exposed, but I think it's fine as we are in alpha. Shouldn't need much change in users' code as long as useLayout is also exposed.

PR #4693, opened 2025-04-15
ci: add emojis to workflow names and improve their security

okineadev:
### Description

I improved the security of workflows by restricting permissions and credentials, before running `zizmor` to check for problems

I also added emojis to the workflow names because they make navigation much better and it's easier to understand what they do

---

> [!TIP]


brc-dd:
I'm not in favor of the emojis. It looks cool and all but feels like overdoing it in CI runs.

`permissions` in cr.yml aren't needed. It should be kept `{}`

`cache-dependency-path` isn't needed. It's automatically inferred by the action.

release-tag.yml doesn't need write access to issues or PRs.

`persist-credentials` in release-tag.yml should be false most likely.

`.github/contributing.md` should be kept at its original place. There is no need to keep it outside.

okineadev:
> `persist-credentials` in release-tag.yml should be false most likely.

Then if I'm not mistaken, there will be no permissions at all

> `.github/contributing.md` should be kept at its original place. There is no need to keep it outside.

No, it's better to have it in the root of the project, that's how it's done everywhere

Is there any reason to store this file in the `.github` folder?

brc-dd:
> Is there any reason to store this file in the .github folder?

To avoid cluttering the main view. 

> No, it's better to have it in the root of the project, that's how it's done everywhere

GitHub allows storing it at `.github`, `docs`, and root (in decreasing precendence)  - https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors. There is no best practice here. It's a matter of opinion.

> Then if I'm not mistaken, there will be no permissions at all

It doesn't need any permission except to create a release. It probably doesn't even need the checkout step.

bluwy:
I’ll close this as we prefer to not use emojis here. If there’s other useful workflow changes it can be followed up in a new pr

PR #4694, opened 2025-04-15
fix(docs): add Netlify redirects for `llms.txt`

okineadev:


netlify.toml
```diff
@@ -26,3 +26,32 @@
...
+  force = true
+  from = "/llms-full.md"
+  status = 200
+  to = "/llms-full.txt"
+
+[[redirects]]
+  force = true
+  from = "/llms-full.txt"
+  status = 200
+  to = "/llms-full.txt"
```
brc-dd, line 57: Why are `/llms.txt -> /llms.txt` and `/llms-full.txt -> /llms-full.txt` needed?

okineadev: > Why are `/llms.txt -> /llms.txt` and `/llms-full.txt -> /llms-full.txt` needed?

Because otherwise it will be a looping redirect: `llms.txt` -> `llms.md` -> `llms.txt` -> `llms.md` -> ...

okineadev: Fixed

PR #4695, opened 2025-04-15
chore: add VS Code configuration for the project

okineadev:
### Description

Added necessary vscode extensions and settings for a flawless development experience

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 .idea
 .temp
 .vite_opt_cache
-.vscode
 dist
 cache
 temp
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["esbenp.prettier-vscode"]
+}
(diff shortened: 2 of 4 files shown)


okineadev:
> We all are using different stuff in vscode settings. Checking .vscode to source control becomes problematic in longer run. I personally get annoyed by unstaging my custom settings every time I'm committing something.

Then if you often encounter such problems, then it is exclusively a problem of your global vscode config

Can you show examples of custom project settings that you have to configure?

> Also I believe no one actually runs `npm fund` ever.

Run. There will also be a button like this on the npm package page:

brc-dd:
> Then if you often encounter such problems, then it is exclusively a problem of your global vscode config

I don't work on just one project. Every project has different settings. Every project cannot even have the same settings. Sometimes you need to enable deno, sometimes disable formatting, disable prettier and enable eslint or biome, sometimes it's just local cspell suggestions, sometimes you need to enable volar for markdown, sometimes volar errors and you need to disable it. You cannot have single global config or even single fixed project-level config for that.

bluwy:
I've recently worked on the project and agree with Divyansh here. Let's keep the current setup, but the funding change here can be introduced if you'd like to add that in another PR.

PR #4696, opened 2025-04-15
Fix example for loading custom font

pelevesque:
### Description

I believe the regex for the file needs to be tested against all asset files.

### Additional Context

If this is a correct fix, I can add the fix to all other languages before this pull request is merged.


brc-dd:
fix is correct
also probably that regex should be `/font-name\.[\w-]+\.woff2/`

pelevesque:
That is so funny. I just spent 15min scratching my head trying to figure out why one of my fonts was loading, but the other wasn't. It was because the hash contained a -. I was coming here to say that the regex also had to be fixed, and you beat me!

I'll fix this pull request with the regex fix, and also changes in all language files.

I have time to do it little later tonight.

PR #4712, opened 2025-04-22
fix: External link icon is not visible in doc content.

dvuckovic:
### Description

While the external link icon indicator is visible for the header links, the style definition is incomplete for doc content. This PR extends the rule to make it visible there too, so all external non-excepted links have a suitable indicator.

### Additional Context

Screenshot 2025-04-22 at 15 00 54


brc-dd:
You probably haven't set `themeConfig: { externalLinkIcon: true }` in your config.

dvuckovic:
@brc-dd You are absolutely right, didn't even realize it was configurable.

Sorry for the noise.

PR #4726, opened 2025-04-28
fix(deploy): update outDir path in VitePress config for GitLab Pages

robinsmith-source:
### Description

This pr fixes the `outDir` path for the site-config as well as for the gitlab pages deployment. The public has to be inside the project environment, with `../public` it's declared to be outside of the projects scope. 

### Additional Context

The wrongly documented `outDir` leads to failing GitLab Pages deployment, because no artifacts can be found at the requested source.


brc-dd:
It's based on the assumption that your vitepress root is `docs` (read the very first paragraph of deploy.md), in that case `../public` will be the root of your repo which is the intended path. You probably don't have your site in the repo root itself, so in that case you'll need to adjust the outDir accordingly.

PR #4747, opened 2025-05-14
fix: skip fields not supported by rolldown for rolldown-vite

sapphi-red:
### Description

Skip fields that are not supported by rolldown when it's running with rolldown-vite.

This PR will make ecosystem-ci pass with rolldown-vite.

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/node/build/bundle.ts
+++ b/src/node/build/bundle.ts
@@ -1,6 +1,7 @@
 import fs from 'fs-extra'
 import path from 'node:path'
 import { fileURLToPath } from 'node:url'
+import * as vite from 'vite'
 import {
   build,
   normalizePath,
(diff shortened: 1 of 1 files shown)


src/node/build/bundle.ts
```diff
@@ -98,9 +99,12 @@ export async function bundle(
...
         },
-        // important so that each page chunk and the index export things for each
-        // other
-        preserveEntrySignatures: 'allow-extension',
+        // @ts-ignore skip setting it for rolldown-vite since it doesn't support `preserveEntrySignatures` yet
+        ...(vite.rolldownVersion
+          ? undefined
+          : // important so that each page chunk and the index export things for each
+            // other
+            { preserveEntrySignatures: 'allow-extension' }),
```
sapphi-red, line 107: This part can be reverted, once https://github.com/rolldown/rolldown/issues/3500 lands.

brc-dd:
One issue is still there - https://github.com/vuejs/vitepress/blob/ab0e0cb5981ad2f849e8dab6a5b2b07f9502768f/src/node/plugin.ts#L59

`chunk.type` is undefined when using rolldown-vite. So lean chunk generation is not working properly. It is used like this:

https://github.com/vuejs/vitepress/blob/ab0e0cb5981ad2f849e8dab6a5b2b07f9502768f/src/node/plugin.ts#L319-L335

If I remove that chunk.type condition it works fine. Is this expected?

brc-dd:
Merged it for now. I believe the rolldown fix will be pulled automatically once rolldown-vite updates to it. Anyway it's not breaking anything major and the sites work fine even with the current rolldown-vite version.

PR #4748, opened 2025-05-14
feat: use `oxc-minify` instead of `transformWithEsbuild` when rolldown-vite is used

sapphi-red:
### Description

Built on top of #4747.

Replaced `transformWithEsbuild` with `oxc-minify` when rolldown-vite is used.

`transformWithEsbuild` still exists for rolldown-vite, but it requires users to install `esbuild` and it'll be deprecated in favor of `transformWithOxc` in the future. That said, `transformWithOxc` does not have minify feature built-in. I replaced it with `oxc-minify` instead. Currently users that wants to use vitepress with rolldown-vite has to install `oxc-minify`, but after Vite migrates to rolldown-vite, vitepress can make `oxc-minify` a normal dependency so that it will be installed automatically.

### Additional Context

---

> [!TIP]

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,6 +165,7 @@
     "minimist": "^1.2.8",
     "nanoid": "^5.1.5",
     "ora": "^8.2.0",
+    "oxc-minify": "^0.69.0",
     "p-map": "^7.0.3",
     "path-to-regexp": "^6.3.0",
     "picocolors": "^1.1.1",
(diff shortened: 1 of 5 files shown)


brc-dd:
This is good to merge, right? It was draft because #4747 wasn't merged?

sapphi-red:
Yes, it is good to merge!

PR #4749, opened 2025-05-14
 Add rel=edit attribute to "Edit this page on GitHub" link

capjamesg:
### Description

This PR adds the `rel="edit"` attribute to the "Edit this page on GitHub" edit link.

rel=edit lets a page indicate that the linked resource can be used to edit the page. It is defined at https://microformats.org/wiki/rel-edit. This can then be parsed by tools like the Universal Edit Button and custom bookmarklets to open the edit page corresponding with a website.

### Linked Issues

N/A

### Additional Context

N/A

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/components/VPDocFooter.vue
+++ b/src/client/theme-default/components/VPDocFooter.vue
@@ -30,7 +30,7 @@ const showFooter = computed(
 
     <div v-if="hasEditLink || hasLastUpdated" class="edit-info">
       <div v-if="hasEditLink" class="edit-link">
-        <VPLink class="edit-link-button" :href="editLink.url" :no-icon="true">
+        <VPLink class="edit-link-button" :href="editLink.url" :no-icon="true" :rel="edit">
           <span class="vpi-square-pen edit-link-icon" />
           {{ editLink.text }}
         </VPLink>


brc-dd:
It should be `rel` not `:rel`. But not sure whether we should add this. WHATWG has rejected this. It is not a well defined rel value. And that extension only has 24 downloads. It is some 15 year old obsolete extension that most likely doesn't even work in latest browsers. Also that wiki article is written by you and you just created the PRs to add this in like 10 other projects, so we don't have an unbiased view of who is even using the "custom bookmarklets" you mentioned.

capjamesg:
I would definitely like to see the rel value take off! The current spec is based off prior art form https://www.w3.org/Bugs/Public/show_bug.cgi?format=multiple&id=8735, where a blocker was that there was no documentation that referenced its use. If there is broader implementation interest, I plan to respond back to that thread with reference to its current state.

I edit a lot of documentation sites so I have felt the need for a standard way to find edit links for a while :D

If it's not of interest right now, feel free to close the PR.

brc-dd:
I'm fine with merging it. But I want to see some interest from others first and intended usage (probably some demo or something). Is there any real (working) extension or software that uses `rel=edit` to provide features?

PR #4758, opened 2025-05-23
Update custom-theme.md

PizzaConsole:
defineConfigWithTheme deprecated — use defineConfig instead

### Additional Context

---

> [!TIP]


brc-dd:
Ah, let's wait for a bit. It's deprecated in v2-alpha. But considering most of the people are still in v1 and it is just deprecated not removed, let's keep it for now. We can merge this around going beta.

bluwy (approved):
It seems ok to do this now since this is the docs and it can avoid mentioning deprecated APIs

PR #4768, opened 2025-05-30
style: adjust lists style

FelipePEduardo:
### Description

In the lists, specially in the ordered list, there is a small misalignment comparing number 1 to the others numbers
image

This will be fixed on this PR, but will also change the line style as you can see in the pictures
image

### Linked Issues

N/A

### Additional Context

N/A

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/styles/components/vp-doc.css
+++ b/src/client/theme-default/styles/components/vp-doc.css
@@ -138,7 +138,7 @@
 
 .vp-doc ul,
 .vp-doc ol {
-  padding-left: 1.25rem;
+  list-style-position: inside;
   margin: 16px 0;
 }
 


brc-dd:
Bit opinionated, but I prefer the first one. Aligning content should be the priority, not the markers. If you've 10 items, surely the 10th marker will be _more outside_ than others. tailwind typography plugin also renders lists like this:

image

PS: the difference isn't much visible in the tailwind screenshot because it's using my system font (SF Pro) which has almost same width for `1` and `2`.

Feel free to add custom CSS (https://vitepress.dev/guide/extending-default-theme#customizing-css) if that's bothering you, but we won't change it. More easily, you may override the font (https://vitepress.dev/guide/extending-default-theme#using-different-fonts) that will keep the overall look without that slight difference between alignment.

PR #4778, opened 2025-06-07
fix(docs): pnpm init command

jackkoskie:
### Description

In the Getting Started Page, the cli init command is listed as `npx vitepress init`. While for many things, pnpm can be a drop in replacement, for download and execute commands, it is not. `pnpm vitepress init` looks for a `vitepress` script in the current node project. The correct command would be `pnpm dlx vitepress init`

---

> [!TIP]

### DIFF
--- a/docs/en/guide/getting-started.md
+++ b/docs/en/guide/getting-started.md
@@ -72,7 +72,7 @@ $ npx vitepress init
 ```
 
 ```sh [pnpm]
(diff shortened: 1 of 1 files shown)


brc-dd:
That’s expected usage. Read the page from the top. Search for other similar PRs. If you use pnpx/pnpm dlx, you will still need to install vitepress manually later.

PR #4870, opened 2025-07-31
feat: add source param to the deadlink check fn

bojanrajh:
### Description

This change allows skipping the deadlink check based on the source file in addition to the destination link.

It also optimizes the flow when the deadlink check is deactivated and skips the check loop which should improve the performance a bit.

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -189,7 +189,7 @@ export async function createMarkdownToVueRenderFn(
           return ignore.test(url)
         }
         if (typeof ignore === 'function') {
-          return ignore(url)
+          return ignore(url, fileOrig)
         }
         return false
       })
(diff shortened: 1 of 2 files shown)


src/node/markdownToVue.ts
```diff
@@ -189,13 +189,14 @@ export async function createMarkdownToVueRenderFn(
...
-          return ignore(url)
+          return ignore(url, fileOrig)
         }
         return false
       })
     }
 
-    if (links) {
+    const { ignoreDeadLinks = true } = siteConfig
+    if (links && ignoreDeadLinks !== true) {
```
brc-dd, line 199: why is this added?
also destructuring here is bit confusing since ignoreDeadLinks defaults to false.

PR #4873, opened 2025-08-01
fix: add `rel="me"` to social links by default

andreynering:
### Description

This is a standard and many sites do this automatically. For example, GitHub adds this automatically for all your social links.

Mastodon also requires a `rel="me"` to be present to consider the link as verified.

* https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/me
* https://joinmastodon.org/verification

/preview


andreynering:
As a side-note, I've been loving VitePress! It's so much simpler to setup compared to other tools that I've used, and the design is very elegant while still being simple. 💯

PR #4885, opened 2025-08-05
refactor: avoid assigning to bundle object

TheAlexLichter:
Backporting https://github.com/vuejs/vitepress/pull/4647 to v1 (so it can be used with `rolldown-vite` without issues)

### DIFF
--- a/src/node/plugin.ts
+++ b/src/node/plugin.ts
@@ -334,36 +334,36 @@ export async function createVitePressPlugin(
       return null
     },
 
-    generateBundle(_options, bundle) {
-      if (ssr) {
-        this.emitFile({
-          type: 'asset',
-          fileName: 'package.json',
-          source: '{ "private": true, "type": "module" }'
-        })
-      } else {
-        // client build:
-        // for each .md entry chunk, adjust its name to its correct path.
-        for (const name in bundle) {
-          const chunk = bundle[name]
-          if (isPageChunk(chunk)) {
-            // record page -> hash relations
-            const hash = chunk.fileName.match(hashRE)![1]
-            pageToHashMap![chunk.name.toLowerCase()] = hash
+    generateBundle: {
+      order: ssr ? null : 'post',
+      handler(_options, bundle) {
+        if (ssr) {
+          this.emitFile({
+            type: 'asset',
+            fileName: 'package.json',
+            source: '{ "private": true, "type": "module" }'
+          })
+        } else {
+          // client build:
+          // for each .md entry chunk, adjust its name to its correct path.
+          for (const name in bundle) {
+            const chunk = bundle[name]
+            if (isPageChunk(chunk)) {
+              // record page -> hash relations
+              const hash = chunk.fileName.match(hashRE)![1]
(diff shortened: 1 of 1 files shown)


brc-dd:
I think other PRs like #4748 and https://github.com/vuejs/vitepress/commit/ed387e89d42a08c15a9f45c9c5e11c6750245490 will also need to be ported for it to work? Also, even then I don't think rolldown-vite had any version corresponding to vite 5 (that vp 1 uses).

brc-dd:
ah it's against v1, so pkg.pr.new and CI won't work

TheAlexLichter:
* #4748 is more of a perf improvement (but wouldn't hurt I assume :D)
* ed387e8 would need a port, yes.

> Also, even then I don't think rolldown-vite had any version corresponding to vite 5 (that vp 1 uses).

That's correct, though given little breaking changes, even using Vite 7-based rolldown-vite worked fine for a lot of VitePress pages.

Do you think it is worth doing the backport work or would you rather consider advising people to switch to v2 alpha instead (given they are up for unstable versions anyway when using rolldown-vite)?

brc-dd:
> would you rather consider advising people to switch to v2 alpha instead (given they are up for unstable versions anyway when using rolldown-vite)?

Yeah.. maybe recommending v2-alpha is fine for rolldown-vite. There aren't any planned breaking changes for v2 for now. I think we can make it beta too 👀 

> is more of a perf improvement (but wouldn't hurt I assume :D)

Ah, I thought rolldown-vite removed esbuild. Looks like it's still there. So, yeah not strictly necessary.

TheAlexLichter:
> Yeah.. maybe recommending v2-alpha is fine for rolldown-vite. There aren't any planned breaking changes for v2 for now. I think we can make it beta too 👀

Should we then instead add a log message to switch to v2 when using v1 and rolldown-vite is detected? 😋

> Ah, I thought rolldown-vite removed esbuild. Looks like it's still there. So, yeah not strictly necessary.

It did - you have to install it manually (similar to oxc-minify in v2 alpha rn)

brc-dd:
> Should we then instead add a log message to switch to v2 when using v1 and rolldown-vite is detected? 

We can, but I don't want to release a version just for this 😅 Better to add to readme or getting started maybe.

TheAlexLichter:
> > Should we then instead add a log message to switch to v2 when using v1 and rolldown-vite is detected?
> 
> We can, but I don't want to release a version just for this 😅 Better to add to readme or getting started maybe.

We should do that too but I doubt that people will necessarily check the readme when trying rolldown-vite 🙈 Most of the time, these are existing projects.
The change could be shipped in a `patch` version in theory!

brc-dd:
Hmm, okay. Can you create/update the PR? An `if (!isRestart && vite.rolldownVersion)` check can be added after this:

https://github.com/vuejs/vitepress/blob/ba9f50f829f8b56f71466e2af4aed9d6b4260e94/src/node/cli.ts#L51

TheAlexLichter:
Will create a separate PR! Thanks for the pointers ☺️

PR #4888, opened 2025-08-05
chore: error when trying `rolldown-vite` with Vitepress v1

TheAlexLichter:
This PR adds an error message when trying Vitepress on v1 with `rolldown-vite`, as both are incompatible. Instead, version two should be used.

Follow-up of #4885


brc-dd:
released

image

PR #4905, opened 2025-08-12
feat: use `@layer __vitepress_base` to wrap the styles in base.css

### LINKED ISSUE #4904: The styles in vitepress's base.css override the styles set by TailwindCSS v4
### Describe the bug

I'm integrating TailwindCSS v4 in Vitepress, and I found that the styles of one of my components were inexplicably not taking effect. After debugging, I found that the styles set by TailwindCSS were declared in `@layer`. Since the CSS layer mechanism is that anonymous layers will override named layers (if there is a conflict), the styles in [base.css](https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/styles/base.css) always have the highest priority (in my case) because it is not wrapped by a layer. According to the [CSS @layer rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer#description), the styles in `base.css` will be collected into the anonymous layer.

More context: https://github.com/nextui-vue/nextui-vue/issues/141#issuecomment-3179391270

### Reproduction

Check the styles of the [Button component](https://nextui-vue-docs-92hlaospb-hotdogc1017s-projects.vercel.app/en/components/button.html#usage), its background color is transparent because the `background-color` set by TailwindCSS is overwritten by VitePress's [base.css](https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/styles/base.css#L158)

### Expected behavior

The set style will not be overwritten

### System Info

```Text
System:
    OS: Linux 6.14 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (18) x64 Intel(R) Core(TM) Ultra 5 125H
    Memory: 13.92 GB / 30.85 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.14.0/bin/npm
    pnpm: 10.14.0 - ~/oss/nextui-vue/node_modules/.bin/pnpm
    bun: 1.2.7 - ~/.bun/bin/bun
  Browsers:
    Chrome: 139.0.7258.66
```

### Additional context

_No response_

### Validations

hotdogc1017:
### Linked Issues

fixes #4904

### Additional Context

none

---

> [!TIP]


innocenzi:
@brc-dd could this be merged please? This is essential for Tailwind CSS 4 to work. Would be appreciated. ♥️

innocenzi:
Could be great to setup https://pkg.pr.new/ for this kind of situation as well

brc-dd:
> Could be great to setup https://pkg.pr.new/ for this kind of situation as well

It's set up actually. Needs /publish comment by author currently. But the author was first time contributor so it still would've needed approval.

You can install the version on main from https://nightly.akryum.dev/vuejs/vitepress

innocenzi:
Ah, thank you—just did, and I confirm it works. Really appreciate it 🙏

PR #4910, opened 2025-08-15
Update asset-handling.md

mehrazmorshed:
### Description

Correction suggested for Assets Handling page of VitePress documentation.


bluwy:
The previous is also fine

PR #4931, opened 2025-09-01
feat(init): validation to make sure config directory is inside cwd

hyperz111:
This is a duplicate of #4876


hyperz111:
cc @brc-dd

PR #4940, opened 2025-09-10
fix(theme): skip link jumps to aside instead main content heading/anchor

userquin:
### Description

This PR just search for the first h1 heading with tabindex inside `#VPContent main` instead just inside `#VPContent`.

This PR also adds `position: fixed` to the skip link anchor, the skip to content anchor scrolls with the content.

~~The result will be similar to clicking on the first aside link if present without changing the route.~~

Since Skip to Content has an anchor to `#VPContent` there is no way to avoid changing the url.

### Additional Context

---

> [!TIP]


userquin:
don't merge yet, we're going to change a few things to simplify the logic

bluwy:
What's the status of this PR?

PR #4943, opened 2025-09-13
fix(a11y,client): improve focus handling and scrolling behavior in router

brc-dd:
Before (key strokes are not shown, but I'm pressing <kbd>tab</kbd> after clicking):

https://github.com/user-attachments/assets/749ef9e0-1c40-4592-82d1-d3290dbbd755

After:

https://github.com/user-attachments/assets/983c7163-d496-4b57-8dd4-9e7786ad6897

### DIFF
--- a/src/client/app/router.ts
+++ b/src/client/app/router.ts
@@ -287,8 +287,12 @@ export function scrollTo(hash: string, smooth = false, scrollPosition = 0) {
 
     // focus the target element for better accessibility
     target.focus({ preventScroll: true })
+
+    // return if focus worked
     if (document.activeElement === target) return
 
+    // element has tabindex already, likely not focusable
+    // because of some other reason, bail out
     if (target.hasAttribute('tabindex')) return
 
     const restoreTabindex = () => {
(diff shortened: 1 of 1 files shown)


brc-dd:
For skip link thing, it still needs #4940. But I'm thinking of going with `<main id="main-content">` and link using `#main-content` instead of manually finding the first heading. I'll update that PR and add main element with id to home too.

src/client/app/router.ts
```diff
@@ -261,35 +261,50 @@ export function scrollTo(hash: string, smooth = false, scrollPosition = 0) {
...
+    // target is not focusable, make it temporarily focusable
+    const tabindex = target.getAttribute('tabindex')
+    target.setAttribute('tabindex', '-1')
+
+    const restoreTabindex = () => {
+      if (tabindex == null) target.removeAttribute('tabindex')
+      else target.setAttribute('tabindex', tabindex)
+      target.removeEventListener('blur', restoreTabindex)
     }
+    target.addEventListener('blur', restoreTabindex)
```
userquin, line 301: check first if the target has tabindex, we can remove it with this logic ;)

src/client/app/router.ts
```diff
@@ -261,35 +261,50 @@ export function scrollTo(hash: string, smooth = false, scrollPosition = 0) {
...
+    : // only smooth scroll if distance is smaller than screen height
+      smooth && Math.abs(targetTop - window.scrollY) <= window.innerHeight
+      ? 'smooth'
+      : 'auto'
+
+  const scrollToTarget = () => {
+    window.scrollTo({ left: 0, top: targetTop, behavior })
+
+    // focus the target element for better accessibility
+    target.focus({ preventScroll: true })
```
userquin, line 289: move the tabindex logic before focusing the target element

userquin (approved):

PR #5002, opened 2025-10-23
fix(cli): gate ora spinner behind tty check

nicolas-goudry:
### Description

This PR gates the `ora` spinner behind a TTY check so VitePress behaves correctly in non-interactive environments (e.g., Nix builds).

Without this, the spinner remains enabled in non‑TTY contexts and suppress/overwrite error output, making build failures appear to "hang" indefinitely.

#### Why this is needed

- Animated terminal UIs assume an interactive TTY. In non‑TTY contexts, `ora`'s frame updates and cursor control can interfere with output and error reporting. This actually hides build errors and looks like a hang
- This change preserves the same behavior for interactive terminals, and prints clear, non-interactive output in non‑TTY environments

#### Minimal public reproduction (Nix + GitHub Actions)

- Repo: [ora-nix-issue-demo](https://github.com/nicolas-goudry/ora-nix-issue-demo)
- It uses a tiny script that rejects a promise after a short delay
- Two modes:
  - Problematic: spinner force enabled in a non‑TTY, error is hidden, looks like a hang
  - Fixed: spinner gated by TTY, error shows reliably
- Includes both a Nix expression and a GitHub Actions workflow to reproduce in non-interactive contexts

#### Impact

- No breaking changes for interactive use
- In non‑TTY environments, errors are visible and builds no longer appear to hang

### Linked Issues

N/A

### Additional Context

N/A

### DIFF
--- a/src/node/utils/task.ts
+++ b/src/node/utils/task.ts
@@ -1,10 +1,11 @@
+import process from 'node:process'
 import ora from 'ora'
 
 export const okMark = '\x1b[32m✓\x1b[0m'
 export const failMark = '\x1b[31m✗\x1b[0m'
 
 export async function task(taskName: string, task: () => Promise<void>) {
-  const spinner = ora({ discardStdin: false })
+  const spinner = ora({ discardStdin: false, isEnabled: !!process.stdin.isTTY })
   spinner.start(taskName + '...')
 
   try {


brc-dd:
Ora is disabled by default if stderr is non-TTY or it's in CI. It effectively uses this condition as default value of `isEnabled`:

```ts
  return Boolean(
    process.stderr && process.stderr.isTTY &&
    process.env.TERM !== 'dumb' &&
    !('CI' in process.env)
  )
```

In your example, are you sure it's not working with the default value? The condition in your example should be `isEnabled: withFix ? !!process.stdin.isTTY : undefined` (the default value is not `true`)

nicolas-goudry:
Thanks for your quick answer.

As per my tests, when VitePress runs inside a Nix build, it hangs until build timeout (IIRC last time I let the build run through the night and it timed out after ~5h).

If I refer to your code sample, when running inside a Nix build, the spinner is enabled since all conditions evaluate to `true`:

```
TERM: xterm-256color
CI: undefined
stdout is tty? true
stderr is tty? true
stdin is tty? undefined
```

The only fix I can think of, without updating the code, is to set the `CI` variable inside the Nix build (which I currently do).

I get that this fix is solving a very Nix-specific issue, so I would understand if wouldn't wish to proceed with it. I can live with having to set the `CI` variable (even empty) in my Nix build.

brc-dd:
I'm fine with adjusting the value of isEnabled. I'm more curious about why it hangs. Might be better to create an issue the ora repo too if it's hanging with just ora.

nicolas-goudry:
Well, I’m curious too, so I’ll open an issue in ora's repo and link it with this one.

I’ve been using Nix for quite some time now, building packages and such, and never seen something like that.

PR #5006, opened 2025-10-27
Add SEO impact description with `metaChunk` option.

TooBug:
### Description

As the doc shows, `metaChunk` options has impact on SEO, add descriptions of it into docs.

### Linked Issues

None.

### Additional Context

I am not sure about:

1. The content is proper or not.
2. What is the doc translation workflow, I can only handle English & Chinese version for now, but I think this change should also affect docs in other languages.


bluwy:
I haven't found this to be true in all my usages of VitePress, are you sure of this SEO impact?

TooBug:
@bluwy According to my site, it's true.

You can visit my post on X for more context (I am not advertising): https://x.com/TooooooBug/status/1982280572052357501

I can confirm the following things:

1. VitePress will produces text on the `.md` path, but in fact these paths not exists. What exist are the url after rendering by VitePress (.html or no appendix).
2. Google do crawl the `.md` paths, even they are not links, nor appeared in the HTML DOM elements. In other words, Google parses url-like things in the HTML document instead of just parses the rendered HTML or DOM.
3. The evidence is the "Pages" - "Soft 404" on Google Web Console. I see hundreds of "404" problems then, all point to `.md` paths. But when I turned the `metaChunk:true`, these problems disappeared (very slow, I didn't realize when, but they are all gone now).

bluwy:
Interesting, but this only impacts the Search Console if I understand correctly? It wouldn't impact your site's SEO or derank your site because of these 404s.

Anyways this option seems to have more upsides, even with its existing description about caching. It's a bit strange to add more benefits here and readers would see this and wonder why a (seemingly no-brainer) option is not enabled by default. 


Maybe @brc-dd knows something about this option? Should we turn it on by default?

TooBug:
I think the difficulty is here:

We don't really know how it impacts SEO, but many people say 404 errors affect the Google SEO performance.

Anyway, you know the full picture now.

brc-dd:
I think we can enable it by default now. Earlier it was experimental and we wanted to see if an extra request will cause any negative perf impact. But now I think it has been enabled for years on our docs and we haven't noticed any issue. It significantly reduces the build size (by like 100s of MBs in larger sites). Also, it's fingerprinted -> immutable cdn cache -> faster subsequent visits with smaller html payloads.

PR #5007, opened 2025-10-27
feat: support code block title

### LINKED ISSUE #3746: How to custom the title of code block like docusaurus?

I have to add a CodeBlock component like #1027 , When to support built-in title?

syntax like:

`yaml [prom.yml]`

or

`yaml title="prom.yml"`

in [Code blocks | Docusaurus](https://docusaurus.io/docs/markdown-features/code-blocks#code-title)

_No response_

### Additional context

_No response_

### Validations

bluwy:
### Description

Supports title in code blocks:

````md
```js [filename.js]
// ...
```
````

The title bar design is based on the code group tabs design. Previously, support for this syntax is indirectly added by https://github.com/yuyinws/vitepress-plugin-group-icons, so I had to make sure the changes I make can still somewhat work with the plugin, however I need to patch it slightly to prevent double title bars added by us and them.

cc @yuyinws the plugin might need a new option to disable its own `[title]` support if this PR is merged. I couldn't figure out an automatic way for the plugin to detect this feature.

### Linked Issues

fixes #3746

### Additional Context

I've skipped support for title bar support for [code snippets](https://vitepress.dev/guide/markdown#import-code-snippets) as there's no syntax to disable it at the moment. The snippets extensions [always adds a title](https://github.com/vuejs/vitepress/blob/be260fda6efc1d6c4b56219d7a17a19ab7a4ba76/src/node/markdown/plugins/snippet.ts#L32), causing it to be shown expectedly, e.g. the [CLI output docs here](https://vitepress.dev/guide/getting-started#setup-wizard) may show `init.ansi` as the title.

### CODE REVIEW

File: src/node/markdown/plugins/preWrapper.ts [hunk 1]
```diff
@@ -16,7 +17,12 @@ export function preWrapperPlugin(md: MarkdownItAsync, options: Options) {
     const [tokens, idx] = args
     const token = tokens[idx]
 
-    // remove title from info
+    // @ts-ignore
+    const isFromSnippet = !!token.src
```
bluwy: `src` here is [added by `snippet.ts`](https://github.com/vuejs/vitepress/blob/be260fda6efc1d6c4b56219d7a17a19ab7a4ba76/src/node/markdown/plugins/snippet.ts#L164-L165), however I think it's better if it's `token.meta.src` instead similar to how [`githubAlerts.ts` uses it](https://github.com/vuejs/vitepress/blob/be260fda6efc1d6c4b56219d7a17a19ab7a4ba76/src/node/markdown/plugins/githubAlerts.ts#L49-L52). But I didn't change it here as I don't want to accidentally break downstream if they use this internal info.

yuyinws:
What if we migrate [vitepress-plugin-group-icons](https://github.com/yuyinws/vitepress-plugin-group-icons) to Vitepress as a built-in feature? 

After this, all projects that used this plugin can safely remove it after upgrading to the latest version of Vitepress.

I'm happy to work for this. 

cc @brc-dd

bluwy:
I think in a way this PR is already a partial migrate to make the title work, and is better to implement this single feature first, then we can decide to implement icon support on top.

PR #5022, opened 2025-11-10
Fix calculation of targetTop in router.ts

Aryaman1792:
Fixes the calculation of targetTop in router.ts.
Previously, the computed scroll target could be incorrect under certain layout conditions (e.g., dynamic headers, nested containers, or offset inaccuracies). This update ensures the scroll position is determined consistently and more accurately.


bluwy:
Broken code

PR #5033, opened 2025-11-18
chore: Update oxc-minify@0.98

leaysgur:
### Description

Hi~ 👋🏻 
We released `oxc-minify@0.98` yesterday and it contains breaking change.

`minify()` is now asnyc API, we need to `await` it or use `minifySync()` instead.


brc-dd:
released

PR #5038, opened 2025-11-21
chore: move markdown-it to dependencies

Barbapapazes:
resolve https://github.com/vuejs/vitepress/discussions/5037

Hello 👋,

Having Markdown It inline makes it impossible to replace it, for example with Markdown Exit.

This PR moves Markdown It and related packages to dependencies to avoid having them inline.


meteorlxy:
The author of markdown-exit created a [hack](https://github.com/serkodev/markdown-exit/commit/3e0cfeb7926355f6896fcd4dd7afd797e7f6689f) to replace markdown-it.

For now we don't plan to move markdown-it to dependencies. Maybe we can consider migrating to markdown-exit directly after its stable release.

tats-u:
You have to add something like ThirdPartyNotice.txt for inlined packages with licenses like MIT.
You will no longer have to do it once you move them to `dependencies`.
In the first place why VitePress inlines every depending package? Users will not be able to update them to the latest patched versions with less bugs.

brc-dd:
will add an option to use markdown-exit later. in newer versions fence related plugins are also breaking if not awaited within each renderer. so it's more effort than it seems.

regarding the third party notices, yeah I'll add some plugin in build script to generate it.

PR #5040, opened 2025-11-22
docs: fix last-updated doc description

hiQianFan:
### Description

The description of the 'last updated' feature is inaccurate and can be misleading. When editLink is enabled, the default behavior should be that it appears on the left side.

### Additional Context

---

> [!TIP]


bluwy:
I think we should also update the english docs since it's the source of truth. Maybe we can change from:

> The update time of the last content will be displayed in the lower right corner of the page.

To

> The update time of the last content will be displayed at the bottom of the doc page.

So we don't have to specify left or right. Similar to how [edit link](https://vitepress.dev/reference/default-theme-edit-link) describes it, and we don't have to clarify for mobile view as well.

Can you update the english docs, then update the chinese docs based on the new text?

PR #5048, opened 2025-12-04
feat: use `minify` function exposed from Vite for Vite 8+

sapphi-red:
### Description

Vite 8+ exposes `minify` function so you don't need to install oxc-minify separately.

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/node/build/build.ts
+++ b/src/node/build/build.ts
@@ -30,8 +30,11 @@ export async function build(
 ) {
   const start = Date.now()
 
-  // @ts-ignore only exists for rolldown-vite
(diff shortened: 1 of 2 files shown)


brc-dd:
I was thinking to make vite 8 beta default 😅. I can push the changes.

Things seems to be working fine except for manualChunks, which I think should be migrated to advancedChunks, but I saw a recent issue on rolldown that the name and API is likely be to changed?

sapphi-red:
Even if we change the name & API, we'll keep the current one for a while, so it'd be fine to use it.

brc-dd:
added in main

PR #5055, opened 2025-12-06
docs(markdown): update mathjax3 install instructions in markdown guides

rockyxwall:
### Description
When you copy the old command `markdown-it-mathjax3^4`, it fails because the syntax is invalid.
This PR replaces that example with correct code-group install commands for all package managers and adds a note recommending `markdown-it-mathjax3@^4` due to breaking changes in newer versions.
All language versions of the markdown guide have been updated.

### Linked Issues

Not applicable, documentation correction.

### DIFF
--- a/docs/en/guide/markdown.md
+++ b/docs/en/guide/markdown.md
@@ -948,22 +948,9 @@ and include it like this:
 
 This is currently opt-in. To enable it, you need to install `markdown-it-mathjax3` and set `markdown.math` to `true` in your config file:
 
-::: code-group
-
-```sh [npm]
-$ npm add -D markdown-it-mathjax3@^4
+```sh
+npm add -D markdown-it-mathjax3@^4
 ```
-
-```sh [pnpm]
-$ pnpm add -D markdown-it-mathjax3@^4
-```
-
-:::
-
-::: tip NOTE
-Use `markdown-it-mathjax3@^4` instead of the latest version.  
-Newer releases introduce breaking changes that are not compatible with VitePress.
-:::
 
 ```ts [.vitepress/config.ts]
 export default {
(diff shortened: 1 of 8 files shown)


brc-dd:
For which package manager is it broken?

rockyxwall:
> For which package manager is it broken?

It is a syntax error. The package works fine, but as I was new and did not know that you have to write `mathjax3^4` as `mathjax3@^4`

When you run `npm add -D markdown-it-mathjax3^4`, you get this error:
`npm error code EINVALIDTAGNAME
npm error Invalid tag name "markdown-it-mathjax3^4" of package "markdown-it-mathjax3^4": Tags may not have any characters that encodeURIComponent encodes.
npm error A complete log of this run can be found in: C:\Users\admin\AppData\Local\npm-cache\_logs\2025-12-06T15_33_23_420Z-debug-0.log`

and for `pnpm add -D markdown-it-mathjax3^4`, you get this error:
`ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/markdown-it-mathjax34: Not Found - 404
This error happened while installing a direct dependency of B:\1Files\username\02_Codeing\01_Github\Lazy-Study-Doc
markdown-it-mathjax34 is not in the npm registry, or you have no permission to fetch it.
No authorization header was set for the request.`

So what I did was just remove the `^4` and install the latest `mathjax3` that is `v5` now. But with `v5` of `mathjax3`, VitePress does not work at all, as shown in the guide, you have to install `v4`. Just by adding a `@` before `^4` fixes the issue if it's an issue at all. It's just a syntax error.

brc-dd:
Ah, I see it was a typo earlier. @ should’ve been there. I still don’t think a note or code block is needed. The commands could simply be fixed. Also, we will be removing that plugin soon.

rockyxwall:
> Ah, I see it was a typo earlier. @ should’ve been there. I still don’t think a note or code block is needed. The commands could simply be fixed. Also, we will be removing that plugin soon.

Oh yes, you are right, I also had put English in all other languages... I just fixed the commands and removed everything else.

Issue #5069, opened 2025-12-29
请求将屏幕最小宽度的判断从 1280px 改成 1200 px


dogodo-cc:
### Describe the bug

如题：

现在很多便携屏都是 1200px 的横向像素，在使用便携屏幕访问 vitepress 构建的网站时，无法完整的看到全部内容 （文章摘要被隐藏了）。我跟踪了一下逻辑，发现最小宽度的判断是 1280，感觉略大了。

很早以前，文档类网站的宽度是 980px，所以我觉得 1200px 已经足够大了，我也在本地尝试调整过 1200px 的状态，是不局促的。

我暂时只能通过 patch-package 来修改。还是希望官方能够考虑将最小宽度调整到 1200px

哪怕开放成一个 “主题的设置项” 也是可以的。

感谢！！

### Reproduction

只要在 1200 宽度的屏幕浏览就可以复现了。

### Expected behavior

我的预期是 1200px 的屏幕宽度依然可以看到 文章摘要

### System Info

```Text
System:
    OS: macOS 15.6.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 78.03 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.4.0 - /Users/alan/Library/pnpm/node
    npm: 11.4.2 - /Users/alan/Library/pnpm/npm
    pnpm: 10.6.3 - /Users/alan/Library/pnpm/pnpm
  Browsers:
    Chrome: 143.0.7499.170
    Firefox: 144.0.2
    Safari: 18.6
```

### Validations

- [x] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [x] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [x] Read the [docs](https://vitepress.dev).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

dogodo-cc:
diff --git a/node_modules/vitepress/dist/client/theme-default/components/VPDoc.vue b/node_modules/vitepress/dist/client/theme-default/components/VPDoc.vue
index dd87ee9..b42dee7 100644
--- a/node_modules/vitepress/dist/client/theme-default/components/VPDoc.vue
+++ b/node_modules/vitepress/dist/client/theme-default/components/VPDoc.vue
@@ -90,7 +90,7 @@ const pageName = computed(() =>
   }
 }
 
[...]

PR #5088, opened 2026-01-12
fix: lower wide layout breakpoint to 1200px

seekskyworld:
## Issue
- https://github.com/vuejs/vitepress/issues/5069

## Summary
- shift default theme wide-layout breakpoint from 1280px to 1200px (aside/nav/local nav/backdrop)

## Motivation
- keep 1200px wide screens from prematurely hiding doc layout and navigation elements

## Validation
- pnpm dev:shared
- pnpm test:unit

### DIFF
--- a/src/client/theme-default/components/VPBackdrop.vue
+++ b/src/client/theme-default/components/VPBackdrop.vue
@@ -33,7 +33,7 @@ defineProps<{
   transition-duration: .25s;
 }
 
-@media (min-width: 1280px) {
+@media (min-width: 1200px) {
   .VPBackdrop {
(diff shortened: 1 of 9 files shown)


brc-dd:
Hmm, we should recommend some other way to do this. With 1200px other items in aside like ads and sponsor components look very condensed (not enough space from the window edge):

image

We can probably suggest using a simple vite plugin with transform to replace 1280px with 1200px everywhere. If many people need it we can provide it later as a configurable option.

seekskyworld:
OK, I have no objections.

PR #5114, opened 2026-02-06
feat(search): add moss search provider support

yatharthk2:
# Description
This pull request introduces Moss as a third search provider option, sitting alongside the existing Algolia and Local search implementations.

The integration enables a seamless, automated indexing workflow directly within the build pipeline. By using the new Moss indexer plugin, documentation content is automatically synced with the Moss platform during the client build, allowing for high-performance semantic search capabilities with minimal configuration.

# Most important changes:

- Moss Search Provider Integration: Added @inferedge/moss for client-side search and @moss-tools/md-indexer for build-time indexing.
- VitePress Build Pipeline: Registered mossIndexerPlugin in the VitePress build process. This plugin automatically syncs markdown content with Moss when enabled, ensuring the search index is always up to date after a deployment.
- Dynamic UI Updates: Updated VPNavBarSearch.vue to dynamically resolve the search provider. If moss is selected in the theme configuration, the new VPMossSearchBox.vue component is rendered with full keyboard shortcut support (Ctrl+K / Cmd+K).
- Conditional Logic: Implemented the __MOSS__ global constant to ensure that Moss-specific code and dependencies are only bundled when the provider is explicitly active, maintaining parity with how Algolia and Local search are handled.

# Reference

- [Moss Website](https://www.moss.dev/)
- [Moss Sample](https://github.com/usemoss/moss-samples)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2655,6 +2655,10 @@ packages:
   minimist@1.2.8:
     resolution: {integrity: sha512-2yyAR8qBkN3YuheJanUpWC5U3bb5osDywNB8RzDVlDwDHbocAJveqqj1u8+SVD7jkWT4yvsHCpWqqWqAxb0zCA==}
 
+  minipass@7.1.2:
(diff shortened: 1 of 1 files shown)


brc-dd:
Hi, we won't be adding any other third-party providers in the core. Please create and publish a separate package. You can use other plugins mentioned in the search docs as reference. We can link here once it's available. You probably will need to export a search component to register in `nav-bar-content-before` slot and the vite plugin you wrote here.

PR #5123, opened 2026-02-22
feat(theme): allow joining attr to external link

WizardsBowl:
### Description

I tried to add `rel="nofollow"` attribute to an external link and found it doesn't work.

My markdown file is like this:

```md
[LINK](https://external.link){rel=nofollow} // use markdown-it-attrs plugin
```

And this is the final html:

```html
<a href="https://external.link" rel="noreferrer" target="_blank">LINK</a>
```

It's `rel` attribute is an unexpected **noreferrer** but not **nofollow**.

That's because the `linkPlugin` always use `attrSet` and override the old value.

This PR add an option telling the `linkPlugin` to use `attrSet` or to use `attrJoin` for each attr.

```ts
linkPlugin(
  md,
  {
    target: '_blank',
    rel: { value: 'noreferrer', join: true },
    example1: 'will-override',
    example2: { value: 'will-override' },
    example3: { value: 'will-override', join: false },
    example4: { value: 'will-join-in', join: true },
    ...options.externalLinks
  },
  base,
  slugify
)

let externalLinks?: Record<string, string | ExternalLinkAttrValue>

export interface ExternalLinkAttrValue {
  value: string
  join?: boolean
}
```

The final html contains both **nofollow** and **noreferrer**:

```html
<a href="https://external.link" rel="nofollow noreferrer" target="_blank">LINK</a>
```
### Linked Issues

None.

### Additional Context

The markdown-it-attrs plugin is built-in with VitePress.

---

> [!TIP]

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -26,7 +26,7 @@ import mditCjkFriendly from 'markdown-it-cjk-friendly'
 import { full as emojiPlugin } from 'markdown-it-emoji'
 import type { BuiltinLanguage, BuiltinTheme, Highlighter } from 'shiki'
 import type { Logger } from 'vite'
-import type { Awaitable } from '../shared'
+import type { Awaitable, ExternalLinkAttrValue } from '../shared'
 import { containerPlugin, type ContainerOptions } from './plugins/containers'
 import { gitHubAlertsPlugin } from './plugins/githubAlerts'
 import { highlight as createHighlighter } from './plugins/highlight'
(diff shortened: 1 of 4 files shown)


bluwy:
This seems ok to me. Alternatively we could design it as `Record<string, string | (existing?: string) => string>` maybe so it's slightly more flexible. But either way is fine.

I thought about hardcoding the attributes that can be joinable, which is only `class` and `rel`, but I guess some people would prefer overriding over merging in some cases.

brc-dd:
Thanks for the PR! This isn't a common enough request to warrant a new API surface, so we've gone with a simpler fix instead (commit referenced above): the link plugin now just respects whatever attributes you've specified.

```md
[LINK](https://external.link){rel=nofollow}
```

now renders `rel="nofollow"` — the default is replaced, not joined. If you want both: `{rel="noreferrer nofollow"}`.

Issue #4471, opened 2025-01-07
allow multiple different color modes in addition to light|dark for themes


michaelcozzolino:
### Is your feature request related to a problem? Please describe.

right now the appearance has the following type:

```
appearance?: boolean | 'dark' | 'force-dark' | 'force-auto' | (Omit<UseDarkOptions, 'initialValue'> & {
        initialValue?: 'dark';
    });
```
the `Omit<UseDarkOptions, 'initialValue'>` allows to have only valueDark and valueLight, but it does not allow  me to have other modes different than dark/light such as custom mode created by the developer giving further color customisation to the theme: light|dark|luxury|business and so on.

### Describe the solution you'd like

i would like to have the possibility to use multiple modes in addition to dark/light. so that the appearance type becomes something like:

```
interface Appearance<T extends string> {
    appearance?: boolean | 'dark' | 'force-dark' | 'force-auto' | (Omit<UseColorModeOptions<T>, 'initialValue'> & {
        initialValue?: 'dark';
    });
}
```

where T is a union type of color modes 'light'|'dark'|'luxury'|'business'

### Additional context

if this is not possible please provide a solution in the case where:
1) I override the theme switcher with a custom theme switcher created by me containing different modes
2) I still have to keep the appearance to be not false in the config.ts else the switcher will disappear

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

PR #5129, opened 2026-02-26
Add force-light appearance

marshallku:
### Description

Add `force-light` mode for supporting light mode only.

### Linked Issues

#4471

### Additional Context

My design system supports light theme, but the color palette that I'm using supports both light and dark theme. \
So with `auto` appearance, it breaks my color palette.

---

> [!TIP]

### DIFF
--- a/docs/en/reference/site-config.md
+++ b/docs/en/reference/site-config.md
@@ -489,7 +489,7 @@ When set to `true`, the production app will be built in [MPA Mode](../guide/mpa-
 
 ### appearance
 
-- Type: `boolean | 'dark' | 'force-dark' | 'force-auto' | import('@vueuse/core').UseDarkOptions`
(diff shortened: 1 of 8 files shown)


brc-dd:
It doesn't fix that issue. Also, `appearance: false` is same as what you want in `force-light`.

marshallku:
Hi @brc-dd 
If I use `appearance: false`, it became automatical mode, which will be set to `dark` if user prefers dark theme.

brc-dd:
> If I use appearance: false, it became automatical mode, which will be set to dark if user prefers dark theme.

No it doesn't. This part will always resolve to false if appearance is false:

https://github.com/vuejs/vitepress/blob/44e2675889588b54dea771a0c82e2db64739364c/src/client/app/data.ts#L70-L81

And on build side, if appearance is false that script isn't injected:

https://github.com/vuejs/vitepress/blob/44e2675889588b54dea771a0c82e2db64739364c/src/node/config.ts#L351

Please share a reproducible example if `appearance: false` isn't working for you.

PR #5158, opened 2026-03-21
fix: keep translation links in the current tab

lilianakatrina684-a11y:
## Summary
- let translation links opt out of external-link handling in `VPLink`
- use that override for navbar and mobile translation menus so localized docs links stay in the current tab
- add unit coverage for the new external-link override logic


brc-dd:
update it here too - https://github.com/vuejs/vitepress/blob/4fc1db83d7c7af53a2dd6f6ec9686fce2ba7df7c/src/client/theme-default/components/VPNavBarExtra.vue#L34

(toggle copilot's low confidence comment for details)

PR #5176, opened 2026-04-08
feat: allow VPContent to use custom components

shigma:
### Description

Currently, `VPContent` hardcodes the use of `VPPage`, `VPHome`, and `VPDoc` for the built-in `page`, `home`, and `doc` layouts. Custom layouts only work via `<component :is>` for non-built-in layout names. This means there is no way to fully override the built-in layout components from a custom theme.

This PR uses `resolveComponent` to check whether the user has registered a global component matching the layout name (`page`, `home`, or `doc`). If a matching component is found, VPContent falls through to `<component :is>` instead of using the built-in component. If no override is registered, the behavior is exactly the same as before.

### Linked Issues

N/A

### Additional Context

N/A

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/components/VPContent.vue
+++ b/src/client/theme-default/components/VPContent.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { resolveComponent } from 'vue'
 import NotFound from '../NotFound.vue'
 import { useData } from '../composables/data'
 import { useLayout } from '../composables/layout'
@@ -8,6 +9,13 @@ import VPPage from './VPPage.vue'
 
 const { page, frontmatter } = useData()
 const { isHome, hasSidebar } = useLayout()
+
+function resolveLayout(name: string) {
+  const component = resolveComponent(name)
+  // resolveComponent returns the component if registered,
+  // or the name string if not found
+  return typeof component === 'string' ? null : component
+}
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


src/client/theme-default/components/VPContent.vue
```diff
@@ -8,6 +9,11 @@ import VPPage from './VPPage.vue'
 
 const { page, frontmatter } = useData()
 const { isHome, hasSidebar } = useLayout()
+
+function resolveLayout(name: string) {
+  const component = resolveComponent(name)
+  return typeof component === 'string' ? null : component
```
brc-dd, line 15: could've just returned true/false I guess for this use case

edison1105:
@brc-dd @shigma 
It is not appropriate to use `resolveComponent` to check whether a component exists, as it will trigger a warning in DEV if the component does not exist. If this check is required, it is recommended to use `resolveDynamicComponent` instead, which can eliminate the warning. Additionally, both helpers are private. 
I'm wondering if there is a more reasonable approach to explicitly use a user-provided layout.

image

brc-dd:
Hmm, yeah hard to do this without resolveComponent/resolveDynamicComponent. Other alternative I see is doing something like this https://github.com/vueform/vueform/blob/main/src/utils/isComponentRegistered.js#L3, but that will need getCurrentInstance 🫠

We can go back to not supporting the default theme names it if there are issues with the vapor implementation.

shigma:
`resolveComponent` is actually a public API, not private. You can see it in the [official docs](https://vuejs.org/api/render-function.html#resolvecomponent).

PR #5179, opened 2026-04-14
feat: vapor mode

edison1105:
### Description

support vapor mode for the default theme.


src/client/theme-default/components/VPTeamPageSection.vue
```diff
@@ -1,15 +1,19 @@
+<script lang="ts" setup vapor>
+import {useSlots} from 'vue'
+const slots = useSlots()
+</script>
 <template>
   <section class="VPTeamPageSection">
     <div class="title">
       <div class="title-line" />
-      <h2 v-if="$slots.title" class="title-text">
```
edison1105: Vue compile doesn't inject `$slot`, `$attrs` for template-only component in Vapor mode. So it is necessary to define `$slots`.

src/node/markdownToVue.ts
```diff
@@ -277,6 +278,9 @@ function injectPageDataCode(tags: string[], data: PageData) {
   })
 
   const isUsingTS = tags.findIndex((tag) => scriptLangTsRE.test(tag)) > -1
+  const defaultExportCode = `\nexport default {name:${JSON.stringify(
+    data.relativePath
+  )}${vapor ? `,__vapor:true` : ``}}`
```
edison1105, line 285: process the markdown file into a Vapor component.

src/client/theme-default/components/VPNavBarSearch.vue
```diff
@@ -176,7 +188,7 @@ function isEditingContent(event: KeyboardEvent): boolean {
         v-if="loaded"
         :algolia-options
         :open-request
-        @vue:beforeMount="actuallyLoaded = true"
+        @beforeMount="actuallyLoaded = true"
```
edison1105, line 191: vnode hooks don't work in Vapor mode

src/client/app/index.ts
```diff
@@ -112,9 +147,21 @@ export async function createApp() {
 }
 
 function newApp(): App {
-  return import.meta.env.PROD
-    ? createSSRApp(VitePressApp)
-    : createClientApp(VitePressApp)
+  if (import.meta.env.SSR) {
+    return createSSRApp(VitePressApp)
```
edison1105, line 151: Vapor mode only works in the browser, so it should always use `createSSRApp` in server render.

brc-dd:
are there any more changes needed in the PR? if not we should just resolve conflict + bump to rc.

PR #5181, opened 2026-04-16
Make markdown guide clearer and more natural

slavbar:
### Description

Improved wording in the markdown guide to make it sound more natural.
Replaced informal phrases like "like this" with clearer alternatives.

### DIFF
--- a/docs/en/guide/markdown.md
+++ b/docs/en/guide/markdown.md
@@ -281,7 +281,7 @@ Wraps in a `<div class="vp-raw">`
   }
   ```
 
-  You can pass its options like this:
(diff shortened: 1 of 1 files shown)


skirtles-code:
I think the current usage of 'like this' is fine. While I can see that it might be considered a little informal, I think it's at the appropriate level of formality for this kind of documentation. Replacing 'like this' with more formal alternatives makes the wording feel less natural to me and I don't think it changes the clarity.

*(I'm not a maintainer for VitePress, just my opinion as a VitePress user)*

bluwy:
I think it’s fine too

PR #5238, opened 2026-06-27
chore: format dependency bump changes

cookesan:
### Summary
Format files touched by the dependency bump so the repo-wide Prettier check passes.

### Testing
- `pnpm check`


brc-dd:
ah nope reverted this. reinstall your deps locally. newer prettier formats them like that.

PR #5239, opened 2026-06-27
feat: allow custom i18n routing

### LINKED ISSUE #3530: themeConfig.true18nRouting should accept a function

The property themeConfig.true18nRouting is complicated, a boolean value is not sufficient for general usage case.

Please allow users to pass a function for the property themeConfig.true18nRouting, so that  they can decide how 18nRouting is done.

_No response_

### Additional context

_No response_

### Validations

cookesan:
### Summary
Add function support for `themeConfig.i18nRouting` so sites can build custom locale links.
Pass the current VitePress data, current hash, and target locale key to the function.
Move `VitePressData` to shared declarations so default-theme types can reference it before `dist` exists.
Document the signature and cover default, disabled, clean URL, and custom routing paths.
Refresh formatting for files affected by the dependency bump with the current lockfile formatter.

### Testing
- `pnpm vitest run -r __tests__/unit __tests__/unit/client/theme-default/composables/langs.test.ts`
- `pnpm check`
- `pnpm docs:build && pnpm docs:lunaria:build`

Closes #3530

### DIFF
--- a/__tests__/unit/client/theme-default/composables/langs.test.ts
+++ b/__tests__/unit/client/theme-default/composables/langs.test.ts
@@ -1,0 +1,69 @@
+import { ref } from 'vue'
+import type { VitePressData } from 'vitepress'
+import type { DefaultTheme } from 'vitepress/theme'
+import { resolveLocaleLink } from 'client/theme-default/composables/langs'
+
+function createData(
+  themeConfig: DefaultTheme.Config,
+  relativePath = 'guide/getting-started.md',
+  cleanUrls = false,
+  hash = '#install'
+) {
+  return {
+    site: ref({
+      cleanUrls,
+      locales: {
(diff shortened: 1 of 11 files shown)


brc-dd:
import `VitePressData` from `src/client/app/data.ts` instead of dist. dist is not there before building.

cookesan:
Thanks, updated. `VitePressData` now lives in the shared public declarations, and `app/data` re-exports it, so `DefaultTheme.I18nRouting` can reference it without going through `dist`.

brc-dd:
Thanks. In future, adjust the gate to `rm -rf dist && pnpm check` maybe (that's more accurate to what the CI does).

cookesan:
Good call, thanks. I'll use that check for VitePress changes going forward.

PR #5247, opened 2026-06-28
fix: stop hidden copy button blocking code text

### LINKED ISSUE #5120: The copy button obstructs the text selection and code copying within the code blocks in mobile view.
### Describe the bug

When trying to select and copy text, the copy button positioned above the text prevents full selection.

### Reproduction

Please watch this video for a detailed look.

https://github.com/user-attachments/assets/b6a5914f-9d82-4b0c-a4e0-c97c67f44afd

### Expected behavior

Perhaps the copy button should be positioned elsewhere in the mobile view.

### System Info

iOS26

### Additional context

_No response_

### Validations

cookesan:
### Description

Keeps code-block copy buttons from covering selectable code text on narrow screens by reserving a small mobile action row above the code content. The copy button still becomes usable on hover or focus, and desktop code blocks keep the existing layout.

### Linked Issues

Fixes #5120

### Additional Context

Checks:
- `pnpm -F=tests-e2e test markdown-extensions/markdown-extensions.test.ts`
- `pnpm check`
- `pnpm docs:build`

### DIFF
--- a/__tests__/e2e/markdown-extensions/index.md
+++ b/__tests__/e2e/markdown-extensions/index.md
@@ -169,7 +169,7 @@ export default config
 
 <<< @/markdown-extensions/foo.md
 
-<<< @/markdown-extensions/foo.md#snippet{1 ts:line-numbers} [snippet with region]
+<<< @/markdown-extensions/foo.md#snippet{1 ts:line-numbers} [snippet with region and a long mobile tab label]
 
 :::
 
--- a/__tests__/e2e/markdown-extensions/markdown-extensions.test.ts
+++ b/__tests__/e2e/markdown-extensions/markdown-extensions.test.ts
@@ -7,6 +7,30 @@ const getClassList = async (locator: Locator) => {
 
 const trim = (str?: string | null) => str?.replace(/\u200B/g, '').trim()
 
+const expectMobileCopyButtonOutsideCode = async (
+  block: Locator,
+  limit?: Locator
+) => {
+  const button = block.locator('> button.copy')
+  const firstLine = block.locator('code > span').first()
+  await block.scrollIntoViewIfNeeded()
+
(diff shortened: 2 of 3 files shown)


brc-dd:
Doesn't really fix the original issue. `.vp-doc [class*='language-']:hover` will match the case described in the issue and it will gain pointer-events even with the changes. I don't think there is any good fix for this other than to move the copy button outside.

cookesan:
Thanks, that makes sense. I pushed 6fa626c to keep the button above the selectable code content on narrow screens instead of relying on the hidden state. The regression now checks the hover-active case too.

Local checks:
- `pnpm -F=tests-e2e test markdown-extensions/markdown-extensions.test.ts`
- `pnpm check`
- `pnpm docs:build`

brc-dd:
hmm, yeah it works, but not really liking the extra space on top:

image<br />

Maybe we should position the button outside the code block (like outside that background), as a regular button in the mobile view (hover is bad from the UX perspective on mobiles anyway)

image<br />

Need to check how it works with code groups / titles though, especially which are longer (where a scrollbar is rendered for tabs). Don't think a padding-right can be decided for tabs without adjusting the structure. Maybe we can have better control when the title bar plugin is a part of the core.

cookesan:
Thanks, agreed. I pushed ca9c4102 to move the mobile copy button out of the code block background instead of adding top padding inside `pre`. The mobile rule now makes the button visible as a regular control while keeping the desktop hover behavior unchanged.

I also expanded the e2e coverage for:
- plain code blocks
- code inside custom containers
- code groups with a long scrollable tab row

Checks:
- `pnpm -F=tests-e2e test markdown-extensions/markdown-extensions.test.ts`
- `pnpm check`
- `pnpm docs:build`

PR #1844, opened 2023-01-26
fix(theme): avoid layout shift caused by scrollbar

zqianem:
fixes #1054

Same as #1568 which was mistakenly closed by me.

PR #5310, opened 2026-07-13
fix: set scrollbar-gutter stable

### LINKED ISSUE #1054: Reduce layout shifts with classic scrollbars

### LINKED ISSUE #4884: Layout shift when search modal toggle
When use Search in VitePress will cause layout shift
VitePress will lock the body scroll when search it's good behavior，but when scrollbar fade away it cause layout shift both localsearch and aligo doc search
https://github.com/user-attachments/assets/6887143c-3d88-46cb-a0ff-93e42c5d92c7

bluwy:
### Description

Use scrollbar-gutter to always preserve a space for scrollbars, even if it exists or not. This addresses the layout shift when navigating between pages that have scrollbars or not, and when opening/closing the search modal

### Linked Issues

fixes #4884 
fixes #1054
alternative to #1844 (that PR makes the scrollbar overlay instead of taking up space) (if this is merged, closes #1844)

### Additional Context

---

> [!TIP]


brc-dd:
One issue with scrollbar-gutter is that the browser doesn't paint things behind the gutter. So, the navbar/footer/gradient will appear to be cut near the right when the gutter is empty. Hold off merging this for now, I'll take a look to see if there is some better way.

bluwy:
In most cases it's not really obvious or visible, so I think it's fine. The worst case seems to be for Safari which forces an ugly theme-color background for the scrollbar (can't seem to workaround this):
image

I think it's still fine for Safari though, for those who use it. Otherwise the only other option is with #1844 which uses `width: 100vw` and `overflow-x: hidden`, which we might need to take special care of handling in a few places. Too bad browsers don't have a supported way to support gutters without taking up space like `overflow-y: overlay`.

PR #5311, opened 2026-07-13
feat: auto-add width/height to local images to avoid layout shift

### LINKED ISSUE #4853: Add `height` / `width` attribute to images to avoid layout shifts

When using `alt`, `height` / `width` attributes are not set and browsers cannot avoid layout shifts caused by the image.

For example, when clicking the `non-inherit` badge on the right side of each header on https://deploy-preview-20471--vite-docs-main.netlify.app/config/dep-optimization-options.html, the image above the link is shown in the viewport after the image is loaded.

Vitepress to generate `height` / `width` attributes for the images.

Add it manually by writing `alt{height=100 width=200}` or `alt{style="aspect-ratio:2; width:100%"}`

### Additional context

_No response_

### Validations

### LINKED ISSUE #5079: Anchor links are not resolving to the correct scroll position
### Describe the bug

If Markdown contains images without explicit width and height, the browser cannot determine the image dimensions until they are loaded or retrieved from cache. Consequently, anchor links may point to the wrong position on the page, sometimes resulting in the target header being hidden or misaligned.

Related issue in Mergin Maps docs:

https://github.com/MerginMaps/docs/issues/656

### Reproduction

Open page with images
Save anchor link
clean image cache in a browser
open link in new tab

### Expected behavior

anchor links works on pages with images.

### System Info

```Text
System:
    OS: Linux 6.16 Pop!_OS 22.04 LTS
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
    Memory: 5.61 GB / 31.05 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.20.4 - /home/marcelkocisek/.nodenv/versions/18/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 10.7.0 - /home/<>/.nodenv/versions/18/bin/npm
  Browsers:
    Firefox: 143.0.3
    Firefox Developer Edition: 143.0.3
  npmPackages:
    vitepress: ^1.6.3 => 1.6.3
```

### Additional context

Is there any workaround or script to prevent mentioned behavior?

### Validations

bluwy:
### Description

Adds a custom markdown-it `imageSizePlugin` to automatically add width and height to local & public images. Does not handle remote images because of fetching and caching complexity.

This uses https://www.npmjs.com/package/image-size to compute the size, it's however archived (due to author's LLM reasons). I haven't found a good alternative that supports a wide-range of file formats and is simple (less deps), e.g. https://www.npmjs.com/package/probe-image-size and https://www.npmjs.com/package/image-dimensions.

I did not use `markdown-it-image-size` as I don't want to bring in the dependencies for remote image support. However, maybe we're fine with it?

### Linked Issues

fixes #4853 
maybe also fixes #5079 (because it handles if only one width or height is set by `markdown-it-attrs`)

### Additional Context

---

> [!TIP]


brc-dd:
image<br />

^ feature comparison table against browser renderable formats, but yeah I think image-size is fine to use. (also bmp, ico, cur, tiff are discouraged to use by mdn)

bluwy:
Interesting, then I suppose we can use `image-dimensions` as an alternative since we don't have to handle the discouraged formats. Only missing SVG support in that case https://github.com/sindresorhus/image-dimensions/issues/9

brc-dd:
Hmm, I think either option is fine. I checked the code, and they both seem to use almost the same detection logic. I don't have a strong preference, although SVG support could be a deal-breaker for some users.

Let's stick with `image-size` for now. The code looks good to me. I'll add a test case for encoded URLs. ~~I expect it to work since the existing image plugin already handles that part.~~ Ah, actually, no, it gets added *after* the image plugin, so its wrapper runs first at render time.

Also, I don't think the comment `custom plugins after markdown-it-attrs (to respect user-defined attributes)` is accurate. Even if you register it before `markdown-it-attrs`, it should behave the same because `markdown-it-attrs` operates on the parser, not the renderer.

Oh, and it also needs to respect `vite.publicDir`.

bluwy:
Ah interesting, I thought I tested the order before in the test and it mattered. Otherwise would you prefer putting the size handling inside the image plugin instead? Initially I had that before I refactored it out.

brc-dd:
I've a local version, I'll push it.

>  Otherwise would you prefer putting the size handling inside the image plugin instead? Initially I had that before I refactored it out.

Yeah lol I moved it inside too 😅

brc-dd:
It has some unrelated changes (around passing down publicDir), and I'll clean up createMarkdownRenderer usage in a follow up PR as it currently returns a singleton instance and the arguments are mostly useless within the same runtime context and I think it should just accept the siteConfig if it even needs something.

bluwy:
Nice! This looks great to me. Thanks for refactoring it.

brc-dd:
#5079 is a bit different from this and is specific to their use case. I think we can close it. For context, they wanted to use `max-height` with `width: auto` so that smaller images would scale up until they reached either the container width or a specified height. However, that interferes with Chromium's layout calculations and doesn't prevent layout shift. I think the current implementation is fine for general use.

PR #5312, opened 2026-07-13
fix: preserve url params when switching languages

### LINKED ISSUE #4772: Preserve URL params when switching languages
Hi! Can't find any information on whether it's possible to preserve the current URL parameters when switching languages using the built-in Internationalization feature.

Example:  
We are on the page
`http://localhost:5173/en/demo?debounce=150`. The current language is `en`.  
Then we use the nav language switcher to select the language `es`. By default, the URL is rewritten to
`http://localhost:5173/es/demo`,  
but we would expect to keep the URL parameters, so the desired result would be:  
`http://localhost:5173/es/demo?debounce=150`.

If this isn't supported by default, what would be the recommended approach to "tweak" the language switcher in the nav bar so it behaves as described?

bluwy:
### Description

Take account of url params (search) similar to the existing hash handling.

### Linked Issues

fix #4772

### Additional Context

NOTE: I notice the current hash references from `useData`. The same info can also be taken from `useRouter` or `useRoute`. It's a be confusing to me why route data exists on `useData`, is this correct? Seems to be added in https://github.com/vuejs/vitepress/pull/3768

Otherwise maybe we should remove the `hash` from `useData` as a breaking change.

---

> [!TIP]

### DIFF
--- a/__tests__/unit/client/theme-default/composables/langs.test.ts
+++ b/__tests__/unit/client/theme-default/composables/langs.test.ts
@@ -1,5 +1,5 @@
 import { resolveLocaleLink } from 'client/theme-default/composables/langs'
-import type { VitePressData } from 'vitepress'
+import type { Route, VitePressData } from 'vitepress'
 import type { DefaultTheme } from 'vitepress/theme'
 import { ref } from 'vue'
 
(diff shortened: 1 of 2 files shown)


brc-dd:
> Otherwise maybe we should remove the hash from useData as a breaking change.

Yeah, I think we should remove that. query and hash weren't present in the route object earlier (when #3768 was added). They were added in #4511.

PR #5313, opened 2026-07-13
feat(build): native LLM-friendly output (llms.txt)

### LINKED ISSUE #4590: Support llms.txt generation

VitePress lacks support for the emerging [`llms.txt`](https://llmstxt.org/) "standard", which provides markdown content in a format optimized for LLM consumption. Documentation sites need a way to make their content more easily digestible by AI assistants for improved accuracy and context preservation.

Implement native `llms.txt` generation in VitePress:
- Auto-generate LLM-optimized versions of documentation
- Support structured content formatting for better AI comprehension
- Provide configuration options to customize output format

Manual content transformation or post-processing scripts, but these create maintenance overhead and version inconsistencies.

### Additional context

The `llms.txt` standard focuses on making documentation more accessible to AI systems through optimized formatting and structure. Supporting this would enhance how VitePress documentation is consumed by AI assistants, resulting in more accurate responses when users query documentation content through LLMs.

### Validations

posva:
I use https://github.com/okineadev/vitepress-plugin-llms in most of my projects but it takes too much time to generate the files and I think this is because of how it works and probably to enable further, more powerful integrations through remark.
This PR aims to cover what I use in Vue Router, Pinia, Pinia Colada, etc while staying negligible (100ms in this repo docs to the total) in build time and not adding any new dependencies.

Closes #4590

## What's included

- `llms: boolean | LlmsOptions` config option (`enabled`, `hostname`, `title`, `description`, `ignoreFiles`)
- `index.md` is ignored automatically
- `llms.txt` index following the sidebar order, `llms-full.txt` bundle, and a per-page `.md` next to each `.html` (includes expanded, rewrites applied)
- `<llm-only>` / `<llm-exclude>` tags to target content at LLMs or humans only
- Docs page under Experimental, and vitepress.dev now dogfoods the feature (drops its `vitepress-plugin-llms` dependency)

## Limitations (intentionally not covered)

- Only the root locale is emitted — translated locales are skipped
- Pages generated from dynamic routes (`[param].md`) are skipped
- `<<<` code snippet imports are not expanded in the markdown output (the directive is left as-is) because the replacement is not exposed where needed
- Image references keep their source paths; they are not rewritten to hashed build assets
- The `llms.txt` TOC only understands the default theme's sidebar config
- `<llm-only>` / `<llm-exclude>` are processed with string-level regexes, so literal occurrences inside code fences are also affected

@okineadev, are you [still interested](https://github.com/vuejs/vitepress/issues/4590#issuecomment-2796391242) in maintaining something like this?

PS: @brc-dd this code is fully generated because I needed to see if it was possible to cover what I needed personally. I haven't (at least yet) reviewed the generated code in depth, because **I don't know if you want to go forward with this feature or not**.

### DIFF
--- a/__tests__/e2e/.vitepress/config.ts
+++ b/__tests__/e2e/.vitepress/config.ts
@@ -154,6 +154,7 @@ const sidebar: DefaultTheme.Config['sidebar'] = {
 export default defineConfig({
   title: 'Example',
   description: 'An example app using VitePress.',
+  llms: true,
   markdown: {
     image: {
       lazyLoading: true
--- a/__tests__/unit/node/build/generateLlmsTxt.test.ts
+++ b/__tests__/unit/node/build/generateLlmsTxt.test.ts
@@ -2,7 +2,10 @@ import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
 import type { Logger } from 'vite'
-import { generateLlmsTxt } from 'node/build/generateLlmsTxt'
+import {
+  collectLlmsSource,
+  generateLlmsTxt
+} from 'node/build/generateLlmsTxt'
 import type { SiteConfig } from 'node/siteConfig'
 
 const logger = {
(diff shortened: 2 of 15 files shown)


posva:
@posva-Ghostty‒ tacobook ❐ 0 ● 12 nvim-2026-07-13-14 45 33@2x


hmm, actually the difference is not _that_ big: in vitepress docs (big docs) it's a ~10% improvement. In the screenshot the second run is with llms: true while the first is only with the plugin. This means I had the wrong perception of the speed impact of the plugin (I really thought it was impacting speed much more). I still think

brc-dd:
> actually the difference is not that big

The approach is wrong. If you want perf gains you need to collect/emit somewhere in between the markdown to vue pipeline. This is currently re-rendering stuff again inside generateLlmsTxt. Also, check the difference with the preview version on pinia/router docs once locally.

posva:
In router, locally, the improvement is too small ~200ms.
But I'm glad the approach is wrong so it can be improved. I will give it a try

Inside netlify still not enough difference:
|llms-plugin|native|
|---|---|
|@posva-Microsoft Edge‒Deploy details  Deploys  vue-router  Netlify-2026-07-13-16 32 59@2x|@posva-Microsoft Edge‒Deploy details  Deploys  vue-router  Netlify-2026-07-13-16 32 30@2x|

posva:
Well, my try was not successful 😆 , the perf is the same but it's more correct (it includes expansions from the build pipeline)

okineadev:
> @okineadev, are you [still interested](https://github.com/vuejs/vitepress/issues/4590#issuecomment-2796391242) in maintaining something like this?

Yes

okineadev:
I thought we can just move all my plugin code here for now so it basically works, but do as you wish

okineadev:
> I thought we can just move all my plugin code here for now so it basically works, but do as you wish

If it is possible to port all this to vitepress, I think I will still leave my plugin for maintenance, I think to maintain compatibility with older versions of vitepress

posva:
> I thought we can just move all my plugin code here for now so it basically works, but do as you wish

It's a good chance to reduce the API surface and take a completely different approach, so copying wouldn't be that interesting compared to just having the plugin as an external package 😄

okineadev:
Will there be any updates on this?

posva:
I don't know. Maybe not yet. I don't think this is needed for vitepress 2.0 and I think it's important to keep the initial api surface small (like here) instead of large (like in vitepress-plugin-llms). But that's all @brc-dd 's call

PR #5230, opened 2026-06-27
fix(build): show dead link line numbers

cookesan:
### Description

Fixes #2847.

Dead-link warnings now include the Markdown source line by preserving line metadata during link collection and printing it as `file:line`.

This keeps the existing `env.links` array intact and stores line numbers separately, so existing consumers that read collected links keep the same data format.

### Validation

- `pnpm vitest run -r __tests__/unit __tests__/unit/node/markdownToVue.test.ts __tests__/unit/node/markdown/plugins/link.test.ts`
- `pnpm check`
- Manual CLI check with a missing link on line 5: `node bin/vitepress.js build /tmp/vp-dead-link-check`; output included `index.md:5`.

PR #5316, opened 2026-07-14
fix: report raw url for dead links

### LINKED ISSUE #4992: [Feature Request] 报错信息中原封不动输出url
### 背景
在构建中文文档站点时，`Found dead link /%E4%BB%BB%E5%8A%A1/%E6%95%85%E4%BA%8B in file 任务\index.md` 这类信息可读性很差。
建议：
1. 对 `%E4%B8%AD%E6%96%87` URL 自动 decode；
2. 提供 `--locale zh` 或 `localizeLog` 选项，让构建日志可本地化。

### 理由
- 目前大量中文项目基于 VitePress（如 VitePress Theme Hope、NaiveUI 文档等）；
- 中文路径在 `dead link` 检查时容易输出乱码；
- decode 一下其实就能极大提升可读性。

bluwy:
### Description

With a setup like `中文.md` with 

```md
[test](./报错信息.md#test)
```

Before:

```
(!) Found dead link ./%E6%8A%A5%E9%94%99%E4%BF%A1%E6%81%AF in file /Users/bjorn/Work/oss/vitepress/docs/en/中文.md:1
```

After:

```bash
(!) Found dead link ./报错信息.md#test in file /Users/bjorn/Work/oss/vitepress/docs/en/中文.md:1
```

This PR renames the `linkLines` property to `linkMetadatas`. `linkLines` was added in #5230 which is recent so it's likely less breaking. Ideally we could consolidate all link-related info into `links` as an array of objects, but maybe it's not worth breaking still.

### Linked Issues

fixes #4992

### Additional Context

Couple things I realized that's not related to the PR:
1. If `ignoreDeadLinks` is a filter function, it looks like we pass the encoded url which probably isn't nice to use with.
2. Link lines didn't take into account of links that are in multiple lines but part of the same paragraph, for example:
	```md
    This is a [paragraph](./a)
    and this is the next part in the [paragraph](./b)
    ```
	It'll point all links to the first line of the paragraph. I can look into fixing this separately.

---

> [!TIP]


src/node/markdown/plugins/link.ts
```diff
@@ -124,7 +126,7 @@ export const linkPlugin = (
     }
 
     // export it for existence check
-    pushLink(url.replace(/\.html$/, ''), env, line)
+    pushLink(url, rawUrl, env, line)
```
bluwy, line 129: I made a small unrelated change to remove the `.replace(/\.html$/, '')` part as it felt unnecessary and should make things more predictable if people were to access it from `MarkdownEnv.link` (especially with how `cleanUrls` can affect it too).

This doesn't change how dead links are detected because the dead link check already also strips out the trailing `.html` itself.

Happy to revert this if there's concerns.

PR #5318, opened 2026-07-15
fix: remove stackTraceLimit Infinity for DEBUG

bluwy:
### Description

This was added since https://github.com/vuejs/vitepress/commit/4dac35fc9ec3eecb4371befa619626ee4db1ba58 for a temporary fix for shiki twoslash, but the workaround since then has been removed, except this part which I don't know if it's intentional.

I think it's best to not auto set this to Infinity though if DEBUG was set, and leave users to control this as needed.

### Linked Issues

n/a

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -13,10 +13,6 @@ import { init } from './init/init'
 import { clearCache } from './markdownToVue'
 import { bindShortcuts } from './shortcuts'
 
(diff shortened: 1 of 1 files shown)


brc-dd:
It's useful many times while debugging, but I guess we can remove it from defaults 🤷‍♂️ No hard opinion.

PR #5319, opened 2026-07-15
docs: update vite logo

bluwy:
### Description

Vite has updated to a new logo some time ago. This updates the logo.

image

I also updated the svgs to be referenced by `src`, since the new Vite logo SVG is kinda large. Plus it helps deduplication a bit between the translated docs.

One hiccup I had is that while the `vite.svg` can handle light-dark theme (based on its internal style tag), but since we render as an `img` via `src`, it doesn't work so I split as `vite-light.svg` and `vite-dark.svg`, which is nearly identical except of the default color styling for light-dark theme. This could be fixed by just following as before - inlining the SVG in the yaml directly, but I think avoiding the large SVG inlining is probably better for now (?).

Also, I guess we'd need to update the VitePress logo eventually to not use the old Vite logo?

### Linked Issues

n/a

### Additional Context

---

> [!TIP]


brc-dd:
light-dark is still broken if I use the theme toggle from navbar (it works if I switch the browser/system color scheme). I'll fix it.

image

brc-dd:
made the icon sizes bit more similar visually:

image<br />

also added the emojis as icons instead of text because that used to render weirdly on windows/linux.

PR #5323, opened 2026-07-16
feat: convert px to rem

### LINKED ISSUE #570: Font normalization makes user font size override in browser impossible

bluwy:
### Description

I used a script (generated by claude and manually tweaked by me) to automate this as best as possible. Please check the script, as I included some checks to skip conversion, e.g. for `box-shadow`, `border`, etc.

`convert-px-to-rem.js`:
```js
#!/usr/bin/env node

import fs from 'fs'
import path from 'path'

function convertFile(filePath) {
  let content = fs.readFileSync(filePath, 'utf8')
  const lines = content.split('\n')
  const result = []

  for (const line of lines) {
    // Skip lines with ok-to-use-px properties
    if (
      /\b(box-shadow|text-shadow|vp-shadow-\d|border|border-left|border-right|border-top|border-bottom)\b:/i.test(
        line
      )
    ) {
      result.push(line)
      continue
    }

    // Convert px to rem (16px = 1rem)
    const converted = line.replace(
      /(-?\d+(?:\.\d+)?)\s*px\b/g,
      (match, value) => {
        const num = parseFloat(value)
        // Ignore px and 1px
        if (num === 0 || num === 1 || num === -1) {
          return match
        }
        const rem = num / 16
        if (rem === Math.floor(rem)) {
          return `${Math.floor(rem)}rem`
        }
        // Remove trailing zeros after decimal
        return `${parseFloat(rem.toFixed(4))}rem`
      }
    )

    result.push(converted)
  }

  fs.writeFileSync(filePath, result.join('\n'), 'utf8')
}

function walkDir(dir) {
  const items = fs.readdirSync(dir)
  for (const item of items) {
    if (item === 'node_modules') continue
    const fullPath = path.join(dir, item)
    const stat = fs.statSync(fullPath)
    if (stat.isDirectory()) {
      walkDir(fullPath)
    } else if (item.endsWith('.css') || item.endsWith('.vue')) {
      convertFile(fullPath)
    }
  }
}

const basePath = process.cwd()
walkDir(path.join(basePath, 'src'))
walkDir(path.join(basePath, 'template'))
walkDir(path.join(basePath, 'docs'))

console.log('Conversion complete!')
```

I then followed up a second commit with handling of edge cases, especially around `position: absolute` and `position: fixed` styling, which might still need to use px. I scanned the changes and verified the changes as best as I could, the design still looks the same, but I might still overlook some parts.

### Linked Issues

fixes #570

### Additional Context

Maybe we'd want to use rem for borders too, but I ignored it for now as usually we don't. Also we'd need to followup updating the outline/indicator script to handle the rems

---

> [!TIP]


brc-dd:
checking this against tailwind once before merging (regarding which properties to keep/skip in the script). we can fix any regressions later if they happen.

PR #5324, opened 2026-07-16
docs: add base path prefix docs to sidebar reference

### LINKED ISSUE #5322: docs: missing documentation for sidebar base property
## Describe the bug / missing documentation

While exploring the VitePress source code, I noticed that the `Sidebar` and `SidebarItem` types support a `base` property to define a base path prefix for nested children (`items`):

```typescript
export interface SidebarMulti {
  [path: string]: SidebarItem[] | { items: SidebarItem[]; base: string }
}

export type SidebarItem = {
  // ...
  /**
   * Base path for the children items.
   */
  base?: string
  // ...
}
```

However, this extremely useful feature is currently not documented in the official sidebar reference guide ([`/reference/default-theme-sidebar`](https://vitepress.dev/reference/default-theme-sidebar)).

Without documentation, users might write repetitive, verbose paths in complex sidebar structures, unaware that this clean built-in solution exists.

## Suggested Improvement
I would like to add a new section in the sidebar reference documentation explaining:

1. What the `base` property is.
2. How to use it in `SidebarMulti` (multi-sidebar configuration).
3. How to use it in nested `SidebarItem` groups.

I will provide documentation for both the English and Chinese versions to keep them in sync.

## Would you like to submit a PR?
[x] Yes, I would like to submit a PR to fix this!

bd-dxg:
- add path prefix section to English sidebar reference docs
- add corresponding translation to Chinese sidebar reference docs
- include code examples for multiple sidebars and nested groups

### Description

This PR adds documentation for the `base` property in the Sidebar configuration (`SidebarMulti` and `SidebarItem`), which was previously undocumented. 

The `base` property allows users to define a base path prefix for nested children (`items`), effectively eliminating repetitive path definitions in both multi-sidebar configurations and deeply nested groups.

We have added:
1. Documentation and realistic code examples in the English sidebar reference guide.
2. The corresponding translated section and code examples in the Chinese sidebar reference guide.

### Linked Issues

fixes #5322

### Additional Context

- The documentation explains that the `base` path must start and end with `/`.
- Examples demonstrate usage in both a root-level multi-sidebar structure and nested groups (showing how a nested `base` overrides the parent prefix).
- Please verify if the wording and translated terms align perfectly with the rest of the reference docs.


docs/en/reference/default-theme-sidebar.md
```diff
@@ -184,3 +184,59 @@ export default {
...
+      {
+        text: 'Reference',
+        base: '/reference/',
+        items: [
+          // This link is resolved to `/reference/site-config`
+          { text: 'Site Config', link: 'site-config' },
+          {
+            text: 'Default Theme',
+            // Nested base overrides the parent path prefix
+            base: '/reference/default-theme-',
```
its-miroma, line 233: This example contradicts the rule above, that bases should start and end with /

brc-dd: Hmm, the example is fine, the description is wrong, base in sidebar can be anything.

its-miroma: I would have called the field `prefix` then, because `base` is already used in the config and requires it to start and end with /

Not a big deal, and it's too late to change now.

bd-dxg: > Hmm, the example is fine, the description is wrong, base in sidebar can be anything.

If the `base` value does **not** end with `/`, then the `link` of every object in `items` must **start** with `/`.

Conversely, if the `base` value **does** end with `/`, then the `link` of each object in `items` should **not** start with `/`.

As the current implementation stands, both forms work correctly and can navigate to the target page. Personally, I would prefer to keep the `base` value wrapped with `/` on both sides (for example, `/guide/`) because it feels more consistent to me.

However, if you still have concerns about this behavior, I’m happy to update the wording in this PR to make the rule clearer. It would be especially helpful if you could provide the exact description you’d like to use, since English is not my native language.

bd-dxg: What is your opinion now? @brc-dd

brc-dd: But the description you've written is ``**The value of `base` should start and end with a `/`.**``, which is not true. We are using this in our docs:

https://github.com/vuejs/vitepress/blob/711840222700804dbb6fb39ee9b9580a3e6220e7/docs/config.ts#L122

This doesn't end with `/`.

Rest of the part seems good, but that line should be removed.

bd-dxg:
docs: remove inaccurate base path slash constraint from sidebar docs

Remove the statement that base must start and end with / from EN docs

Remove the corresponding note from zh docs

its-miroma (approved):

Issue #5056, opened 2025-12-06
Improve screen reader accessibility


StrangeGirlMurph:
### Is your feature request related to a problem? Please describe.

I love vitepress but I noticed that the accessibility to screen readers could be improved. 

### Describe the solution you'd like

Here is a list of things I noticed and things to consider:

- nav element of active page should have `aria-current="page"`
- no alt text on logo in header
- the local search popup is just non-existent for a screen reader. not sure if minisearch is to blame or vitepress. 
- possibly put the home page actions in a nav
- nav elements could be in list
- social links could be announced as such and be a list 

Hope that helps :)

### Validations

- [x] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [x] Read the [docs](https://vitepress.dev).
- [x] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [x] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

lllleolin-max:
I revisited this against `main` at `3a49d35002f54531b121c99ee1b35239b99b7342`. A useful next step may be to narrow the remaining local-search work into an interaction contract, rather than treating all six items as still unimplemented.

Some context already exists:

- Exact-page `aria-current` is covered by [the navigation-link fix](https://github.com/vuejs/vitepress/commit/0f0fe135762489f1a6c73348bd8b0a86db969248) and the current [navigation tests](https://github.com/vuejs/vitepress/blob/3a49d35002f54531b121c99ee1b35239b99b7342/__tests__/e2e/navigation.test.ts). #5397 subsequently improved navigation disclosures and mobile focus handling.
- Logo alternative text is configurable with `logo: { src, alt }`, as explained in #3277. The logo-only example with `siteTitle: false` could demonstrate this directly so authors do not accidentally leave their home link unnamed.
- #5252 added search loading/busy semantics. The [earlier theme audit](https://github.com/vuejs/vitepress/blob/navbar-redesign-triage/triage/a11y-i18n.md) already records the remaining hardcoded loading announcement.

[...]

PR #5326, opened 2026-07-17
feat: use ul and li for lists

bluwy:
### Description

For list-like content, use `ul` and `li` tags. I go through these by searching for `v-for` and decide whether they should be using the `ul` and `li` tags.

Note that some changes may result in a nested `li` tag, but visually they should look the same. Also, I have to "reset" some of the `ul` and `li` margins as they may be added by `vp-doc`.

I manually verified the UI as best as I could to ensure nothing visually changes.

### Linked Issues

Addresses part of #5056

### Additional Context

Some markup I didn't treat as a list, such as sidebar groups, sponsor tier sections, and home page action buttons. Since they're not much of a list of content.

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/components/VPFeatures.vue
+++ b/src/client/theme-default/components/VPFeatures.vue
@@ -37,8 +37,8 @@ const grid = computed(() => {
 <template>
   <div v-if="features" class="VPFeatures">
     <div class="container">
-      <div class="items">
-        <div
+      <ul class="items">
+        <li
           v-for="feature in features"
           :key="feature.title"
           class="item"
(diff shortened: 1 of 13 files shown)


brc-dd:
probably need to disable the marker in sponsors/team related stuff

image<br />

https://deploy-preview-5326--vitepress-docs.netlify.app/reference/default-theme-team-page

bluwy:
Ah didn't notice that. I was using Safari to test and it wasn't shown.

brc-dd:
ah lol, yeah just checked. why isn't the marker shown there though? 👀 

unrelated but seems like external link icons are broken on safari as well (caused by https://github.com/vuejs/vitepress/pull/5232, inline vs inline-block):

image

bluwy:
Yeah I noticed the icons too and haven't had time to check that, because well there's only a dozen of Safari users like me I guess to prioritize 😄 

> ah lol, yeah just checked. why isn't the marker shown there though? 👀

I guess Safari has some kind of z-index layering for the list style that gets hidden if padding is 0

PR #5329, opened 2026-07-22
Fix TypeError on the home page

kantuni:
### Description

Fixes the following error:

```js
TypeError: can't access property "removeEventListener", e.value is null
```

See screenshot below

Screenshot 2026-07-22 at 00 09 03

### Linked Issues

N/A

### Additional Context

There might be a better fix here than a simple "no-op" 🤷‍♂️ 

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -100,7 +100,6 @@ export function useActiveAnchor(
 
   onUnmounted(() => {
     window.removeEventListener('scroll', onScroll)
-    container.value?.removeEventListener('click', onClick)
   })
 
   function onClick(e: MouseEvent) {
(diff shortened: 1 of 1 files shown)


src/client/theme-default/composables/outline.ts
```diff
@@ -100,7 +100,7 @@ export function useActiveAnchor(
 
   onUnmounted(() => {
     window.removeEventListener('scroll', onScroll)
-    container.value.removeEventListener('click', onClick)
+    container.value?.removeEventListener('click', onClick)
```
brc-dd: hmm, I think just delete this line. I don't think it's reachable because container will be destroyed on unmount and we don't need to clear the handlers manually.

```suggestion
```

Issue #5134, opened 2026-03-01
OOM (heap out of memory) during `vitepress build` with large dynamic routes (~26,000 pages)


tomiaa12:
## Describe the bug

When building a site with a large number of dynamic route pages (generated via `.paths.ts`),
the build process crashes with a JavaScript heap out of memory error during the "rendering pages" phase.

This happens even with `NODE_OPTIONS=--max-old-space-size=8192` (8 GB heap).

## Reproduction

- **VitePress version:** `2.0.0-alpha.16`
- **Node.js version:v22.21.1
- **OS:** Windows 11

**Route setup:**

- 20 locale directories, each with a `[source]-[target].paths.ts`
- Each paths file generates ~1,300 route pairs
- Total pages: ~26,000

```ts
// [source]-[target].paths.ts (simplified)
export default {
  paths: () =>
    generatePairs(list).map(([source, target]) => ({
      params: { source, target, ...meta },
    })),
}
```

**Error output:**

```
✓ building client + server bundles...
⠧ rendering pages...
<--- Last few GCs --->
[...] Mark-Compact 8012.2 (8226.2) -> 7997.8 MB ...
FATAL ERROR: Ineffective mark-compacts near heap limit
Allocation failed - JavaScript heap out of memory
```

## Expected behavior

Either:
1. VitePress streams/batches page rendering internally to keep memory bounded, or
2. Provides an official API / config option (e.g. `build.pagesPerChunk`) to limit how many pages are held in memory at once during SSR rendering.

## Current workaround

I implemented a manual chunked-build script that sets `BATCH_INDEX` / `BATCH_TOTAL` env vars,
filters `.paths()` output per batch, and merges the output dirs afterward.
This works but is fragile and requires maintaining a custom script.

## Questions

1. Is there a planned solution for large-scale static generation in VitePress?
2. Would it be feasible to add a streaming/incremental rendering mode (similar to how Nuxt/Next handle ISR)?
3. Is there a recommended way to handle sites with 10,000+ dynamic pages today?

### Describe the solution you'd like

Add a `build.maxPagesPerChunk` (or similar) config option that makes VitePress
render pages in batches internally, instead of loading all SSR-rendered pages
into memory at once.

Ideal behavior:
- Pages are rendered in streaming/incremental batches (e.g., 500 at a time)
- Memory usage stays bounded regardless of total page count
- No change required to user config or `.paths.ts` files
- Works transparently with dynamic routes, multi-locale sites, and sitemaps

[...]

Issue #4833, opened 2025-07-09
JavaScript heap out of memory when building large documentation sites with extensive mathematical formulas


cell-juicy:
### Describe the bug

When attempting to build my documentation site (containing thousands of markdown notes with extensive mathematical formulas) using VitePress, the build process fails with a JavaScript heap out-of-memory error. 

**What I did:**

1. Development server (`vitepress dev`) runs successfully
2. Production build (`vitepress build`) consistently fails with OOM error
3. Tried increasing Node.js memory limit to 8GB as suggested in common solutions
4. Build still fails with similar memory error.

**What I don't know:**

- Why development mode works but production build fails
- I tried splitting the file into two parts of similar size and executing the deploy command separately, and both build processes completed successfully. Is the problem simply because the site has too many files?

**Console output of failed build command:**

```bash
$ node --max-old-space-size=8192 node_modules/vitepress/bin/vitepress.js build docs

  vitepress v1.6.3


<--- Last few GCs --->

[7244:0000027693210080]   803322 ms: Mark-Compact (reduce) 8046.6 (8237.9) -> 8004.1 (8189.2) MB, 68426.88 / 0.00 ms  (average mu = 0.144, current mu = 0.012) allocation failure; GC in old space requested
[7244:0000027693210080]   869959 ms: Mark-Compact (reduce) 8050.2 (8235.2) -> 8004.0 (8188.7) MB, 65427.66 / 0.00 ms  (average mu = 0.087, current mu = 0.018) allocation failure; GC in old space requested


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 00007FF7D94D39EB node::SetCppgcReference+18459
 2: 00007FF7D9443238 DSA_meth_get_flags+93160
 3: 00007FF7D9EBC771 v8::Isolate::ReportExternalAllocationLimitReached+65
 4: 00007FF7D9EA5EC8 v8::Function::Experimental_IsNopFunction+1336
 5: 00007FF7D9D07970 v8::Platform::SystemClockTimeMillis+659328
 6: 00007FF7D9D13BF3 v8::Platform::SystemClockTimeMillis+709123
 7: 00007FF7D9D11554 v8::Platform::SystemClockTimeMillis+699236
 8: 00007FF7D9D04690 v8::Platform::SystemClockTimeMillis+646304
[...]

Issue #4415, opened 2024-12-09
Crash with JS heap OOM when editing LaTeX-heavy pages under `vitepress dev`


Locietta:
### Describe the bug

Recently, I wrote posts in [my vitepress blog](https://github.com/Locietta/blog-lost-pieces) with lots of LaTeX. While these LaTeX-heavy pages can be built and bundled without issue, I periodically get OOM when edit those pages with `vitepress dev`. (OOM doesn't happen on the initial launch, it often occurs after several edits. Is there memory leaking during hmr or something?)

Crash Log:
```
<--- Last few GCs --->
ince start of marking 674 ms) (average mu = 0.373[8712:000001C960111000]  1240866 ms: Mark-Compact (reduce) 4093.1 (4097.3) -> 4092.2 (4097.3) MB, pooled: 0 MB, 610.60 / 0.00 ms  (+ 0.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 615 ms) (average mu = 0.394[8712:000001C960111000]  1241870 ms: Mark-Compact 4092.2 (4097.3) -> 4092.2 (4098.3) MB, pooled: 0 MB, 594.69 / 0.00 ms  (average mu = 0.401, current mu = 0.408) allocation failure; GC in old space requested


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 00007FF75678199B node::SetCppgcReference+15707
 2: 00007FF7566EB928 DSA_meth_get_flags+97560
 3: 00007FF7573EC801 v8::Isolate::ReportExternalAllocationLimitReached+65
 4: 00007FF7573D9046 v8::Function::Experimental_IsNopFunction+2710
 5: 00007FF75720FED0 v8::internal::StrongRootAllocatorBase::StrongRootAllocatorBase+34016
 6: 00007FF7572091CD v8::internal::StrongRootAllocatorBase::StrongRootAllocatorBase+6109
 7: 00007FF7572050ED v8::Platform::SystemClockTimeMillis+655117
 8: 00007FF756A2044D BIO_ssl_shutdown+189
 9: 00007FF75720CA64 v8::internal::StrongRootAllocatorBase::StrongRootAllocatorBase+20596
10: 00007FF7572224C6 v8::Isolate::GetHeapProfiler+8150
[...]

Issue #4242, opened 2024-09-30
[Memory Leak]: Builds incorporating Shiki Twoslash crash with JavaScript heap out of memory errors


IMax153:
### Describe the bug

Hello 👋 from the [Effect-TS](https://github.com/Effect-TS) community :) 

We have recently been experimenting with using different frameworks for building our [documentation website](https://effect.website/docs). I decided to try building our documentation with Vitepress because of your native integration with Shiki Twoslash (via `@shikijs/vitepress-twoslash`), which we make extensive use of in our docs. 

However, when I attempt to build our documentation with Vitepress and Shiki Twoslash, I consistently receive a `JavaScript heap out of memory` error. I have attempted to increase the `--max-old-space-size` to ~8GB and still encounter this issue.

**With Twoslash Integration**

![Example of Vitepress build resulting in OOM error](https://github.com/user-attachments/assets/cb9e2341-6aa7-4022-88e1-3af18596036d)

In my `.vitepress/config.ts`, if I remove the `transformerTwoslash` from the `markdown.codeTransformers` array, the build succeeds without issue in ~20 seconds.

**Without Twoslash Integration**

![Example of Vitepress build succeeding](https://github.com/user-attachments/assets/19672a94-0e7b-4bea-9ada-ae51a83daa86)

Also, please let me know if it is more appropriate to move this issue to the Shiki project. If so, I have no problem closing this issue and moving it 👍 Thank you for your time!

### Reproduction

1. Clone our docs experiment

    ```sh
    git clone https://github.com/IMax153/vitepress-memory-leak-repro/tree/main
    ```

2. Install dependencies

    ```sh
    pnpm install
    ```

3. Attempt to build

    ```sh
    pnpm build
    ```

### Expected behavior

My expected behavior would be for Vitepress to be able to build our docs with Shiki Twoslash integrated without issue.

### System Info

```Text
System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.99 GB / 16.00 GB
    Shell: 5.9 - /nix/store/w53pgqwh5dm38skdd7n7wn0k6cw72jzj-zsh-5.9/bin/zsh
Binaries:
    Node: 20.17.0 - /nix/store/hxl1k8qgmrm1vfq5f419iv4wybz9szqq-nodejs-20.17.0/bin/node
[...]

PR #5342, opened 2026-07-30
perf(md): various performance improvements

its-miroma:
@IMB11 and I have been working on performance improvements for large sites such as the Fabric Docs. We are still working on a much bigger change related to batching the ssr builds with multiple workers.

In the meantime, these are some smaller, unrelated fixes that can probably be merged early.

### Linked Issues

Some issues which might be related to this PR, but probably not fixed by it:

#4242, #4415, #4833, #5134

### Additional Context

the commit messages include important context and reasoning for the changes; would be great if this PR weren't squashed.

Note regarding the implementation: I'm now pulling a new dep (p-limit); if you prefer I can probably commit a minimal implementation instead.

---

> [!TIP]

### DIFF
--- a/src/node/markdown/markdown.ts
+++ b/src/node/markdown/markdown.ts
@@ -510,6 +510,10 @@ export async function createMarkdownRenderer(
   if (options.component !== false) {
     componentPlugin(md, normalizePluginOptions(options.component))
   }
+  // pass an empty options object to gray-matter, otherwise it would memoize
+  // the results in an unbounded cache, where the key is the full file content.
+  // https://github.com/jonschlinkert/gray-matter/blob/310f9349381775d10a221cef903989eb5acc8843/index.js#L44-L47
+  ;(options.frontmatter ??= {}).grayMatterOptions ??= {}
   frontmatterPlugin(md, options.frontmatter)
   if (options.headers) {
     headersPlugin(md, {
--- a/src/node/markdownToVue.ts
+++ b/src/node/markdownToVue.ts
@@ -1,5 +1,6 @@
 import { resolveTitleFromToken } from '@mdit-vue/shared'
 import { LRUCache } from 'lru-cache'
+import { hash } from 'node:crypto'
 import fs from 'node:fs'
 import path from 'node:path'
 import { createDebug } from 'obug'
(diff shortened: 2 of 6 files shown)


src/node/utils/getGitTimestamp.ts
```diff
@@ -136,26 +138,36 @@ export async function cacheAllGitTimestamps(
...
-  let output = ''
-  child.stdout.on('data', (d) => (output += String(d)))
-  await once(child, 'close')
-
-  const ts = Number.parseInt(output.trim(), 10) * 1000
-  if (!(ts > 0)) return 0
-
-  cache.set(file, ts)
-  return ts
+  const pending = limitGit(async () => {
```
brc-dd: does limiting it make much difference? because for most cases this won't even be reachable as cache is run on configResolved.

src/node/markdown/markdown.ts
```diff
@@ -510,6 +510,7 @@ export async function createMarkdownRenderer(
   if (options.component !== false) {
     componentPlugin(md, normalizePluginOptions(options.component))
   }
+  ;(options.frontmatter ??= {}).grayMatterOptions ??= {}
```
brc-dd, line 516: this is interesting. can you add a comment here about gray-matter's unbounded cache behavior and that passing options disables that? (no need to add it at other `matter` call sites)

src/node/markdownToVue.ts
```diff
@@ -24,7 +25,13 @@ import { getGitTimestamp } from './utils/getGitTimestamp'
 import { processIncludes } from './utils/processIncludes'
 
 const debug = createDebug('vitepress:md')
-const cache = new LRUCache<string, MarkdownCompileResult>({ max: 1024 })
+const cache = new LRUCache<string, MarkdownCompileResult>({
+  max: 1024,
```
brc-dd: we can maybe drop max now, so that people having more than 1024 files can also use the cache till maxSize exhausts.

brc-dd:
rest looks good. only thing I don't think worth adding is the p-limit around git timestamp. it would've been useful earlier but now in almost all CI builds the cache hit rate should be 100%.

if possible please edit the commits and force push so that I can rebase-merge it instead of squashing as the commits are independent of each other.

src/node/utils/getGitTimestamp.ts
its-miroma: I have dropped the limiting.

You say that cache misses will happen rarely, so should I drop the deduplication of the calls too? As in, caching the Promise.

brc-dd: yeah I guess revert that file completely except for the stricter undefined check.

its-miroma: even that check is now redundant because no other logic paths will ever set the timestamp to 0.

PR #5363, opened 2026-08-12
fix: don't let meta attribute order affect head tag deduping

### LINKED ISSUE #5362: Head meta tags are omitted depending on attribute order
### Describe the bug

[`mergeHead` rigidly assumes that the first attribute in each head item is the name/key](https://github.com/vuejs/vitepress/blob/db28226b9a092a6510672d75527d55e7fe78e879/src/shared/shared.ts#L206). Not only is [this is not documented](https://vitepress.dev/reference/site-config#head), but if a user wanted to have multiple meta tags with different names but the same content (e.g., [`og:image` and `og:image:url`](https://ogp.me/#structured)), they must take care to put the `name` attribute before the `content` attribute, otherwise VitePress will only output the final meta tag.

### Reproduction

[Reproduction example](https://stackblitz.com/edit/vite-vfgeeawu?file=docs%2F.vitepress%2Fconfig.ts). Add the below to the VitePress config file:

```typescript
  // ...
  head: [
    // Only the second meta tag will show up in the final HTML:
    // <meta content="contentA" name="name2">
    [
      "meta",
      {
        // Order of attributes matters here!!!
        content: "contentA",
        name: "name1",
      }
    ],
    [
      "meta",
      {
        // Order of attributes matters here!!!
        content: "contentA",
        name: "name2",
      }
    ],

    // Both of these meta tags will show up in the final HTML:
    // <meta name="name3" content="contentB">
    // <meta name="name4" content="contentB">
    [
      "meta",
      {
        // Order of attributes matters here!!!
        name: "name3",
        content: "contentB",
      }
    ],
    [
      "meta",
      {
        // Order of attributes matters here!!!
        name: "name4",
        content: "contentB",
      }
    ]
  ],
  // ...
```

Note how the attribute order changes (i.e., `content` before `name` for the first two tags, then `name` before `content` for the latter two tags).

### Expected behavior

I expect VitePress to not silently omit meta tags. Moreover, I expect to be able to repeat meta tags (e.g., [`author`](https://zhead.dev/meta/author/), regardless of the order of the attributes within each head item object.

### System Info

```Text
System:
    OS: Linux 7.0 Ubuntu 26.04 LTS 26.04 LTS (Resolute Raccoon)
    CPU: (16) x64 AMD
    Memory: 5.76 GB / 17.88 GB
    Container: Yes
    Shell: 5.3.9 - /bin/bash
  Binaries:
    Node: 24.16.0 - /run/user/1000/fnm_multishells/147846_1786511297976/bin/node
    npm: 11.13.0 - /run/user/1000/fnm_multishells/147846_1786511297976/bin/npm
  npmPackages:
    vitepress: ^2.0.0-alpha.19 => 2.0.0-alpha.19
```

### Additional context

_No response_

### Validations

lazerg:
### Description

`mergeHead` picks whatever attribute happens to be first in the object as the key for deduping meta tags. If a tag is written with `content` before `name` (or `property`, etc.), it gets keyed on the content value instead of the name, so two different meta tags with the same content collide and one silently disappears. This skips `content` when choosing the key attribute, so dedup keys on the actual identifying attribute regardless of how the object is written.

### Linked Issues

fixes #5362

### DIFF
--- a/__tests__/unit/shared/shared.test.ts
+++ b/__tests__/unit/shared/shared.test.ts
@@ -1,0 +1,19 @@
+import { mergeHead } from 'shared/shared'
+
+describe('shared/shared', () => {
+  test('mergeHead dedupes meta tags regardless of attribute order', () => {
+    const result = mergeHead([
+      ['meta', { content: 'contentA', name: 'name1' }],
+      ['meta', { content: 'contentA', name: 'name2' }],
+      ['meta', { name: 'name3', content: 'contentB' }],
+      ['meta', { name: 'name4', content: 'contentB' }]
+    ])
+
+    expect(result).toEqual([
+      ['meta', { content: 'contentA', name: 'name1' }],
+      ['meta', { content: 'contentA', name: 'name2' }],
+      ['meta', { name: 'name3', content: 'contentB' }],
+      ['meta', { name: 'name4', content: 'contentB' }]
+    ])
+  })
+})
--- a/src/shared/shared.ts
+++ b/src/shared/shared.ts
@@ -203,7 +203,9 @@ export function mergeHead(...headArrays: HeadConfig[][]): HeadConfig[] {
   for (const current of headArrays) {
     for (const tag of current) {
       const [type, attrs] = tag
-      const keyAttr = Object.entries(attrs)[0]
+      const keyAttr =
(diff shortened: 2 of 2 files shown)


brc-dd:
I don't think any special handling is necessary. It should just be documented. Also, this won't fix the issue author's need to use multiple meta tags with same name.

Issue #5050, opened 2025-12-04
Top Navigation Version Switch and Language Dropdown Menus Not Displaying Properly in Safari (v26.1)


jiangshengdev:
### Describe the bug

https://github.com/user-attachments/assets/72b6e3be-a471-477a-b4d6-384d21bcd4ac

On websites using the VitePress default theme, the version switch dropdown menu (2.0.0-alpha.15)​ and language switch dropdown menu​ located on the right side of the top navigation bar fail to display correctly in Safari browser (version 26.1 (21622.2.11.11.9)).

[https://vitepress.dev/zh/guide/getting-started](https://vitepress.dev/zh/guide/getting-started)

### Reproduction

Open a VitePress-built website in Safari (version 26.1).
Click the version switch button​ (displaying e.g., "2.0.0-alpha.15") or the language switch button​ on the far right of the top navigation bar.

### Expected behavior

The dropdown menu should appear normally and be fully visible

### System Info

```Text
npx envinfo --system --npmPackages vitepress --binaries --browsers

  System:
    OS: macOS 26.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 368.81 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.1 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 11.6.2 - /usr/local/bin/npm
    pnpm: 10.23.0 - /usr/local/bin/pnpm
    Deno: 2.5.6 - /Users/jiangsheng/.deno/bin/deno
  Browsers:
    Chrome: 143.0.7499.40
    Safari: 26.1
  npmPackages:
    vitepress: 2.0.0-alpha.15 => 2.0.0-alpha.15
```

### Validations

- [x] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
- [x] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [x] Read the [docs](https://vitepress.dev).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

PR #5364, opened 2026-08-12
fix(theme): clip overflowed navbar menu

### LINKED ISSUE #2842: When browser width is between 840px and 770px a horizontal scrollbar appears when there are too many items
### Describe the bug

When browser width is between 840px and 770px, a horizontal scrollbar appears when there are too many items.

### Reproduction

- Add 10 items or more to your `nav` in `themeConfig`.
- Resize your browser to any width between 840px and 770px.
- Observe the horizontal scrollbar.

### Expected behavior

Sidebar items should overflow in an overflow menu.

### System Info

```sh
Binaries:
    Node: 19.6.1 - J:\nodejs\node.EXE
    npm: 9.4.0 - J:\nodejs\npm.CMD
    pnpm: 8.6.12 - C:\users\(redacted for PII)\Downloads\pnpm.EXE
Browsers:
    Edge: Spartan (44.22621.2199.0), Chromium (116.0.1938.54), ChromiumDev (117.0.2045.1)
    Internet Explorer: 11.0.22621.1
```

### Additional context

image

### Validations

bluwy:
### Description

In the navbar, make the navbar menu `flex-grow: 1` (instead of the search bar) and clip it if there's no more space left, e.g. the screen isn't wide enough. Now, the navbar shouldn't ever exceed the screen width and cause the scrolling in the first place

I'll make some comments below explaining some of the changes.

Here's also a video of the new behaviour:

https://github.com/user-attachments/assets/e38f8c41-1109-41c2-8d01-db7f9a169e9b

### Linked Issues

fix #2842

### Additional Context

In the linked issue, the scrolling happened in the first place on smaller screens is because the navbar became `position: relative`, while in larger screens it was `position: fixed`. I decided to fix this differently by making the navbar menu properly clipped in both cases instead.

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/components/VPFlyout.vue
+++ b/src/client/theme-default/components/VPFlyout.vue
@@ -82,14 +82,12 @@ function onBlur() {
 .button[aria-expanded="false"] + .menu {
   opacity: 0;
   visibility: hidden;
-  transform: translateY(0);
 }
 
 .VPFlyout:hover .menu,
 .button[aria-expanded="true"] + .menu {
   opacity: 1;
   visibility: visible;
-  transform: translateY(0);
 }
 
 .button {
@@ -132,6 +130,6 @@ function onBlur() {
   right: 0;
   opacity: 0;
   visibility: hidden;
-  transition: opacity 0.25s, visibility 0.25s, transform 0.25s;
+  transition: opacity 0.25s, visibility 0.25s;
 }
 </style>
--- a/src/client/theme-default/components/VPNavBar.vue
+++ b/src/client/theme-default/components/VPNavBar.vue
@@ -164,6 +164,7 @@ const { isHome, hasSidebar } = useLayout()
 
 .content {
   flex-grow: 1;
+  min-width: 0;
 }
 
 @media (min-width: 60rem) {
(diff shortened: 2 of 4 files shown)


src/client/theme-default/components/VPFlyout.vue
```diff
@@ -82,14 +82,12 @@ function onBlur() {
 .button[aria-expanded="false"] + .menu {
   opacity: 0;
   visibility: hidden;
-  transform: translateY(0);
```
bluwy, line 85: This was removed because of a Safari bug that causes the menu to get wrongly clipped. It was also reported in https://github.com/vuejs/vitepress/issues/5050 before.

I noticed an easy fix here is to not use `transform` here (as it's also not doing anything). This could break for users that set `translateY(-2px)` or something to have a cool floating effect, but I think it's fine? Alternative I also found that setting `will-change: transform` fixes it, but this technically incurs some small perf overhead for everyone.

src/client/theme-default/components/VPNavBar.vue
```diff
@@ -164,6 +164,7 @@ const { isHome, hasSidebar } = useLayout()
 
 .content {
   flex-grow: 1;
+  min-width: 0;
```
bluwy, line 167: This `min-width: 0` (on the flex item parent) allows shrinking the content beyond its content, same for the `min-width: 0`  in `VPNavBarMenu`. Both is needed to work.

brc-dd:
Moved to a different approach in #5397. Please check once if something from this PR is still relevant.

Issue #5360, opened 2026-08-11
Upgrade caused a font dilemma: numbered symbols break italics, and vice versa


yyx-dev:
### Describe the bug

After I updated to the latest version, I noticed some "issues" with the font rendering. Since the new version prioritizes the Inter font over system-ui (Microsoft YaHei on Windows), symbols like ① in my articles are rendered extremely large by Inter. In the stable version, they were rendered by Microsoft YaHei and appeared at a normal size.

However, if I move system-ui ahead of Inter in the font-family stack, the numbered symbols display correctly, but then italic rendering becomes problematic—whereas in the stable version, italics were rendered by Inter.

<img width="1787" height="623" alt="Image" src="https://github.com/user-attachments/assets/5dd73365-8b06-408f-8fe3-5edc3014d38a" />

So no matter how I adjust the order of system-ui and Inter, either the numbered symbols or the italics end up looking off—yet in the stable version, both were just right. But I don't want to give up the latest version, because the compilation speed is significantly faster.

Additionally, I noticed that the new version has removed system-ui from the font-family stack, and the top navigation bar is now rendered with Inter. The display effect for Chinese text is as follows—it's quite hard to make out the group headings and article titles.

<img width="265" height="682" alt="Image" src="https://github.com/user-attachments/assets/8e1d4269-54a3-4300-92ea-1a1166f9d19b" />

I suspect that something in the @font-face rules has changed, but I don't know much about front-end development and am hesitant to mess with it. Could anyone offer some guidance?

### Reproduction

This is what happened after I upgraded to the new version.

### Expected behavior

look like stable version.

### System Info

```Text
PS C:\Users\yyx_d> npx envinfo --system --npmPackages vitepress --binaries --browsers

  System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 AMD Ryzen 7 4800H with Radeon Graphics
    Memory: 8.71 GB / 15.37 GB
  Binaries:
    Node: 24.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.16.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 151.0.7922.76
    Internet Explorer: 11.0.19041.5794
```

### Additional context

Should I just copy all the @font-face rules from the stable version's font.css into the new version? If that really doesn't work, I'll just roll back to the stable version.

### Validations

- [x] Check if you're on the [latest VitePress version](https://github.com/vuejs/vitepress/releases/latest).
[...]

PR #5365, opened 2026-08-12
fix(theme): fix menu dropdown group dividers

bluwy:
### Description

#5326 introduced a bug where items in a dropdown menu didn't have the dividers properly rendered. This is because the styles are like `.VPMenuGroup + .VPMenuGroup` and `.VPMenuGroup:first-child` which didn't work because there's now a `li` wrapping the `.VPMenuGroup`.

In this PR, I took the simple approach of removing the nesting, and rendering the `li` directly in the components, i.e. `VPMenuLink` and `VPMenuGroup`. I don't know if this is the best pattern for project, but it seems somewhat safe as both are only used by `VPMenu`.

Alternatively, I was testing with `:global(li:has(> .VPMenuGroup) + li:has(> .VPMenuGroup))` for example, but it didn't look great so I tossed it. However, it's still an option on the table.

### Linked Issues

n/a. found this while checking things myself.

### Additional Context

Note, this doesn't affect `VPNavScreenMenu`. It looks like its components don't depend on this kind of dependent style.

Screenshots before & after (using groups syntax in the config):

image

image

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/components/VPMenuGroup.vue
+++ b/src/client/theme-default/components/VPMenuGroup.vue
@@ -9,13 +9,15 @@ defineProps<{
 </script>
 
 <template>
-  <div class="VPMenuGroup">
+  <li class="VPMenuGroup">
     <p v-if="text" class="title">{{ text }}</p>
 
-    <template v-for="item in items" :key="JSON.stringify(item)">
-      <VPMenuLink v-if="'link' in item" :item />
-    </template>
-  </div>
+    <ul>
+      <template v-for="item in items" :key="JSON.stringify(item)">
+        <VPMenuLink v-if="'link' in item" :item />
+      </template>
+    </ul>
+  </li>
 </template>
 
 <style scoped>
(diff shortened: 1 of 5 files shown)


src/client/theme-default/components/VPMenuGroup.vue
```diff
@@ -9,13 +9,15 @@ defineProps<{
...
 <template>
-  <div class="VPMenuGroup">
+  <li class="VPMenuGroup">
     <p v-if="text" class="title">{{ text }}</p>
 
-    <template v-for="item in items" :key="JSON.stringify(item)">
-      <VPMenuLink v-if="'link' in item" :item />
-    </template>
-  </div>
+    <ul>
```
bluwy, line 15: A drive-by change: Wrapped a `ul` here so that the group links are properly within a child `ul`.

brc-dd:
Thanks, yeah I noticed it here as well - https://github.com/vuejs/vitepress/issues/5360#issuecomment-5250365841

PR #5369, opened 2026-08-13
fix(theme): fix nav screen and divider appearance

bluwy:
### Description

So I ended up implemented no2 of https://github.com/vuejs/vitepress/issues/4972#issuecomment-5276394598 by adjusting the `top` and using `padding-top` to offset it back. no3 feels too jarring to me, you can try no3 out by checking out the first commit though.

I fixed the divider issue by using the divider element always instead of the border-bottom.

### Linked Issues

fix https://github.com/vuejs/vitepress/issues/4972
fix https://github.com/vuejs/vitepress/issues/5361

### Additional Context

Here's a video of me testing:

https://github.com/user-attachments/assets/ed7693c6-b1ed-4e53-ba76-bf24604ce5a8

If you're testing yourself, something I found difficult to handle is if the page is pinch-zoomed, the offset starts to become erratic. It happens previously too, but with this PR's `y` calculation, it's slightly more erratic. And sometimes to offset when zoomed seems wrong and covers over the divider (which it shouldn't).

Tried some tricks like `visualViewport.scale` but couldn't figure out one that works.

---

> [!TIP]
> The author can publish a _preview release_ by commenting `/publish` after creating the PR.

PR #5374, opened 2026-08-13
fix(theme): adjust z-index for VPNavBar and simplify VPNavScreen styles

brc-dd:
follow up on #5369

check if it doesn't cause any regressions while fixing the original issues + that zoom thing mentioned in that PR


bluwy (approved):
Looks a lot better 👍

brc-dd:
it could cause issues with global banners if the banners don't have z-index > 0, but it should be one line fix for users.

PR #4735, opened 2025-05-06
feat: add scroll support for the TOC

Plumbiu:
### Description

This PR implements that the table of contents scrolls as the page scrolls:

PixPin_2025-05-06_11-52-48

### Linked Issues

fixes #4734

### Additional Context

---

> [!TIP]
> The author of this PR can publish a _preview release_ by commenting `/publish` below.

PR #5377, opened 2026-08-14
feat(theme): scroll toc if active item is out of view

### LINKED ISSUE #4734: TOC should scroll as the page scrolls

When one of the items in the TOC goes out of the page, I need to scroll manually, which is inconvenient.

Image

Example:

Image

_No response_

### Additional context

_No response_

### Validations

bluwy:
### Description

Adds a `activeLink.scrollIntoView({ block: 'nearest', behavior: 'smooth' })` to auto-scroll to the toc active item if it's out of view. An alternative to #4735 where it always scrolls, this PR is less aggressive.

### Linked Issues

fixes #4734

### Additional Context

Test this on https://deploy-preview-5377--vitepress-docs.netlify.app/reference/site-config

NOTE: Because the outline marker has a transition on `top`, it'll look a bit jittery when auto-scrolling for both smooth and non-smooth scrolling. I found enabling smooth scrolling when calling `scrollIntoView` to be less distracting though.

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/composables/outline.ts
+++ b/src/client/theme-default/composables/outline.ts
@@ -171,19 +171,17 @@ export function useActiveAnchor(
   }
 
   function activateLink(hash: string | null) {
-    if (prevActiveLink) {
-      prevActiveLink.classList.remove('active')
-    }
+    const activeLink =
+      hash != null
+        ? container.value.querySelector<HTMLAnchorElement>(
+            `a[href$="${decodeURIComponent(hash)}"]`
+          )
+        : null
 
-    if (hash == null) {
-      prevActiveLink = null
-    } else {
-      prevActiveLink = container.value.querySelector(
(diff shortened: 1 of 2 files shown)


src/client/theme-default/composables/outline.ts
```diff
@@ -171,19 +171,17 @@ export function useActiveAnchor(
...
-    } else {
-      prevActiveLink = container.value.querySelector(
-        `a[href$="${decodeURIComponent(hash)}"]`
-      )
-    }
+    if (activeLink === prevActiveLink) return
 
-    const activeLink = prevActiveLink
+    prevActiveLink?.classList.remove('active')
+    prevActiveLink = activeLink
```
bluwy, line 184: I needed a small refactor here so that we early return if the next and previous active link is the same, otherwise a small scroll in the window might always cause the toc item to scroll into view.

PR #5382, opened 2026-08-17
fix(build): fail the build when SSR rendering errors occur

btea:
### Description
Refs: https://github.com/vitejs/vite/pull/23201

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/client/app/ssr.ts
+++ b/src/client/app/ssr.ts
@@ -7,6 +7,30 @@ export async function render(path: string) {
   const { app, router } = await createApp()
   await router.go(path)
   const ctx: SSGContext = { content: '', vpSocialIcons: new Set<string>() }
+
+  // Vue's SSR renderer reports component errors through the app's error
+  // handler and continues rendering instead of rejecting renderToString,
+  // which would silently ship broken pages. Collect the errors and fail
+  // the build instead.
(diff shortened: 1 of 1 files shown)


brc-dd:
Use `app.config.throwUnhandledErrorInProduction = true` in enhanceApp instead. Wrap it in `if (import.meta.env.SSR) { ... }` if you want the behavior similar to your PR.

Adding it by default, the way you did in PR, will leave no way for users to handle those errors if they wanted to suppress them or handle them differently.

If we want to be strict we can enable it in SSR by default, but that should be done [before the enhanceApp call](https://github.com/vuejs/vitepress/blob/b0da3845f8a740f08777d0b46783179998dd30ed/src/client/app/index.ts#L96), so that it can still be disabled.

sapphi-red:
> If we want to be strict we can enable it in SSR by default, but that should be done [before the enhanceApp call](https://github.com/vuejs/vitepress/blob/b0da3845f8a740f08777d0b46783179998dd30ed/src/client/app/index.ts#L96), so that it can still be disabled.

I think enabling by default makes sense. Are you open with doing so?

brc-dd:
Yeah, please create a PR. I'll merge it. Or I can do it too, it's 2-3 lines IG.

PR #5387, opened 2026-08-18
fix(theme): use relative length units to auto-close nav screen menu

bluwy:
### Description

The `VPNavScreenMenu` auto-closes when it's tablet size or bigger, but it's based on `768px` rather than `48rem` as converted in #5323. This PR updates to based on `48rem` with refactors and optimization.

### Linked Issues

n/a

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/client/theme-default/composables/nav.ts
+++ b/src/client/theme-default/composables/nav.ts
@@ -1,3 +1,4 @@
+import { useMediaQuery, whenever } from '@vueuse/core'
 import { useRoute } from 'vitepress'
 import { ref, watch, type InjectionKey } from 'vue'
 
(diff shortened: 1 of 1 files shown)


src/client/theme-default/composables/nav.ts
```diff
@@ -6,24 +7,19 @@ export function useNav() {
...
 
-  /**
-   * Close screen when the user resizes the window wider than tablet size.
-   */
-  function closeScreenOnTabletWindow() {
-    window.outerWidth >= 768 && closeScreen()
-  }
+  // Close screen when the user resizes the window wider than tablet size.
+  const isTablet = useMediaQuery('(min-width: 48rem)')
+  whenever(isTablet, closeScreen)
```
bluwy, line 22: This media query will now always listen instead of only when the screen is opened like before, but it should still be performant as the browser only triggers a change event when it crosses the threshold.

PR #5391, opened 2026-08-20
docs(markdown): update and correction of 'Part file' translation

lonewolfyx:
### Description

image

### Additional Context

---

> [!TIP]


docs/zh/guide/markdown.md
brc-dd: It's there in other places in that file too, can you update them too? Those file contents can probably be translated too.

PR #5400, opened 2026-08-26
docs(wizard-bun): fix error when run setup wizard with bun without x

farhnkrnapratma:
### Description

This error happen when you run setup wizard without `x` in `bun`:

```bash
$ bun vitepress init
error: Script not found "vitepress"
```

fix:
```bash
$ bunx vitepress init
```

or

```bash
$ bun x vitepress init
```


brc-dd:
You're supposed to install vitepress first. See the section above that.

Issue #2794, opened 2023-08-16
Disable automatically set direction in `<html>`


qmhc:
### Is your feature request related to a problem? Please describe.

If the direction of `<html>` is set dynamically via some js scripts in runtime, since the route path changed, this attribute (`dir`) will be override to `'ltr'`.

### Describe the solution you'd like

User can disable this behavior.

For example, user can specify `dir: false` to effect above.

### Describe alternatives you've considered

Now I'm using below workaround:

```ts
watch(() => router.route.path, syncVitepressDir)

function syncVitepressDir() {
  site.value.dir = isRtl.value ? 'rtl' : 'ltr'

  if (isClient) {
    document.documentElement.dir = site.value.dir
  }
}
```

But the `dir` attribute of `<html>` will be set to `'ltr'` then set to `'rtl'` in every time the route path changed.

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vitepress.dev).
- [X] Read the [Contributing Guidelines](https://github.com/vuejs/vitepress/blob/main/.github/contributing.md).
- [X] Check that there isn't already an issue that asks for the same feature to avoid creating a duplicate.

brc-dd:
For your rtl locale you can override dir in your config file, right? Why override it from runtime?

qmhc:
I'm writing rtl feature for a UI lib project. In its document website, user can toggle direction via a switch, for previewing the rtl feature.

bluwy:
The cause in discussion: https://github.com/vuejs/vitepress/blob/ee028266a8fee777a8ee247b1c4490432c0a830e/src/client/app/index.ts#L43-L49

PR #5437, opened 2026-09-13
feat(theme): native rtl support

brc-dd:
Closes #5034
Closes #5071

## Summary

Set `dir: 'rtl'` and the default theme mirrors itself. No PostCSS plugin is needed anymore. This works site-wide, per locale, and per page through frontmatter.

Both earlier PRs took the same direction: logical properties, a multiplier for transforms, and isolation for mixed-direction text. This PR re-derives that work against the current theme and completes it, so it supersedes both. Thanks to @rami-elementor and @ws-rush for the groundwork.

## Layout

- Every physical inline-axis declaration now uses its logical form. That covers the `-left` and `-right` margins, paddings, borders and insets, the physical corner radii, and `text-align: left/right`.
- The block axis is untouched. Converting it changes nothing in a horizontal writing mode.
- All 33 `/*rtl:…*/` control comments are gone.

## Transforms and icons

- Some things have no logical form: the sidebar slide, the appearance-switch knob, the hero image nudge, and the hamburger bars. They multiply by `--vp-direction-multiplier`, which is `1` on `:root` and `[dir='ltr']` and `-1` on `[dir='rtl']`.
- Any `dir` attribute resets the multiplier. An LTR island inside an RTL page stays put at any depth.
- The horizontal arrows and chevrons, plus `vpi-align-left`, mirror at the source with the `scale` property. It composes with the components' own `transform: rotate()` rules instead of replacing them.
- The external-link arrow is an inline pseudo-element and cannot be transformed. It swaps to a mirrored mask instead.

## Why there is no `:dir()`

Vite's default Lightning CSS target is chrome111. At that target `:dir()` is rewritten into a `:lang()` allowlist, and only in production builds. Attribute selectors and logical properties pass through the build untouched, so the theme uses only those.

## Code blocks

- The wrapper carries `dir="ltr"`, as the shiki `<pre>` already did. The code, the copy button, the language label and the line-number gutter keep their physical layout.
- Inline code uses `unicode-bidi: plaintext`. Things like `--flag` and `useData()` keep their written order inside RTL prose.
- The copied tooltip and the copy fallback's off-screen textarea follow the same rule.

## Mixed-direction text

Labels from config or content resolve their own direction. This covers nav items, flyouts, menus, sidebar items, outline links, pager titles, the hero, features, team pages, the footer, badges, search crumbs and the 404 page.

- Inline labels get `dir="auto"`.
- Labels inside block-level containers get a `<bdi>`, so the container keeps the page's alignment.
- The last-updated date is a `<time dir="auto">`.
- The search shortcut keeps its ⌘K order.

## Plumbing

- The dev server's HTML shell now carries the requested locale's `lang` and `dir`. It had neither before, so every dev page painted LTR until mount.
- The client sets both attributes from a watcher created before the theme mounts, so mounted hooks measure the right layout.
- `dir` is typed as `'ltr' | 'rtl' | 'auto'`.

## Behaviour changes to review

- The appearance switch mirrors in RTL. The knob rests on the right and travels left. The old rtlcss directives pinned it physical.
- The hamburger's staggered bars mirror.
- The mobile language menu's open-state chevron has pointed left since the icons moved to CSS masks. It now points up.
- `aside: 'left'` means the reading-start side, which is the right in RTL. This is documented; the option itself is unchanged.
- Narrowing `dir` from `string` is a type-level break for configs that pass a plain string.

## Migration

- Remove `postcss-rtlcss` or any other RTLCSS plugin. It would now flip the mirrored transforms a second time.
- Prefer logical properties in custom CSS. Mirror your own directional icons with `[dir='rtl'] .icon { scale: -1 1 }`. The i18n guide covers both.

## Docs

- The RTL section of the i18n guide is rewritten. Only the English copy is updated; the translations stay as they are until lunaria flags them.
- `dir` gets entries in the site-config and frontmatter references. The `aside` option notes that its sides mirror.
- The docs site drops `postcss-rtlcss` and gains RTL variants of its angled hero gradients.

## Tests

- A new e2e page with `dir: rtl` frontmatter checks the sidebar side, the outline marker's side and its scroll tracking, mirrored carets, the external-link arrow, the code islands, inline-code glyph order and the mobile off-canvas slide. It runs in dev and build mode.
- A unit test scans the theme's styles for physical inline-axis declarations, including one-line rules and asymmetric shorthands.
- Beyond the suites, the branch was checked in Chromium, Firefox and WebKit at four viewports. That covered site-level, locale-level and frontmatter `dir`, `base: '/sub/'` and `base: './'`, hydration, and toggling `dir` at runtime. The `/fa/` docs locale was the live check.

## Known limitations

- A `build.cssTarget` below Chrome 89 / Safari 15 makes Lightning CSS lower logical properties into `:lang()`-keyed rules. The layout would then follow `lang` rather than `dir`. The default target is unaffected.
- GFM table column alignment (`:---`) stays physical, per the spec.
- The Algolia DocSearch modal ships its own physical CSS.
- Arabic and Hebrew font stacks and `letter-spacing` resets are typography rather than direction. Worth a follow-up.

## Related

- Answers discussions #1468, #2336 and #2673.
- #2794 and #5416 (`dir: false` for runtime toggling) are a separate concern and unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brc-dd:
some parts will need follow ups and manual testing, but merged this for now. lets fix them as they are discovered.

PR #5438, opened 2026-09-13
fix: update font-family settings for improved language support

tats-u:
### Description

<dl>
<dt>Prefers Noto Sans CJK to good old Microsoft fonts</dt>
<dd>

Good old Windows standard fonts from before 2010, such as Microsoft YaHei and Meiryo, are overly optimized for 96 dpi displays. The Noto Sans CJK fonts included with recent versions of Windows (e.g., Noto Sans SC and Noto Sans JP; note that only the subset for the CJK language in use is installed by default) feature character shapes with fewer quirks and provide a consistent look with Linux-based operating systems and other CJK languages.

Furthermore, Noto Sans CJK offers a wide range of weights from extremely thin to extremely bold, and its aesthetic is similar to Hiragino Sans, making it possible to match the appearance on Apple devices.

Even on a 96 dpi display, as long as the text is not rendered extremely small, I think Noto's readability is not much inferior to the older Windows standard fonts. Rather, it has an appropriate stroke weight and renders smoothly. Also, unlike older Windows standard fonts other than Yu Gothic, Noto supports text-spacing-trim, which removes the excess whitespace that occurs with certain combinations of consecutive punctuation marks (as defined by the typesetting rules of each language).

Meiryo and YaHei cannot remove the extra spacing that occurs between parentheses, brackets, or quotation marks and adjacent punctuation:

image

</dd>
<dt>Respects Traditional Chinese locales</dt>
<dd>

In current VitePress, Simplified Chinese fonts are assigned to Traditional Chinese pages. When it is clear that a page is for Taiwan, Hong Kong, or Macao, fonts optimized for those regions should be prioritized. When it is unclear which of those three regions the page is for, the user's preference should be respected. Microsoft JhengHei is optimized for Taiwan, so it should not be prioritized for pages intended for Hong Kong or Macao.
</dd>
<dt>Reflects internal lang change</dt>
<dd>

In the language selector, "<span lang="ja">日本語</span>" on the Simplified Chinese page and "<span lang="zh">简体中文</span>" on the Japanese page are not rendered in the font of their respective languages:

image

"<span lang="zh">简体中文</span>" is rendered by Meiryo but it should not be.

image

"<span lang="ja">日本語</span>" is rendered by YaHei but it should not be.

</dd>
<dt>Adjusts Yu Gothic weights for non-Japanese Windows</dt>
<dd>

Yu Gothic, which is used when displaying Japanese pages on non-Japanese versions of Windows, has a Regular weight that is only about as thick as the Light weight of other fonts. It is simply much too thin. Therefore, I made it use Medium by default, and Bold (though it is still only about as thick as the SemiBold of a typical font) for bold text. See https://github.com/tats-u/yu-gothic-css/ for details.

</dd>
</dl>

### Additional Context

---

> [!TIP]


tats-u:
Needs to adjust subsetFonts.py.

tats-u:
**Before**

image

YaHei looks a little too pixelated compared to Inter. The Chinese ellipsis is positioned at the bottom rather than in the center. Given our commitment to fine-tuning the Latin font, this flaw cannot be overlooked.

image

Meiryo looks too thin and pixelated compared to Inter.

**After**

image

image

Noto Sans CJK (Noto Sans SC/JP) harmonizes with the look of Inter better than the "good" old built-in Windows CJ(K) fonts.

PR #5446, opened 2026-09-13
fix(markdown): preserve URL settings in container titles

lllleolin-max:
Links in custom container titles and details summaries lose the page's URL settings. For example, with `base: './'`, `[Guide](/guide/start.md)` on `nested/page.md` becomes `../guide/start.html` in the body but `/guide/start.html` in the title, breaking links when the build is mounted below the origin root or opened from disk. Titles also incorrectly retain `.html` when `cleanUrls` is enabled.

Pass the existing `cleanUrls`, `relativePath`, and `relativizeUrls` fields into title rendering. Keep the environment selective so page footnotes are not rendered into titles, and preserve the content loader's intentional site-absolute links.

Validation: the new URL matrix fails three cases on current main and passes with the fix; all 489 unit tests, `pnpm test:types`, `pnpm build` (including typecheck and package checks), and formatting checks pass. A minimal site with `base: './'` also builds with the expected relative links in both a tip title and a details summary. Tested on Windows with Node 24.18.0 and pnpm 11.19.0, using the unchanged frozen lockfile; no browser suite was run.

AI assistance was used to implement the fix and verify the regressions. The repository declares pnpm 11.24.0; local validation used the version noted above without changing the lockfile.


its-miroma:
Question: after this PR, is this section in the docs still relevant?

> Keep [`cleanUrls`](../reference/site-config#cleanurls) off (the default): portable output needs links that end in `.html`, since there is no server to rewrite pretty URLs.
> — https://vitepress.dev/guide/deploy#relocatable-builds-relative-base

brc-dd:
> is this section in the docs still relevant?

Yes? How's related to this PR? 👀 If you skip `.html`, things like serving from file:// scheme or ipfs won't work. You can enable cleanUrls + relative base if your hosting does the rewrite.

